PR #905, opened 2014-06-05
feat(elementexplorer): allow cmd params

lksv:
allow to paste parametrs form command line, namely:
* chromeOnly - use chrome driver directly
* confiFile  - initialize browser by specified protractors' config file

... and other params same as in protracotr


juliemr:
This will be taken care of in #1693, so closing this issue.

PR #942, opened 2014-06-17
Add configurable delay between concurrent instances

ashaffer:
When running a large number of instances simultaneously, launching many (in our case, 15+) instances of chrome (maybe other browsers too, but I only tested with chrome) at exactly the same moment causes webdriver to fail with the error "cannot get automation extension from unknown error: failed to close UI debuggers".

Staggering their launch by 1 second or so seems to resolve this completely.


timlib:
I also have this problem.

PR #966, opened 2014-06-23
feat(webdriver-manager): use proxy for webdriver-manager

hankduan:


katlimruiz:
at first it didnt work for me.
How to solve? Set env variables as below:
$ set HTTP_PROXY=http://proxy.com:8080
$ set HTTPS_PROXY=http://proxy.com:8080
and then run your command
(in my case it was due to angular-seed, so when bower install ran, it picked up these variables)

ericis:
Doesn't work for a couple of us. We tried `HTTP_PROXY` and `HTTPS_PROXY` and setting node config proxy and still get the error: `Error: tunneling socket could not be established, cause=getaddrinfo ENOTFOUND 9091 9091:80`. Other internet commands are working with the above proxy settings correctly

PR #968, opened 2014-06-24
feat(webdriver-manager): minor proxy enhancements

nwinkler:
Added error handling for request - previously, any errors coming from the request module were silently swallowed.

Also evaluating both uppercase and lowercase proxy variables. Many tools use proxy variables in the form https_proxy, others use HTTPS_PROXY. I changed the code so both forms are evaluated. See the following links for more info on this practice:
- http://serverfault.com/q/571887/140074
- https://wiki.archlinux.org/index.php/proxy_settings#Environment_variables


ericsoucy:
hello,
how do you pass proxy authentication info (username/password) on windows 7 ?

PR #1155, opened 2014-08-04
Change defer-bootstrap logic

hankduan:
*Bootstrap is done immediately when you call (add/remove/clear)MockModule
*Angular will deal with subsequent bootstrap without pausing for
   protractor (and bootstrap)
*browser.get (and other navigation) will be a straight call to wd
*back/forward will start working for protractor (with injected module)
*if another site redirects to Angular (e.g. login), mock modules can now be
   injected

Note: testapp/lib/angular_v1.2.9/angular.js is included for reference
      and will not be submitted as part of this PR (there will be a PR
      for angular)

Breaking changes:
*(add/remove/clear)MockModule will take an addition param (url), which
   is the url you are modifying the mock modules for. Injection will
   be done immediately.
*browser.get or any action will timeout if you have a $timeout > 10s
   in the app. ($timeout with waitForAngular was not supported before,
   just that now even page will not load subsequently)

Compatibility:
*bootstrap will not work for subsequent get/navigate if your browser
   does not support localSession
*bootstrap will not work for subsequent get/navigate if you are
   testing an older version of Angular

[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/1155)

### DIFF
--- a/debugging/polling_spec.js
+++ b/debugging/polling_spec.js
@@ -37,3 +37,16 @@ describe('synchronizing with pages that poll', function() {
     browser.ignoreSynchronization = false;
   });
 });
+
+/* 
+ * Note that after this test runs, $timeout will be running
+ * continuously, so any action/browser.get (anything that requires
+ * calling waitForAngular will fail
+ * i.e.
+ *   it('form after polling', function() {
+ *     browser.get('index.html#/polling');
+ *     var startButton = element(by.id('pollstarter'));
+ *     startButton.click();
+ *     browser.get('index.html#/form');
+ *   });
+ */ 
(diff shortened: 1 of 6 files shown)


juliemr:
After discussions today: - let's rebase this and actually take a look at it! We think it will be backwards compatible so shouldn't introduce problems. Also need to check browser compatibility for localstorage stuff.

americoneto1:
@juliemr I need to use addMockModule with manual bootstrap (requirejs) for my E2E tests. Do you know if there is any workaround to resolve this issue?

sjelin:
@juliemr I assume Hank is no longer working on this?

sjelin:
@juliemr @mgiambalvo @cnishina Ping

sjelin:
@juliemr ping ¿hola como estas? this issue hasn't gone away I still keep getting questions about it

juliemr:
We're going to add this to our things to address in Q2.

juliemr:
I think this is out of date enough to just close - we want to look at these issues, but leaving this PR open ain't helping.

PR #1391, opened 2014-10-01
fix(page get): Make protractor return destination when using file://

### LINKED ISSUE #1390: Protractor failing when using file://
I'm not sure if I am using it exactly in the correct way as I can't see much documentation around this feature. But I saw that there is the possibility of using `file://` as the `baseUrl` option. 

Now I am using Chrome in order to run my Protractor tests and wanting to run them whilst not running against a server as there really isn't any need for the implementation. When running Protractor with this configuration it can't seem to get the file the only thing that happens is that the address comes up with 

`data:text/html,<html></html>`

Which then leaves me with the following output 

```
Failures:

  1) Test that my App can get Mocked data Should be on the correct page
   Message:
     Error: Timed out waiting for page to load after 10000ms
Wait timed out after 10062ms
   Stacktrace:
     Error: Timed out waiting for page to load after 10000ms
Wait timed out after 10062ms
==== async task ====
Timed out waiting for page to load after 10000ms
    at null.<anonymous> (/home/nwhite/Sites/apibyexample/abe-protractor/tests/e2e/app/myTestApp.spec.js:3:17)
==== async task ====
Asynchronous test function: beforeEach()
Error
    at null.<anonymous> (/home/nwhite/Sites/apibyexample/abe-protractor/tests/e2e/app/myTestApp.spec.js:2:5)
    at Object.<anonymous> (/home/nwhite/Sites/apibyexample/abe-protractor/tests/e2e/app/myTestApp.spec.js:1:63)

Finished in 12.435 seconds
1 test, 1 assertion, 1 failure

Shutting down selenium standalone server.
```

Is there any documentation available around this? is it possible using Chrome as I could only see something mention `node-webkit`? Is there any different configuration options that I need to implement?

Thanks in advanced

nwhite89:
To allow protractor to use file:// for running tests, it is not possible to force setting window location via JavaScript.

This will close #1390

### DIFF
--- a/lib/protractor.js
+++ b/lib/protractor.js
@@ -1211,12 +1211,12 @@ Protractor.prototype.get = function(destination, opt_timeout) {
   var timeout = opt_timeout ? opt_timeout : this.getPageTimeout;
   var self = this;
 
-  destination = this.baseUrl.indexOf('file://') === 0 ?
-    this.baseUrl + destination : url.resolve(this.baseUrl, destination);
-
-  if (this.ignoreSynchronization) {
+  if (this.ignoreSynchronization || this.baseUrl.indexOf('file://') === 0) {
+    destination = this.baseUrl + destination;
     return this.driver.get(destination);
   }
+
+  destination = url.resolve(this.baseUrl, destination);
 
   this.driver.get(this.resetUrl);
   this.driver.executeScript(


mackwic:
Hi there. Sorry for digging up, but node-webkit applications using angular are fully implemented with file://, and these angular application depends on code not available in browser context.
The support for file:// is really wanted, here.

@juliemr Could you explain the issue with the mocks modules ?
@nwhite89 it seems that you used node-webkit, anything to share ? How did you do the testing ?

Issue #876, opened 2014-05-28
Cucumber-JS stops on failure


stejanse:
Since 0.23.0, Cucumber-JS stops after a failing step in Protractor. Before 0.23.0 it continues executing other scenario's and reporting the final results. This appears to be a bug in Protractor.

**0.23.0:**

``` cucumber
Feature: some feature

  Scenario: some scenario
   When some step
   Then some FAILURE
```

**before:**

``` cucumber
Feature: some feature

  Scenario: some scenario
   When some step
   Then some FAILURE

  Scenario: other scenario
    When ...

xx scenarios (1 failed)
xx steps (1 failed, xx passed)
```

jlin412:
@jbpros I try cuke-tractor with the latest protractor 2.0.0 and cucumber.js 0.4.8.
It seems that it is failing again for the following test case:
```
 this.Then(/^I click on a bad locator$/, function (next) {
    $('bad_css_path').click().then(next);
  });
```

jbpros:
@jlin412 :cry: did you try to find the cause of that new breakage?

jlin412:
I haven't pinpointed which commit.  I just know that the last working build
is 1.8.0.  I will test it today or tomorrow.

jlin412:
@jbpros I update https://github.com/jlin412/cuke-tractor
With latest 2.1.0, failure case is different.
Original failure in 2.0.0 is passing now.
``` javascript
 this.Then(/^I click on a bad locator$/, function (next) {
    $('bad_css_path').click().then(next);
  });
```
but the following two cases inside *then* block are failing:
```
  this.When(/^something fails via exception in then function$/, function (callback) {
    $('css_bad_path').isPresent().then(function(isElmentPresent) {
      throw new Error('#FAIL via exception');
      callback();
    });
  });

  this.When(/^failed inside then call$/, function (callback) {
    $('css_bad_path').isPresent().then(function(isElmentPresent) {
      expect(isElmentPresent).to.equal(true);
      callback();
    });
  });
```
I have uploaded my test cases in my fork of cuke-tractor.
https://github.com/jlin412/cuke-tractor

aluzardo:
I am getting the same error if I use a promise inside my step definition.

This is the promise

this.waitUrl = function (urlRegex, time) {
        return browser.wait(function() {
            return browser.getCurrentUrl().then(function(url) {
                return urlRegex.test(url);
            });
        }, time, "url has not changed to "+urlRegex);
    };


[...]

PR #1394, opened 2014-10-02
fix(runner): prevent source-map-support module from exiting process on `...

kayhadrin:
This is a quick patch for issue #876 which allows us to let Cucumber keep going even when a test step fails (the problem seems to mostly happen when it's a protractor promise that fails).

However, I can still see that the error stacks showing CoffeeScript source files are displayed with line numbers that actually match the js generated code, not the CoffeeScript code.
So I think that the source-map-support configuration is not correct when used with Cucumber js, but this should be fixed in another PR I reckon.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -201,7 +201,30 @@ Runner.prototype.setupGlobals_ = function(driver) {
   global.by = global.By = protractor.By;
 
   // Enable sourcemap support for stack traces.
-  require('source-map-support').install();
+  var sourceMapSupport = require('source-map-support');
+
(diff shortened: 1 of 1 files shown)


sjelin:
Thoughts on this @juliemr?  Looks fine to me but I'm not super familiar with the details of how these things work.

jlin412:
Is this PR still relavent after recent fix from cucumber.js 0.4.7?

hankduan:
Closing as this is no longer necessary after https://github.com/angular/protractor/issues/876 is fixed from the cucumber side.

Issue #1247, opened 2014-08-27
add by.exactRepeater()


elgalu:
[This came up in a Stack Overflow question](http://stackoverflow.com/a/25533985/511069)

Given inside the ng-repeat attribute value indexOf is used:

```js
for (var i = 0; i < repeatElems.length; ++i) {
  if (repeatElems[i].getAttribute(attr).indexOf(repeater) != -1) {
    rows.push(repeatElems[i]);
  }
}
```

NickTomlin:
Closing as this was included in https://github.com/angular/protractor/commit/324f69d6aa7c23ad77f1d50e26e0a56bade40132. Docs [here](https://angular.github.io/protractor/#/api?view=ProtractorBy.prototype.exactRepeater).

PR #1662, opened 2014-12-23
feat(exactRepeater): Add a new locator by.exactRepeater - Protractor iss...

guicejas:
...ue #1247

Doing a new PR after solve some CLA issues


sjelin:
Adding a new parameter for `by.repeater` to require exact matches should be pretty easy.  Thoughts @juliemr?

sjelin:
Hi @guicejas, sorry for taking so long.  We're going to do this, but in a way that involves less code duplication.  See https://github.com/angular/protractor/pull/1662

Issue #362, opened 2013-12-18
Jasmine 2.0 is out, explore updating to that


juliemr:

nishakchaudhari:
So timeout for individual test has been fixed? I tried running my test with jasmine2. It gives me timeout error.

hankduan:
I'm not sure what is your question. Jasmine2 doesn't make timeouts all magically disappear. If your test doesn't finish without a specified time, or your app isn't load, etc, your test will still time out. See https://github.com/angular/protractor/blob/master/docs/timeouts.md

nishakchaudhari:
So I want to know how to setup timeout for individual test with jasmine2.
Before I was doing it in this way : it(description, testFn, timeout_in_millis). And it worked when my jasmine version was <2.
I upgraded jasmine version and now I get following error:
Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL

vsathyak:
@nishakchaudhari  : I started using protractor and jasmine 2 and I am also facing the same issue. Any work around for this??

nishakchaudhari:
upgrade jasmine version to 2.3.1 . jasmine 2.3.1 supports timeout for individual test .

srajwadi:
How to update jasmine version. I am currently using 1.3.1 and protractor version is 2.1.0. Can anyone say that how to update jasmine version. I am new to this.
Thanks..

hankduan:
http://www.protractortest.org/#/jasmine-upgrade

mocsharp:
how can I have beforeAll wait for async calls in Jasmine 2.0? Thanks.

PR #1675, opened 2014-12-31
add jasmine2 framework

hankduan:
This PR enables protractor to work with jasmine2.0 (see https://github.com/angular/protractor/issues/362)

Pending https://github.com/angular/jasminewd/pull/13 for the release of jasminewd2

Note, passing timeout as part of `it`, `fit`, etc does not work yet. Pending https://github.com/jasmine/jasmine/commit/a4faa80be464b1054293ac3179eb21b2ccc8c472

Also, stacktrace is messed up when calling `done.fail('xyz')` because jasmine2 does not allow custom stacktrace. Filed an issue at https://github.com/jasmine/jasmine/issues/734

### DIFF
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -236,7 +236,7 @@ exports.config = {
   // ----- The test framework --------------------------------------------------
   // ---------------------------------------------------------------------------
 
-  // Test framework to use. This may be jasmine, cucumber, or mocha.
+  // Test framework to use. This may be jasmine, jasmine2, cucumber, or mocha.
   //
   // Jasmine is fully supported as a test and assertion framework.
   // Mocha and Cucumber have limited beta support. You will need to include your
@@ -255,6 +255,18 @@ exports.config = {
     includeStackTrace: true,
     // Default time to wait in ms before a test fails.
     defaultTimeoutInterval: 30000
+  },
+
+  // Options to be passed to jasmine2.
+  //
+  // See the full list at https://github.com/jasmine/jasmine-npm
+  jasmineNodeOpts: {
+    // If true, print colors to the terminal.
+    showColors: true,
+    // Default time to wait in ms before a test fails.
+    defaultTimeoutInterval: 30000,
+    // Function called to print jasmine results
+    print: function() {},
   },
 
   // Options to be passed to Mocha.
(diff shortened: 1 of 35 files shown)


hankduan:
Docs to go along with this: https://github.com/angular/protractor/pull/1679

juliemr:
Note that the travis error is due to jasminewd2 not yet being published.

docs/referenceConf.js
```diff
@@ -236,7 +236,7 @@ exports.config = {
   // ----- The test framework --------------------------------------------------
   // ---------------------------------------------------------------------------
 
-  // Test framework to use. This may be jasmine, cucumber, or mocha.
+  // Test framework to use. This may be jasmine, jasmine2, cucumber, or mocha.
```
juliemr: Let's change the value listed below to 'jasmine2' as well

juliemr:
Change the default value in configParser to jasmine2

spec/basic/locators_spec.js
```diff
@@ -12,10 +12,16 @@ describe('locators', function() {
 
     it('should allow custom expectations to expect an element', function() {
       this.addMatchers({
```
juliemr: It might be helpful to show an example more like this in the documentation PR.

spec/basicConf.js
```diff
@@ -4,6 +4,8 @@ var env = require('./environment.js');
 exports.config = {
   seleniumAddress: env.seleniumAddress,
 
+  framework: 'jasmine2',
```
juliemr: If we change the default then we can remove these edits.

juliemr: Ah but changing the default would break people who rely on it. Never mind, don't do that.

juliemr:
Do you have a preference for how this gets released? We could do a separate branch in the same repository, like we're doing with minijasminenode, or make a new repo for jasminewd2.

docs/referenceConf.js
hankduan: As in "Jasmine2 is fully supported as a test and assertion framework."?

spec/basic/locators_spec.js
hankduan: replaced the example in the documentation PR with this.

PR #1679, opened 2015-01-02
docs(jasmine-upgrade): add docs for upgrading jasmine from 1.3 to 2.0

hankduan:


sjelin:
LGTM

docs/jasmine-upgrade.md
```diff
@@ -0,0 +1,76 @@
+## Upgrading from Jasmine 1.3 to 2.0
+
+First, please read the official upgrade document: http://jasmine.github.io/2.0/upgrading.html
```
juliemr: `First, please read [Jasmine's official upgrade documentation](http://jasmine.github.io/2.0/upgrading.html).`

docs/jasmine-upgrade.md
```diff
@@ -0,0 +1,76 @@
...
+
+#### Focused specs
```
juliemr: Add a section on what you actually have to change in your Protractor configuration file.

docs/jasmine-upgrade.md
```diff
@@ -0,0 +1,76 @@
...
+Now:
+```javascript
+it("async spec", function(done) {
+  setTimeout(function() {
+    if (passed) {
+      done(); // When done
+    } else {
+      done.fail('failure message'); // To fail spec
+        // or
+      this.fail('failure message'); // Also fails spec
```
juliemr: I'm surprised that `this` works here, since it's inside a function. Should it be using the `self = this` trick?

docs/jasmine-upgrade.md
```diff
@@ -0,0 +1,76 @@
+## Upgrading from Jasmine 1.3 to 2.0
+
+First, please read the official upgrade document: http://jasmine.github.io/2.0/upgrading.html
+
```
juliemr: Add a note about how minijasminenode provided asynchronous support before, so the upgrade isn't as large as it is from unmodified Jasmine 1.3 to Jasmine 2.0

docs/jasmine-upgrade.md
```diff
@@ -0,0 +1,76 @@
...
+}
+```
+
+#### Asynchronous specs
+
+You can still pass in the done parameter as part of your asynchronous spec, but the syntax for failing it has changed.
+
+Before:
+```javascript
+it("async spec", function(done) {
```
juliemr: nit: should be single quotes

docs/jasmine-upgrade.md
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+      done.fail('failure message'); // To fail spec
+        // or
+      this.fail('failure message'); // Also fails spec
+    }
+  }, 5000);
+});
+```
+
+#### Spies and clock
```
juliemr: I'm not sure these are common or helpful in e2e tests - I think I'd just remove this section.

juliemr:
We should link to these docs from the table of contents, website menus, and frameworks doc - that could be a separate PR, up to you.

docs/jasmine-upgrade.md
hankduan: yes good catch. I didn't actually run the example, and it probably won't work. 
I'm just going to remove the reference to `this.fail` since it feels more consistent to use `done.fail` anyway

hankduan:
Added links to toc, website, and framework doc. Comments addressed.

PR #1690, opened 2015-01-06
feat(jasmine2): add 'grep' option to jasmine2

hankduan:
Allow users to filter the specs that they want to run using simple string match.
To use this feature, either:
1) specify jasmineNodeOpts.grep in your conf.js file
  or
2) via commandline like "protractor conf.js --jasmineNodeOpts.grep='string to match'"

### DIFF
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -265,8 +265,13 @@ exports.config = {
     showColors: true,
     // Default time to wait in ms before a test fails.
     defaultTimeoutInterval: 30000,
-    // Function called to print jasmine results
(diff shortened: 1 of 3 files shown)


juliemr:
LGTM

hankduan:
Can you take a quick look again? I added an invert option

juliemr:
As we discussed offline let's make it a regexp to be consistent with mocha, then LGTM

hankduan:
committed with https://github.com/angular/protractor/commit/0b930035905d1868225667de358222e51394f3ac

PR #1691, opened 2015-01-06
docs(mobile-setup): split docs on mobile setup into their own file

### LINKED ISSUE #1673: Docs: Mobile should have its own section
It's difficult to find within 'setting up the browser' right now, and it will probably only grow.

juliemr:
Closes #1673


juliemr:
@hankduan can you take a look?

Issue #1508, opened 2014-11-07
Expose currently running suite name, e.g. `protractor.currentSuite`


elgalu:
Useful to know the currently running suite name if any, should be `undefined` if none.
Should work not only when CLI param `--suite` is provided but when omitted too as long as there is a `suites` config.
Should also work when multiple suites are specified to the CLI, e.g. `--suite=smoke,fast,full`

Use case for me is a plugin I'm working on, also for CI integrations.

juliemr:
Can you clarify what behavior you'd like when there's no `--suite` parameter? Would it then be all the suites specified in the config?

elgalu:
I was thinking that `protractor.currentSuite` should have only 1 value at a current time, depending of which test is running, if that test/spec file belongs to a suite then set `protractor.currentSuite` to the suite name else `undefined`.
When `--suite` parameter is set to 1 value `protractor.currentSuite` is easy, simple assigns to argv.suite.
When `--suite` is set to multiple values or omitted then it gets trickier because the current value of `protractor.currentSuite` depends on pattern matching the current test file name with the suites listed in the configuration.

Perhaps I'm wishing suites to be a first class citizen, meaning we attach more info to it, for example to do other things like providing a summary of test results at suite level:

```sh
  Summary of run suites:
    ch_camp_enUS - ✔
    ch_packages_enUS - ✔
    ch_paste_enUS - ✔
    ch_smoke_enUS - ✔
    ch_loginvalid_enUS - ✔
    ie10_camp_enUS - skipped
    ie10_packages_enUS - ✔
    ie10_paste_enUS - skipped
    ie10_smoke_enUS - FAILED!
```

hhaamm:
@juliemr Please note that I'm currently working on a PR for this issue. Just saying to avoid duplicating effort :)

hankduan:
Can you elaborate on your use case? Do you just need to know from the runner, or do your tests need to know this information?

If you're trying to map each file name and `config.suite` against `config.suites`, there is some complexity in that jasmine runner does not run each file separately, but globbed together. This means protractor cannot determine which file is currently being run. (Of course, we can manually invoke jasmine multiple times to get the files to run separately, but before doing that, if you can explain your use case, it'd be great).

elgalu:
> Do you just need to know from the runner, or do your tests need to know this information?

My tests also need to know

> This means protractor cannot determine which file is currently being run

[...]

PR #1692, opened 2015-01-07
 Expose currently running suite name, e.g. `protractor.currentSuite` #1508 

hhaamm:
This change fixes issue #1508 for Jasmine 1.3. But it will be needed further work for making it work with Jasmine2 and the other testing frameworks.


lib/protractor.js
```diff
@@ -110,6 +110,13 @@ var Protractor = function(webdriverInstance, opt_baseUrl, opt_rootElement) {
   this.driver = webdriverInstance;
 
   /**
+   * When running a test suite, this variable should have the name of the suite.
+   *
+   * @type {String}
+   */
+  this.currentSuite = undefined;
```
hankduan: This was never set

hankduan:
Why do you need to add so much complexity into the path?
Unless I'm missing something, all you need to do is have one line that sets `protractor.currentSuite` or `browser.currentSuite` as `config.suite` in the runner right?

lib/protractor.js
hhaamm: @hankduan Sorry, the line that sets that variable is on minijasminenode. It seems I didn't commit that because it's ignored by .gitignore. I've just realized that it's a different project, so this solution is flawed.

hhaamm:
@hankduan It's possible to run more than one suite at once (and probably two test suites can have the same test). So that variable can only be set when running test by test, I think.

The code in minijasminenode was something like this:

    for (var i = 0, len = specFiles.length; i < len; ++i) {
      ...
      // If there is no protractor defined we are running a test suite against minijn    
      if (typeof protractor !== 'undefined') {
        protractor.currentSuite = specInfo.suite;
      } 
      require(path.resolve(process.cwd(), specInfo.spec));
      ...

hankduan:
Closing as this does not work

Issue #1600, opened 2014-12-05
ElementFinders should work as params in executeScript calls


gologox:
The following code causes infinite loop on the selector, i.e : 
12:15:11.418 INFO - Executing: [find elements: By.selector: p.note])
12:15:11.424 INFO - Done: [find elements: By.selector: p.note]

``` javascript
it('should scroll into an unhindering page view of all checkboxes', function(done) {
    var bottomElement = element.all(by.css('p.note')).last().then(function(element) {
      return element;
    });
      browser.driver.executeScript("arguments[0].scrollIntoView(true);", bottomElement).then(function() {
        done();
      });
  });
```

or

``` javascript
it('should scroll into an unhindering page view of all checkboxes', function(done) {
    element.all(by.css('p.note')).last().then(function(element) {
      browser.driver.executeScript("arguments[0].scrollIntoView(true);", element).then(function() {
        done();
      });
    });
  });
```

or 

``` javascript
function scrollIntoView(element) {
  arguments[0].scrollIntoView();
}

it('should scroll into an unhindering page view of all checkboxes', function(done) {
    element.all(by.css('p.note')).last().then(function(element) {
      browser.driver.executeScript(scrollIntoView, element).then(function() {
        done();
      });
    });
  });
```

**_browser.executeScript()**_ does the same thing. If I remove the execute script, the code no longer loops infinitely through the selector. 

**_Note: This was working fine prior to the 1.4.0 update from 1.2.0~**_

Any ideas?

juliemr:
@mgiambalvo I'm not sure if this works now or not, can you check and let's figure out a workaround if not.

mgiambalvo:
It works now!

Issue #1315, opened 2014-09-11
Have element-explorer use embedded chrome driver


FrankyBoy:
I really like the option to use the embedded chrome driver for protractor. It would be very convenient to be able to use the same in the element-explorer.

hankduan:
After #1693, you can enter element explorer via flag `--elementExplorer`. You may use the embedded chrome driver by specifying either a conf or options like you would normally do when running protractor, for example:
  `protractor --elementExplorer --directConnect` 
or
  `protractor --elementExplorer conf.js`.

FrankyBoy:
this is seriously neat!

hankduan:
Done at https://github.com/angular/protractor/pull/1693#issuecomment-69835756

PR #1693, opened 2015-01-07
feat(elementExplorer): Combine browser.pause with elementExplorer

### LINKED ISSUE #1314: Add command for element explorer
Hi!
Please add a command for the element-explorer. Its a very useful tool, but having to do "node [huge-path-to-js]" is very inconvenient as well as hiding the (again: very useful) tool from the users.
Ty :)

PS: sorry if there is already a defect for this, didn't find any.

hankduan:
* reuse logic for browser.pause for elementExplorer
* introduce browser.enterRepl
* allow customization of driver for elementExplorer
* fix bug where repl cannot return an ElementFinder (related #1600)

Closes #1314, #1315


lib/cli.js
```diff
@@ -123,8 +129,10 @@ if (!configFile) {
     configFile = './protractor.conf.js';
   }
 }
-if (!configFile && args.length < 3) {
-  optimist.showHelp();
+
+if (!configFile && !argv.elementExplorer && args.length < 3) {
+  console.log('you must either specify a configuration file ' + 
+    'or at least 3 options. See "protractor --help" for more info.');
```
juliemr: Let's show help here automatically as well.

lib/debugger/clients/wdrepl.js
```diff
@@ -0,0 +1,92 @@
+var repl = require('repl');
```
juliemr: Maybe we could rename this file - it feels like we're overloading 'repl' by having it both be a client and a mode.

lib/frameworks/repl.js
```diff
@@ -0,0 +1,22 @@
+var q = require('q');
```
juliemr: Ditto with file renaming - what about calling this mode 'explorer'?

lib/launcher.js
```diff
@@ -153,6 +154,26 @@ var init = function(configFile, additionalConfig) {
       }
     });
   }).then(function() {
+    // 3) If we're in `elementExplorer` mode, run only that.
+    if (config.elementExplorer) {
+      if (config.multiCapabilities.length != 1) {
+        throw new Error('Must specify only 1 browser while using elementExplorer');
```
juliemr: We should probably throw this warning if someone does something silly like manually set the framework to 'repl' (or 'explorer') but specifies multiCapabilities.

juliemr:
Awesome, looks good overall, just a couple clarification comments.

juliemr:
And I assume a following PR will change the docs and the element explorer binary?

lib/launcher.js
hankduan: The steps above (1 and 2) resolve getMultiCapabilities/capabilities/multiCapabilities into just multiCapabilities. So someone can manually set the framework to `repl`/`explorer` and one of those capabilities config, but it should be fine as long as there is only 1 capability being resolved no?

juliemr: I was thinking that someone will have a bad time if their config is:

```
multiCapabilities: [one browser, second browser],
framework: explorer
```

hankduan:
I'll send a follow up PR to change the docs. I was thinking of removing the element explorer binary since you can just do 'protractor --elementExplorer'. I can keep the binary too, but it would just be a one-liner. What do you think?

juliemr:
Let's keep the one liner binary for discoverability.

lib/launcher.js
hankduan: But as is, 
```
if (config.elementExplorer) {
      if (config.multiCapabilities.length != 1) {`
```

would already throw an error for that case right?

juliemr: But they don't have elementExplorer set, just framework.

hankduan: Ahh, I see what you mean. I did mark the explorer framework as private so people shouldn't be messing around with it...but I'll change the conditional to safeguard it too.

hankduan:
Addressed comments.

juliemr:
Great - let me do one more run-through and actually try it out on my computer and then I'll LGTM.

hankduan:
Docs: https://github.com/angular/protractor/pull/1717

bin/elementexplorer.js
```diff
@@ -32,124 +32,15 @@
...
-    util.puts('Getting page at: ' + url);
-    driver.get(url);
-
-    startRepl();
-  });
-};
-
-startUp();
+console.log('Please use "protractor --elementExplorer [options] [configFile]"' + 
+  ' for full functionality\n');
```
juliemr: Can you change this to 

`Please use protractor [configFile] [options] --elementExplorer`

The order can be very confusing if, for example, you run

`protractor --elementExplorer myConfig.js`

This will actually ignore myConfig.js because optimist interprets it as the value of `elementExplorer`.

bin/elementexplorer.js
```diff
@@ -32,124 +32,15 @@
...
+
+if (process.argv.length > 3) {
+  console.log('usage: elementexplorer.js [urL]');
```
juliemr: Can you remove since this is no longer valid (url won't be used)

juliemr: I guess, it will be used, but it's just added as the base url, not grabbed automatically. Since we're breaking this functionality, I'd say let's not bother with changing the baseUrl either.

hankduan: the explorer will automatically navigate to the baseurl on initialization.

bin/elementexplorer.js
hankduan: fixed.

bin/elementexplorer.js
juliemr: Gotcha. I guess this will fail if baseUrl isn't a valid Angular site, but I'm not sure which behavior is better here.

hankduan: Same thing if you do `browser.get('non-angular-site')` right?

juliemr: True, but I could imagine using a base Url which isn't a real webpage, since it's just a string prefix. I think it's OK the way it is though, it should be obvious what the error is.

juliemr:
LGTM

hankduan:
merged with https://github.com/angular/protractor/commit/fb099dedf92a64732d88401dd1b0d4d30b22650d

tb:
How can I can access test scenario varibles in `browser.enterRepl()`? In particular I am interested to have access to libs/helpers/page objects loded via require`HttpBackend = require('http-backend-proxy')`

hankduan:
@tb I agree it sounds like a good feature to have. However, it is not possible right now since Protractor is running user code (repl code) in a vm. It might be possible to inject relevant helpers/librarys into the vm's context during the set up, but it would look something like `context = {foo: 3, bar: function() {}}; browser.enterRepl(context)`; in other words, a user would need to pass those in. If this is something of use to you, please open a new feature request issue and we can track there.

PR #1697, opened 2015-01-07
fix(element): Improve error reporting

isaaclyman:
I often get "Index out of bound" errors, explaining that there are 0 elements in an "element.all" array, when a more correct/consistent error would be "No element found using locator [Locator]". This is a quick fix. I may have a mis-scoped object or two in my "throw" statement. If so, I'd appreciate a push in the right direction.

Isaac Lyman

### DIFF
--- a/lib/element.js
+++ b/lib/element.js
@@ -257,16 +257,10 @@ ElementArrayFinder.prototype.get = function(index) {
         i = parentWebElements.length + i;
       }
       if (i < 0 || i >= parentWebElements.length) {
-        if (parentWebElements.length == 0) {
-          throw new webdriver.error.Error(
-              webdriver.error.ErrorCode.NO_SUCH_ELEMENT,
-              'No element found using locator: ' +
-                  elementArrayFinder.locator_.toString());
-        } else {
         throw new Error('Index out of bound. Trying to access element at ' +
             'index: ' + index + ', but there are only ' +
-            parentWebElements.length + ' elements');
-        }
+            parentWebElements.length + ' elements that match locator ' +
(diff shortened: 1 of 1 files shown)


isaaclyman:
*Signed the CLA.

hankduan:
When do you see this issue? Like this?

```javascript
  it('test', function() {
    browser.get('index.html');
    var el = element.all(by.model('WRONG'));
    el.get(1).click();
  });
```

juliemr:
I prefer the error the way it is now - special casing for when there's 0 elements seems odd here. The current error tells you exactly what the problem is, and lets you know that you're dealing with a (potential) list of elements.

juliemr:
What about if we just edit the error message to always include information about the locator?

isaaclyman:
@juliemr I could go with that. My assumption is that if there are 0 elements, then the locator didn't work at all, in which case there is no list and a get() will fail regardless of the index. If I understand you correctly, the advantage of the current error message is that it indicates the usage of element.all() instead of just element()?

I don't feel like information about the locator is totally necessary if you're trying to access index 5 of an element array with only 4 elements--in that case, the error makes total sense, and the locator is probably fine.

It does occur to me that maybe someone would use element.all with a program flow that determines what to do next based on if the resulting array is empty or not, and loops for each element in the array. In which case an "index out of bounds" error might be more meaningful to them than an "element not found" error.

isaaclyman:
The latest commit just includes locator information no matter what.

juliemr:
Thanks! Merged as 2a288881fb57d7bb380f8568011e0da3a9d3a33a

PR #1703, opened 2015-01-09
feat(expectedConditions): add helper library for syncing with non-angula...

hankduan:
...r apps

### DIFF
--- a/lib/expectedConditions.js
+++ b/lib/expectedConditions.js
@@ -1,0 +1,293 @@
+var webdriver = require('selenium-webdriver');
+
+/* globals browser */
+
+/**
+ * Represents a library of canned expected conditions that are useful for
+ * protractor, especially when dealing with non-angular apps.
+ * 
+ * Each condition returns a function that evaluates to a promise. You may mix 
+ * multiple conditions using `and`, `or`, and/or `not`. You may also
+ * mix these conditions with any other conditions that you write.
+ *
+ * See https://selenium.googlecode.com/git/docs/api/java/org/openqa ...
+ *     /selenium/support/ui/ExpectedConditions.html 
+ *
+ *
+ * @example
+ * var EC = protractor.ExpectedConditions;
+ * var button = $('#xyz');
+ * var isClickable = EC.elementToBeClickable(button);
+ *
+ * browser.get(URL);
+ * browser.wait(isClickable, 5000); //wait for an element to become clickable
+ * button.click();
+ *
+ * // You can defined your own expected condition, which is a function that
+ * // takes no parameter and evaluates to a promise of a boolean. 
+ * var urlChanged = function() {
+ *   return browser.getCurrentUrl().then(function(url) {
+ *     return url === 'http://www.angularjs.org';
+ *   });
+ * };
+ *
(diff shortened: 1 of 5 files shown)


hankduan:
usage:
```javascript
var EC = protractor.ExpectedConditions;
var button = $('#xyz');
var isClickable = EC.elementToBeClickable(button);

browser.get(URL);
browser.wait(isClickable, 5000); //wait for an element to become clickable
button.click();
```

You can also customize the conditions:

```javascript
var urlChanged = function() {
  return browser.getCurrentUrl().then(function(url) {
    return url != 'http://www.angularjs.org';
  });
};

var condition = EC.and(urlChanged, EC.titleContains('foo'), EC.textToBePresentInElement($('def'), 'bar'));
$('navButton').click();
browser.wait(condition, 5000); //wait for an element to become clickable
// do other things
```

juliemr:
Can you add the usage notes to the commit message?

spec/basic/expected_conditions_spec.js
```diff
@@ -0,0 +1,167 @@
+var EC = protractor.ExpectedConditions;
+
+describe('expected conditions', function() {
+  beforeEach(function() {
+    browser.get('index.html#/form');
+  });
+
+  it('should have alertIsPresent', function() {
+    var alertIsPresent = EC.alertIsPresent();
+    expect(alertIsPresent.call()).toBe(false);
```
juliemr: I might be being dense, but why not:

`expect(alertIsPresent()).toBe(false);`?

lib/expectedConditions.js
```diff
@@ -0,0 +1,267 @@
+var webdriver = require('selenium-webdriver');
+
+/* globals browser */
+
+/**
+ * Represents a library of canned expected conditions that are useful for
+ * protractor, especially when dealing with non-angular apps.
+ *
```
juliemr: Add some example usage here as well (especially using in the context of `browser.wait`)

spec/basic/expected_conditions_spec.js
hankduan: Yea that works too. It's just that in cases where I didn't use an extra variable `expect(EC.alertIsPresent().call())` looked a little more pleasing than `expect(EC.alertIsPresent()())`, but functionally they're the same.

hankduan: Alternatively, EC.func can return a promise instead of a function that returns a promise. The down side is that then wait conditions (which I think will be the primary use case will look like `browser.wait(function(){EC.func(x,y)}, 100)` instead of `browser.wait(EC.func(x,y), 100)`

juliemr: I agree that we should optimize for being used in `browser.wait`. 

I'm fine with either `call()` or just `()`, so you can leave as is.

lib/expectedConditions.js
hankduan: done

hankduan:
Added usage examples

juliemr:
LGTM!

hankduan:
committed at https://github.com/angular/protractor/commit/9bc1c53e40161521b0c125a810f86235c974f100

JemiloII:
How do we test for non clickable with this?

Issue #1594, opened 2014-12-04
Feature Request: Allow protractor to use promise as capabilities for multiCapabilities


hankduan:
related:
http://stackoverflow.com/questions/27264146/setting-firefox-profile-in-a-setup-with-multiple-browsers
http://stackoverflow.com/questions/26435438/creating-another-session-with-a-separate-firefox-profile-in-protractor
https://github.com/angular/protractor/pull/1040
https://github.com/angular/protractor/pull/1592

Right now, protractor can only accept promise as capabilities if we are NOT using multicapabilities. The reason for this is because multiCapabilities runs each task in a new process, so the promise (function) cannot be passed (single capabilities work because we're not forking). Alternatively we could resolve capabilities in the launcher, before passing the resolved capabilities into the new processes; however, this will break https://github.com/angular/protractor/pull/1040, which relies on capability promises to be resolved after driverProvider setup (i.e. for proxy purposes). 

I can't think of an easy way of doing this (without large refactoring), but it is definitely doable. Just creating this issue to get a feel of how many people would benefit from being able to use promises in multiCapabilities and/or collect ideas.

Issue #1354, opened 2014-09-22
Chrome preferences for browserstack


alecxe:
I'm experiencing a weird issue and I'm not sure on which side things go wrong.

I have a couple of tests with disabled javascript and disabled cookies. In order to disable js, you need to pass `--disable-javascript` argument in the chrome arguments; in oder to disable cookies, you need to set  `profile.default_content_settings.cookies` preference to `2`. This is proven to work for me.

The problem is that setting `prefs` object inside the `chromeOptions` doesn't work for browserstack:

``` javascript
capabilities: {
    'browserstack.user' : browserstackUser,
    'browserstack.key' : browserstackKey,
    'browserstack.local' : 'true',
    'browserstack.debug': 'true',

    'browserName' : 'chrome',
    'os' : 'OS X',
    'os_version' : 'Mavericks',
    'resolution' : '1024x768',

    chromeOptions: {
        args: [
            '--disable-javascript'
        ],
        prefs: {
            'profile.default_content_settings.cookies': 2
        }
    },

    specs: [
        'disabledCookies/*.js'
    ]
},
```

In this case javascript would get disabled, but cookies would be still enabled.

FYI, running the tests locally (changing `seleniumAddress` and omitting browserstack-specific keys) works perfectly.

Is it something on the protractor's side, or the problem is browserstack-specific? Thank you.

UmangSardesai:
This works for me. Just checked on BrowserStack. 

```
var capabilities = {
  'browserName' : 'chrome',
  'browser_version' : '46.0',
  'browserstack.user' : '<username>',
  'browserstack.key' : '<automate-key>', 
  'chromeOptions': {
        prefs: {
          profile: {
            default_content_settings: {
              cookies: 2
            }
          }
        }
    } 
}
```

alecxe:
@UmangSardesai thank you, I'll try that!

Issue #1195, opened 2014-08-14
Feature: Distinct Junit-Results with multiCapabilities


zakrhol:
When setting up multiCapabilities and adding a jasmine-junitReporter, then specs running for all capabilities produces the same junit-result.xml file, where one replaces the other.

I have read in [#60] (https://github.com/angular/protractor/issues/60#issuecomment-32863940) that it is possible to change the filename of the result for each capability. But as far as I have tried, that is not always sufficient.
 *  Although the name of the junit-xml was changed, the name of the test-suites and test-classes remain equal and a CI-System like Jenkins is not able to display the results properly. I have tested this: when jenkins faces the same suite-name again, it is simply ignored - if the first capability passes and the second not, the test shows as passed. This may be a bug in Jenkins, but even if it shows all tests I cannot see in which capability it has failed.
 * Changing the name by adding browsername and version as shown in [#60] (https://github.com/angular/protractor/issues/60#issuecomment-32863940) is also not enough for my cases: I always use the same browser but only change browser size to test for Table/SmartPhone resolutions (Bootstrap). Unfortunately `browser.getCapabilities();` does not contain something useful (and readable) to distinguish the capabilities.

To get it working properly I made the following changes and I want to ask whether that makes sense or can be solved better.
I give each capability a name. To access the capabilities in the `onPrepare` function (`browser.getCapabilities();` only has driver/browser stuff), I changed runner.js
```javascript
Runner.prototype.runFilenamesOrFns_ = function(source) {
...
if (typeof filenameOrFn === 'function') {
        filenameOrFn();
```
to 
```javascript
Runner.prototype.runFilenamesOrFns_ = function(source) {
...
if (typeof filenameOrFn === 'function') {
        filenameOrFn(this.config_);
```
In the `onPrepare`  the capability-name is appended to the suite and spec name and also the junit.xml filename.
My config.js now looks like this (simplified):
```javascript
specs: [
      'test/protractorTests/all/**/*.js'
  ],

  multiCapabilities: [{
	    'browserName': 'chrome',
		'chromeOptions' : {
			args: ['--lang=en',
			       '--window-size=800,800']
		},
		name: '[md]',
		specs: 'test/protractorTests/md/**/*.js'
	  },{
	    'browserName': 'chrome',
		'chromeOptions' : {
			args: ['--lang=en',
			       '--window-size=350,650']
		},
		name: '[xs]',
		specs: 'test/protractorTests/xs/**/*.js'
	  }],
[...]

PR #1708, opened 2015-01-10
Pass the current runner config into the onPrepare hook

crrobinson14:
Currently the `onPrepare()` hook has no arguments passed into it. This means that, even though it is called once per capability, it has no idea which capability is being executed so it can't make context-sensitive decisions. This PR passes that configuration block into `onPrepare()` so it can do that, like including requirements that only certain tests need (for performance reasons, or where only certain environments require/support them).

Since `runFilenameOrFn_()` already supports the concept of passing arguments to the callback, this only requires a 1-line code change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+# 1.6.1
+
+## Bug Fixes
+
+- ([92c5d17](https://github.com/angular/protractor/commit/92c5d17844a2b4dc56c483ab4a65e2bf631175f9)) 
+  fix(element): test crashes when using certain locators with `fromWebElement_`
+
+  Protractor crashes when one uses locators with findElementsOverride (i.e. any custom protractor
+  locator like by.binding/repeater/etc) in map/filter/then/each/reduce
+
 # 1.6.0
 
 ## Features
--- a/docs/frameworks.md
+++ b/docs/frameworks.md
@@ -13,7 +13,7 @@ Currently, Jasmine Versions 1.3 and 2.0 are supported. Jasmine 1.3 is the defaul
 Using Mocha
 -----------
 
-_Note: Limited support for Mocha is available as of December 2013. For more information, see the [Mocha GitHub site](http://visionmedia.github.io/mocha/)._
+_Note: Limited support for Mocha is available as of December 2013. For more information, see the [Mocha documentation site](http://mochajs.org/)._
 
 If you would like to use the Mocha test framework, you'll need to use the BDD interface and Chai assertions with [Chai As Promised](http://chaijs.com/plugins/chai-as-promised).
 
--- a/docs/jasmine-upgrade.md
+++ b/docs/jasmine-upgrade.md
@@ -1,4 +1,5 @@
-## Upgrading from Jasmine 1.3 to 2.x
+Upgrading from Jasmine 1.3 to 2.x
+=================================
 
 First, please read [Jasmine's official upgrade documentation](http://jasmine.github.io/2.0/upgrading.html).
 
--- a/docs/jasmine-upgrade.md
+++ b/docs/jasmine-upgrade.md
@@ -14,6 +14,26 @@ exports.config = {
 
 ```
 
+Similar to jasmine 1.3, you may include `jasmineNodeOpts` in the config file. However, because we changed the runner from "https://github.com/juliemr/minijasminenode" to "https://github.com/jasmine/jasmine-npm", the options have changed slightly. In particular, we will only support the options in the official "jasmine-npm":
+
+```javascript
+jasmineNodeOpts: {
+  // If true, print colors to the terminal.
+  showColors: true,
+  // Default time to wait in ms before a test fails.
+  defaultTimeoutInterval: 30000,
+  // Function called to print jasmine results.
+  print: function() {},
+  // If set, only execute specs whose names match the pattern, which is
+  // internally compiled to a RegExp.
+  grep: 'pattern',
+  // Inverts 'grep' matches
+  invertGrep: false
+}
+```
+
+Notably options `print` and `grep` are new, but we will no longer support options `isVerbose` and `includeStackTrace` (unless, of course, "jasmine-npm" introduces these options).
+
 ### In your specs
 
 #### Focused specs
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -276,7 +276,7 @@ exports.config = {
 
   // Options to be passed to Mocha.
   //
-  // See the full list at http://visionmedia.github.io/mocha/
+  // See the full list at http://mochajs.org/
   mochaOpts: {
     ui: 'bdd',
     reporter: 'list'
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -193,6 +193,14 @@ exports.config = {
     // are using Jasmine, you can add a reporter with:
     //     jasmine.getEnv().addReporter(new jasmine.JUnitXmlReporter(
     //         'outputdir/', true, true));
+    //
+    // As a convenience, if you need access back to the current configuration
+    // object, you may use a pattern like the following:
+    //     browser.getProcessedConfig().then(function(config) {
+    //       // config.capabilities is the CURRENT capability being run, if
+    //       // you are using multiCapabilities.
+    //       console.log('Executing capability', config.capabilities);
+    //     });
   },
 
   // A callback function called once tests are finished.
--- a/lib/element.js
+++ b/lib/element.js
@@ -258,8 +258,9 @@ ElementArrayFinder.prototype.get = function(index) {
       }
       if (i < 0 || i >= parentWebElements.length) {
         throw new Error('Index out of bound. Trying to access element at ' +
-            'index:' + index + ', but there are only ' +
-            parentWebElements.length + ' elements');
+            'index: ' + index + ', but there are only ' +
+            parentWebElements.length + ' elements that match locator ' +
+            elementArrayFinder.locator_.toString());
       }
       return [parentWebElements[i]];
     });
--- a/lib/element.js
+++ b/lib/element.js
@@ -207,7 +207,7 @@ ElementArrayFinder.prototype.filter = function(filterFn) {
       var list = [];
       parentWebElements.forEach(function(parentWebElement, index) {
         var elementFinder = 
-            ElementFinder.fromWebElement_(parentWebElement, self.locator_);
+            ElementFinder.fromWebElement_(self.ptor_, parentWebElement, self.locator_);
 
         var filterResults = filterFn(elementFinder, index);
         if (filterResults instanceof webdriver.promise.Promise) {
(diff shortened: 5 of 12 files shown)


crrobinson14:
Signed CLA.

sjelin:
If you're going to pass anything like this, you should be passing `this.config_.capabilities`, not the whole `config` object.

Also, I don't like the idea of exposing internal variables like that regardless.  This seems like a bit of an edge case, which you could always resolve (albeit inelegantly) by having multiple config files.

crrobinson14:
I'm not sure I understand the concern. These aren't "internal" variables - they came FROM the very same config that onPrepare() is being executed in. It's not passing `this` (which has driverprovider and all sorts of things)... just `config`:

```javascript
var Runner = function(config) {    /// <------ I give you the config in the first place...
  log.set(config);
  this.preparer_ = null;
  this.driverprovider_ = null;
  this.config_ = config;                   /// <----- You're just giving it back as a convenience...
  // ...
};
```

The only vars in that block that don't come from the caller are:

```javascript
     // The CURRENT capabilities being run
    capabilities: {},
    // The command used to start Protractor
    '$0': 'node ./node_modules/protractor/bin/protractor',
    // An array of args used to start Protractor
    _: [ 'test-client/protractor.conf.js' ],
    // ... There may be an additional var or two depending on how the Selenium config was interpreted
```

I would still argue that these additional vars are useful. onPrepare()'s purpose is very well aligned with the need for these variables. To do anything context-sensitive when you aren't using `multiCapabilities` you need to take an internal reference to your own config object before you export it. It's just a sensible convenience to pass it back for that case. That's especially valuable when onPrepare is an external script - a nice feature added recently but it's super awkward to pass config data to it...

Finally, I would argue that multiple configs is a terrible answer. This whole `multiCapabilities` feature must have been added for a reason, but there are half a dozen issues posted over the last year asking how to do things with it where the answer was "sorry, you can't, use multiple configs". When it comes down to the trade-offs, what advantage does Protractor gain by NOT making this easier in exchange for avoiding the perceived risk of... making this easier?

juliemr:
In general, we try to avoid feature creep in the configuration file. `multiCapabilities` was added because it was highly requested, addresses a common use case, and provides a lot of benefits. On the other hand, it has significant costs, as does anything that complicates the config. The idea of multiCapabilities is that you want to run the _same setup_ on a couple different browsers. We don't want, and cannot support, one config file being able to support all cases. If what you're requesting is easy to do with multiple config files, that's a great solution!

For this specific request - you can get at the current capabilities using `browser.getCapabilities` (returns a promise). Could you let us know if this doesn't cover your use case? If so, what other context are you looking for?

If we do introduce a way to get the config (which is not out of the question), I'd like it to be more general than being passed as a parameter to onPrepare functions.

crrobinson14:
Well... this is certainly disheartening. :(

First, it's not one-browser-many-cases. It's one-case-many-browsers. But browser.getCapabilities() doesn't address it because it returns Selenium's block, not Protractor's. Suppose you wanted to use a configuration similar to the one discussed in this issue:

https://github.com/angular/protractor/issues/1354

Suppose you wanted your screenshot tool at the end of each test to drop the file in a directory whose name contained the resolution being captured. There's no way to do this currently - browser.getCapabilities will give you the browser name and version back, but not the other options. But with a single additional option in the capabilities block you could pass a parameter to each test that was specific to that block. This makes it a single line of code to marry these up.

Over the past year it seems like this kind of thing has been repeatedly requested:

https://github.com/angular/protractor/issues/1594
("I would definitely benefit from this...")

https://github.com/angular/protractor/issues/1195
``var capsPromise = browser.getCapabilities(); //does not contain useful stuff```

http://stackoverflow.com/questions/22305297/run-protractor-tests-with-different-window-sizes

etc. Various requests to be able to pass in dynamic elements to a test like user/pass, test data, browser resolutions, etc. are all over StackOverflow and here.

config.params is messy because it's hard-coded. onPrepare is given as the specific answer for how to do things like set resolution in a number of places as well, e.g. here:

http://ramonvictor.github.io/protractor/slides/#/34

I'm not saying Ramon is the de-facto source for this information... but there certainly isn't anything to contradict it elsewhere...

Multiple configs is a bad answer because 95% of my protractor config is identical between each test - it's only the elements in each capabilities block that changes. (It's the same app... just across a range of browsers and resolutions...) Having to duplicate 200 lines of configuration data just to change ONE line of "put THESE screen shots in THIS folder" is a huge maintenance hassle.

Finally, it's disheartening because this was the first in a collection of contributions I had planned to make. I thought it was perfectly in line with a testing framework designed specifically for a JS framework that values declarative definitions, dependency injection, DRYness, etc. Discovering that the TESTING framework prefers basically the opposite - inferential connections between items, significant repetition, and call-for rather than injecting dependencies makes me wonder if the rest of what I had planned to suggest was misguided.

Is there a document somewhere that outlines the design principles behind this project? I have what I believe are some valuable ideas, but I don't want to be a nuisance... and I certainly don't want to repeatedly trigger 8-hour discussion about 8-byte code changes! :) I'm happy to comply with whatever those principles are... but where can I find them outlined?

juliemr:
I don't mean to be disheartening, and I definitely don't want to discourage you from submitting PRs! We just want to carefully consider the complexity of changes, guard against making breaking changes in the future, and make sure that PRs solve real use cases. There is no doc with design principles specifically laid out. This change is small in bytes, but conceptually it's a big problem ('what information should the test be able to use about the configuration'). I think that you bring up some legitimate concerns - let's work together to find a good solution. 

FYI re multiple configs: Since configs are node objects, you can create a base config and `require` it to share code. This is what the angular.js project does.

@hankduan what do you think about adding a `browser.getProcessedConfig` (name could be better) function in `Runner.prototype.createBrowser`? This would give the config post TaskScheduler/TaskRunner processing so that the specs and capabilities line up with what's actually being run by this one browser instance.

hankduan:
@crrobinson14 
From the examples you provided, it seems like you don't actually need the entire config. Rather, you want to do use protractor to help you distribute custom capability-specific configs like this?

```
multiCapabilities: [{
    'browserName': 'chrome',
    count, maxInstances, specs, shardTestFiles ... (configs used by protractor/webdriver),
    resolution, browser-specific data, ... (configs not used by protractor/webdriver)
  }],
```

and use the custom configs in your onPrepare to customize your test:
```
onPrepare: function(capability) {
  browser.resolution = capability.resolution;
  jasmine.on_fail = //take screenshot at "capability.browserName" + "capability.resolution"
}
```

crrobinson14:
Thanks, all, for bearing with me, especially @juliemr for being so diplomatic!

Yes, @hankduan, that's roughly the goal, although I'm trying to be a little generic with my description because running multi-resolution tests is just one use-case. I can think of dozens of others when you get into enterprise-app space.

My position is that onPrepare is perfectly time for doing test-run-specific setup - that could mean configuring the browser's resolution, setting up some environment variables such as where some test outputs will be dropped, setting up a timing analytic, etc. These are things that are not ideally done before starting Protractor itself (e.g. environment variables) because of the added value of multiCapabilities. m/c is perfect for setting up a collection of browser combinations that you want to test against. Using Chrome + Firefox and adding Safari later or IE on a box that supports it is the obvious fit there, and it's the one the documents suggest.

But as soon as you say "oh, I can say I want BOTH Firefox and Chrome" the logical next step is to say "Chrome 1400x1050 and also Chrome 1024x768". This is a single line of code in onPrepare... as long as you know that this is the resolution desired. But resolution is not the only use-case - as I said above perhaps you could be testing i18n functionality and setting the browser's language. You can do that with the --lang=LANG flag in Chrome's arguments in the capabilities - but how do you tell your test to go to LANG.mydomain.com and that you want your test reports dumped out to build/reports/LANG? Yes, multiple configs are possible - but they're almost identical so it's a pain to maintain if you decide to switch to SauceLabs one day. On the other hand, a 1-line-of-code solution is possible if you make the 'current' capabilities config accessible somewhere onPrepare (and others?) can get to it. You just `browser.get('http://' + currentCapabilityConfig.lang + '.mydomain.com/');` in your test script, or `browser.takeScreenshot().then(....   // Write the PNG to 'build/screens/' + currentCapabilityConfig.lang + '/home-page.png')`

I also still think this would be very valuable when you start using Protractor's recently-added support for making these hooks call external scripts. You could have some very elegant and powerful configurations with very little work.

I could probably imagine several dozen cases where you might want to test some browser-specific aspect of your site/app, something whose only functional difference is a slight change in the capabilities config block.

I suppose the generic feature-request is to provide some way of enabling context-sensitive work to be performed inside hooks (and test scripts?) And I believe the single thing required to make all that happen is to have access to the current capability configuration object being executed. If you prefer not to inject it, perhaps a simple call to obtain it? IMO it's a little weird because there's already browser.getCapabilities() which suggests that it does this (but returns something else), but anything in this category would totally solve the problem for me.

Just brainstorming here... This may be just a horrific idea (I blame a sleepless night) but it seems like Angular's own dependency injection is a clean way to address a lot of these kinds of things. The provider/injector concept is already very familiar to Angular authors, and it seems like it wouldn't be much more than an else case and an array iterator (with calls to providers to obtain the things to inject) in the spot where hooks are being called. Totally understand if you hate the idea, but it does seem like another way to address this kind of problem in a relatively generic way that doesn't lead to hard-coding of hook arguments - and potentially even allows test authors/maintainers to create their own providers...

Or maybe just browser.getCapabiitiesConfig() and we're done here? :) Would you accept a PR that provided that?

juliemr:
I'm leaning towards `browser.get[...]Config` since that has the flexibility of also being available inside a test (where we couldn't use dependency injection, since we shouldn't change the arguments of the test `describe` and `it` functions). It sounds like that would cover all of your use cases. I'd take a PR to do that! 

I think a good place to add the function would be at https://github.com/angular/protractor/blob/master/lib/runner.js#L189

We should also add some documentation inside the reference config saying how it will be available.

@hankduan did you have an alternate idea?

hankduan:
No I think this is good, although it might be good for the config to be a bit more structured. (i.e. `browser.config.getCapabilities()`, `browser.config.getSpecs()`, because we switch those up a bit internally, so someone passing `multicapabilities` into the config might be surprised that their current capabilities is found in `capabilities`

juliemr:
Yeah, that's why I was thinking calling it `browser.getProcessedCapabilities` might hint that things have been processed :)

crrobinson14:
Let me know what naming convention you'd both prefer and I'd be happy to re-submit the PR around that!

hankduan:
It seems like @crrobinson14 is talking about wanting all configs, not just capabilities though. Is that right?

Alternatively we can make our config parser more featured, and parse all configs that we would use internally using the config parsers instead of accessing configs as key/value pairs. Then, we would just pass an instance of the config parser.

juliemr:
Bah I meant `browser.getProcessedConfig`.

Let's just pass it as an object and use that.

crrobinson14:
A lot of these methods return promises... but this method doesn't require it. Should it return one anyway (immediately resolved) just to follow the same convenient as browser.getCapabilities()?

crrobinson14:
Actually I'm going to close this PR while I fix this up. I borked my local repo reverting the original commit and don't want to keep triggering Travis. Back shortly.

PR #1718, opened 2015-01-13
bug(element): test crashes when using certain locators with `fromWebElem...

hankduan:
...ent_`

Protractor crashes when one uses locators with findElementsOverride (i.e.
any custom protractor locator like by.binding/repeater/etc) in
map/filter/then/each/reduce

### DIFF
--- a/lib/element.js
+++ b/lib/element.js
@@ -207,7 +207,7 @@ ElementArrayFinder.prototype.filter = function(filterFn) {
       var list = [];
       parentWebElements.forEach(function(parentWebElement, index) {
         var elementFinder = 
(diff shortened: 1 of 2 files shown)


juliemr:
Can you rename the commit to `fix(element)....`

juliemr:
Add a test that would have caught this issue, then LGTM.

juliemr:
LGTM my bad, I missed the test :)

hankduan:
merged: https://github.com/angular/protractor/commit/92c5d17844a2b4dc56c483ab4a65e2bf631175f9

Issue #1053, opened 2014-07-09
Smoke-test Node 0.11


mzgol:
Node 0.11.13 is technically unstable, but there are reports the stable 0.12 version is coming. It would be good to at least smoke-test on Node 0.11 to be able to catch simple regressions and be compatible with 0.12 once it's released.

See e.g. #1005 that makes Protractor Selenium installation impossible with Node 0.11 currently.

juliemr:
This is now done with 0.12

PR #1720, opened 2015-01-13
node 0.11 ...

cybernet:
run test with node 0.11 :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 node_js:
   - "0.10"
+  - "0.11"
 
 env:
   global:


juliemr:
See also https://github.com/angular/protractor/issues/1053

juliemr:
We should start adding some smoke tests against node .11 but I don't think to re-run our entire suite to catch errors (this will create more travis traffic for all Angular builds). Let's only test against a small suite.

@cybernet would you like to update this PR to do that?

sjelin:
@juliemr Tomorrow if you want to explain to me how to do this, I'll handle it

sjelin:
Continued over at https://github.com/angular/protractor/pull/1783

PR #1724, opened 2015-01-13
Provide access to the processed configuration block

crrobinson14:
This is a continuation of PR #1708 with a few changes:

1. Un-bork my fork (that OUGHT to be a song).
2. Use the maintainers' recommended pattern for the call.
3. On instinct I went with a promise return to be consistent getCapabilities() - let me know if that should change (a promise is not actually required here).
4. Added a comment block to referenceConfig.js to illustrate its use.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -188,6 +188,17 @@ Runner.prototype.createBrowser = function() {
   var self = this; 
 
   /**
+   * Get the processed configuration object that is currently being run.
+   *
+   * @return {q.Promise} A promise which resolves on finish.
+   */
+  browser_.getProcessedConfig = function() {
+    var deferred = q.defer();
+    deferred.resolve(config);
+    return deferred.promise;
+  };
+
+  /**
    * Fork another instance of protractor for use in interactive tests.
    *
    * @param {boolean} opt_useSameUrl Whether to navigate to current url on creation
--- a/spec/basic/processedconfig_spec.js
+++ b/spec/basic/processedconfig_spec.js
@@ -1,0 +1,11 @@
+var util = require('util');
+
+describe('configuration elements', function() {
+  it('should have access to the processed config block', function() {
+    browser.getProcessedConfig().then(function(config) {
+      expect(config.capabilities.browserName).toEqual('chrome');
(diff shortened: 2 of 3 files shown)


sjelin:
Looks good, maybe add a test though?  It's pretty simple, just put a jasmine/protractor test in `/spec/basic` and name it something ending in `_spec.js` and it run automatically.

crrobinson14:
I added a simple test as you suggested.

juliemr:
Looks good to me - I'm going to update the commit message to match with our pattern and merge.

spec/basic/processedconfig_spec.js
```diff
@@ -0,0 +1,11 @@
+var util = require('util');
+
+describe('configuration elements', function() {
+  it('should have access to the processed config block', function() {
+    browser.getProcessedConfig().then(function(config) {
```
juliemr: This will actually fail in our CI where the browser is not always Chrome - I'll modify slightly.

lib/runner.js
```diff
@@ -188,6 +188,17 @@ Runner.prototype.createBrowser = function() {
...
 
   /**
+   * Get the processed configuration object that is currently being run.
+   *
+   * @return {q.Promise} A promise which resolves on finish.
+   */
+  browser_.getProcessedConfig = function() {
+    var deferred = q.defer();
+    deferred.resolve(config);
+    return deferred.promise;
```
juliemr: You can just do

`return q(config);`

I'll switch it.

spec/basic/processedconfig_spec.js
```diff
@@ -0,0 +1,11 @@
...
+      expect(config.capabilities.browserName).toEqual('chrome');
+      expect(config.capabilities.version).toEqual('ANY');
+      expect(config.params.login.name).toEqual('Jane');
```
juliemr: Also a quick note - the test will not wait for this 'then' block (and so not execute these tests) since it's a q promise and not a webdriver promise. I'm resolving this by making `getProcessedConfig` return a webdriver promise, so that it will automatically be waited for.

crrobinson14:
Thanks for all the help!

PR #1741, opened 2015-01-20
feat(protractor): Change the default root element selector

RobJacobs:
Changes the default angular app element CSS selector from <body> to [ng-app] which will look for an element with the ng-app attribute.

### DIFF
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -160,8 +160,8 @@ exports.config = {
   baseUrl: 'http://localhost:9876',
 
   // CSS Selector for the element housing the angular app - this defaults to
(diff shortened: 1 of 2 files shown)


juliemr:
This would be a breaking change for some users. Anyone who uses another form of the directive, such as `ng:app`, or users with multiple `ng-app` elements, for example - so I don't think we can merge as-is. 

I do think that the rootElement setting can be confusing though, and we can probably do better than defaulting to `body`.  I started a discussion at https://github.com/angular/protractor/issues/1742

RobJacobs:
@juliemr Ok, thanks for considering this suggestion.

PR #1747, opened 2015-01-23
feat(frameworks): add support for custom frameworks

elgalu:
This way people can use their customized versions of a framework (forks) or write their own.

Usage:
```js
exports.config = {
  framework: 'custom',
  frameworkPath: '../lib/frameworks/custom_jasmine.js',
};
```

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -286,6 +286,12 @@ Runner.prototype.run = function() {
     } else if (self.config_.framework === 'explorer') {
       // Private framework. Do not use.
       frameworkPath = './frameworks/explorer.js';
+    } else if (self.config_.framework === 'custom') {
+      if (!self.config_.frameworkPath) {
+        throw new Error('When config.framework is custom, ' +
+          'config.frameworkPath is required.');
+      }
+      frameworkPath = self.config_.frameworkPath;
     } else {
       throw new Error('config.framework (' + self.config_.framework +
           ') is not a valid framework.');
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -244,7 +244,15 @@ exports.config = {
   // ----- The test framework --------------------------------------------------
   // ---------------------------------------------------------------------------
 
-  // Test framework to use. This may be jasmine, jasmine2, cucumber, or mocha.
+  // Test framework to use. This may be one of:
+  //  jasmine, jasmine2, cucumber, mocha or custom.
+  //
+  // When the framework is set to "custom" you'll need to additionally
+  // set frameworkPath with the path relative to the config file or absolute
+  //  framework: 'custom',
+  //  frameworkPath: './frameworks/my_custom_jasmine.js',
+  // See github.com/angular/protractor/blob/master/lib/frameworks/README.md
+  // to comply with the interface details of your custom implementation.
   //
   // Jasmine is fully supported as a test and assertion framework.
   // Mocha and Cucumber have limited beta support. You will need to include your
--- a/lib/configParser.js
+++ b/lib/configParser.js
@@ -146,7 +146,8 @@ ConfigParser.getSpecs = function(config) {
 ConfigParser.prototype.addConfig_ = function(additionalConfig, relativeTo) {
   // All filepaths should be kept relative to the current config location.
   // This will not affect absolute paths.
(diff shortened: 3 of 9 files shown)


sjelin:
I like it.  My only concern is that if we were to do this, and then later decide we wanted to change the interface for frameworks, we would then break all the third-party frameworks.

juliemr:
I like it too - could you update docs/referenceConf.js to include how to use this?

lib/runner.js
```diff
@@ -286,6 +286,12 @@ Runner.prototype.run = function() {
     } else if (self.config_.framework === 'explorer') {
       // Private framework. Do not use.
       frameworkPath = './frameworks/explorer.js';
+    } else if (self.config_.framework === 'custom') {
+      if (!self.config_.frameworkPath) {
+        throw new Error('When config.framework is custom, ' +
+          'config.frameworkPath is required.');
+      }
+      frameworkPath = self.config_.frameworkPath;
```
juliemr: We should resolve `frameworkPath` against the location of the config file, e.g. add it here: https://github.com/angular/protractor/blob/master/lib/configParser.js#L149

lib/frameworks/custom_jasmine.js
```diff
@@ -0,0 +1,9 @@
+/**
```
juliemr: This file should be in the spec folder. I'd put it at `spec/custom/framework.js`

hankduan:
Can you add a note here https://github.com/angular/protractor/blob/master/docs/referenceConf.js#L247 if we allow people to use a custom framework. In the note, reference https://github.com/angular/protractor/blob/master/lib/frameworks/README.md for the interface of this custom framework. 

I'm also a little concerned about what Sammy said "My only concern is that if we were to do this, and then later decide we wanted to change the interface for frameworks, we would then break all the third-party frameworks." -> Maybe even include a disclaimer mentioning that the interface could change without a major version bump; try to either send a PR to change/add to our official framework before using their own

elgalu:
I understand your concern @sjelin, third-party framework maintainers will have to deal with any breaking change if they want to keep upgrading Protractor.

@hankduan reference updates & disclaimer added to frameworks/README.md and also updated referenceConf.js#L247

All done @juliemr , thanks for pointing out to configParser.js#L149, I was stuck with absolute paths.

Side note: tried to test the runner.js validations at scripts/test.js with no luck:
```js
// assert bad configured custom framework throws error
executor.addCommandlineTest('node lib/cli.js spec/errorTest/customFailingFramework.js')
    .expectExitCode(1)
    .expectErrors({
      message: 'When config.framework is custom, ' +
          'config.frameworkPath is required.X'
    });
```

lib/runner.js
elgalu: done!

lib/frameworks/custom_jasmine.js
elgalu: done!

hankduan:
What was the problem?

elgalu:
@hankduan I can't surpass this error:
```bash
running: node lib/cli.js spec/errorTest/customFailingFramework.js
>>> fail: Error: ENOENT, no such file or directory 'test_output_1422021213179.tmp'
```

### customFailingFramework.js
```js
var env = require('../environment.js');

exports.config = {
  seleniumAddress: env.seleniumAddress,

  framework: 'custom',
  // omitted frameworkPath in purpose

  specs: [
    '../custom/smoke_spec.js'
  ],

  capabilities: env.capabilities,

  baseUrl: env.baseUrl,
};
```

hankduan:
Protractor might have some problem writing to your environment since the failure tests writes the results in a json format before reading and parsing it.  

But anyway, I don't think this error test case is necessary. This is a better suit for a unit test:
See https://github.com/angular/protractor/blob/master/spec/unit/runner_test.js

```javascript
  it('should fail when no custom framework is defined', function(done) {
    var config = {
      mockSelenium: true,
      specs: ['*.js'],
      framework: 'custom'
    };
    
    var runner = new Runner(config);
    runner.run().then(function() {
      done.fail('expected error when no custom framework is defined')
    }, done);
  });
```

elgalu:
Thanks @hankduan, test added & PR updated (push --force)

juliemr:
Thanks! Looks good to me, merged as eb9d56755fa93401502e7608c7c3d0f16927c082

PR #1755, opened 2015-01-27
docs(browser-support): split browser support into its own docs section

juliemr:
@sjelin could you take a look?

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -41,7 +41,7 @@ once before any of your tests. See this example ([withLoginConf.js](https://gith
 
 Which browsers are supported?
 -----------------------------
-The last two major versions of Chrome, Firefox, IE, and Safari. See details at [Setting Up the Browser](/docs/browser-setup.md).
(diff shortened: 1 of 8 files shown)


docs/faq.md
```diff
@@ -91,7 +91,7 @@ browser.takeScreenshot().then(function(png) {
 ```
 
 The method to take a screenshot automatically on failure would depend on the type of failure.
-* For failures of entire specs (such as timeout or an expectation within the spec failed), you can add a reporter as such:
+* For failures of entire specs (such as timeout or an expectation within the spec failed), you can add a reporter as such (this is jasmine 1.3 syntax):
```
sjelin: I think this would be better as a comment inside the code block instead of as a parenthetical

sjelin:
LGTM

docs/faq.md
juliemr: Done!

juliemr:
Thanks! Merged as ec985d2a355aa55f398c018aa5392ea296f07486

PR #1756, opened 2015-01-27
How can I get the Protractor configuration object inside my test

Gillespie59:


Gillespie59:
CLA signed.

hankduan:
Please see here for how to access config variables:
https://github.com/angular/protractor/blob/master/docs/referenceConf.js#L197
and https://github.com/angular/protractor/blob/master/lib/runner.js#L190

@juliemr do you think it's better to place to docs for retrieving the config at the top or bottom of referenceConf, instead of in the middle of onPrepare?

juliemr:
I think this will be resolved once the next release is out, as Hank described above.

Gillespie59:
Hi, 

Thanks for your reply. It's the first time I see this getProcessedConfig method,

Manu

PR #1763, opened 2015-01-29
chore(docs): cucumber support disclaimer to 2015

elgalu:
I think this is still true so moving year from 2014 to 2015


hankduan:
Thanks @elgalu, committed via https://github.com/angular/protractor/commit/a9035b8563851d9f182e3e96bdb6478ef1213bb6

Issue #1740, opened 2015-01-20
$$ should be listed under "element.all(locator)" instead of "element(locator)"


marneborn:
$$ is the shortcut for element.all (right?), but in the API documentation it's listed under "element(locator)", it would be more clear if it were under "element.all(locator)".

juliemr:
`$$` is a shortcut for `element.all(by.css(...))`, not generic `element.all`.

This is slightly confusing, but those entries in the docs are actually describing the fact that you can chain element with $.

`element(by.id('foo')).$('.myclass')`
or
`element(by.id('foo')).$$('.myclaass')`

So it makes sense that they're listed under `element`. What we probably should do is highlight the exported globals (`$`, `$$`, `browser`, `element`, etc) in the docs somewhere).

sjelin:
Fixed

PR #1769, opened 2015-01-31
chore(docs): document $ and $$ better

sjelin:
Before, the documentation for the global instances of $ and $$ was where the
documentation for the chaining of $ and $$ should have been, and the
documentation for chaining was no where.  This fixes that.

Closes issue #1740.


juliemr:
LGTM.

Thanks!

PR #1770, opened 2015-01-31
feat(configParser): allow configParser to understand arrays

hankduan:
e.g. allow user to pass `multiCapabilities` via cli

See http://stackoverflow.com/questions/28239378/is-there-any-way-to-pass-multiple-browser-via-protractor-cli for context

### DIFF
--- a/lib/configParser.js
+++ b/lib/configParser.js
@@ -48,6 +48,38 @@ var ConfigParser = function() {
 };
 
 /**
+ * Returns true if 'obj' is either 
+ * - an array, or
+ * - a dictionary with all its keys represented as positive integers.
+ */
+var isArrayLike_ = function(obj) {
+  if (obj instanceof Array) {
+    return true;
+  }
+  if (!(obj instanceof Object)) {
+    return false;
+  }
+
+  for (var key in obj) {
+    var n = parseInt(key);
+    if (String(n) !== key || n < 0) {
+      // is not positive integer
+      return false;
+    }
+  }
+  return true;
+};
+
+/**
+ * Returns true if 'obj' is a dictionary-like object.
+ */
+var isDictionary_ = function (obj) {
+  return (obj instanceof Object &&
+        !(obj instanceof Function) &&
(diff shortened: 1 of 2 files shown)


lib/configParser.js
```diff
@@ -58,9 +90,8 @@ var ConfigParser = function() {
  */
 var merge_ = function(into, from) {
   for (var key in from) {
-    if (into[key] instanceof Object &&
-        !(into[key] instanceof Array) &&
-        !(into[key] instanceof Function)) {
+    if ((isArrayLike_(into[key]) && isArrayLike_(from[key])) || 
+        (isDictionary_(into[key]) && isDictionary_(from[key]))) {
```
juliemr: This means that arrays won't be replaced anymore, right? So it would change behavior where you have a config file with `specs: ['test1.js', 'test2.js']` and then use the command line with `--specs test1.js` so that it would not override and just run test1.js. 

I like that this enables multiConf from the command line, but I'm not sure it's worth losing the scenario above.

juliemr: Actually, I realize that my situation above would work because optimist will parse the command line as a string. But what about:
config:
`specs: ['1.js', '2.js', '3.js']`
command line:
`--specs 1.js,2.js`

hankduan: That's a good point. But by the same token, if config has `param: {a: 2, b:3}`, and command line passes `--param.a 4`, it should override both right?
Maybe the right thing to do is convert array-like structures from command line args to arrays, and always just directly replace the value that's there in the current config?

juliemr: Discussed offline: We think that overriding values in an object make sense (e.g. `--jasmineNodeOpts.showColors false` should not remove all your other jasmineNodeOpts), so we're going to leave this as-is for now. For the multiCapabilities issue, I feel that this is really hard to express via command line in a way that is intuitive and doesn't break other use cases, but I'm happy to take a look at a PR if anyone has a brilliant idea.

PR #1778, opened 2015-02-03
Update locators.js

akhlesh:
example is misleading as there is no DIV which returned by cssContainingText


sjelin:
Good catch.  Merged in at https://github.com/angular/protractor/commit/81715906ebe742cb86c888d2fe2c01b5ecd1be86

PR #1791, opened 2015-02-07
chore(lint): ran the closure linter all over our code

sjelin:
Bunch of tiny fixes

### DIFF
--- a/bin/elementexplorer.js
+++ b/bin/elementexplorer.js
@@ -32,7 +32,7 @@
  * Typing tab at a blank prompt will fill in a suggestion for finding
  * elements.
  */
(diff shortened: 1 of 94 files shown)


hankduan:
LGTM. Can we add this as 'posttest' so it runs automatically?

juliemr:
LGTM. Looks like mostly trailing whitespace - unfortunately, looks like jshint doesn't check for this, so if we can easily add another tool like Hank suggested that would be great.

sjelin:
Merged in.  As I mentioned offline, the closure linter throws a lot of warnings that we don't care about (e.g. complaining that we failed to explain what `module.exports` is), so it would be a bad idea to add it to posttest.

PR #1803, opened 2015-02-12
chore(docs): added "name" property to referenceConf.js

### LINKED ISSUE #1773: referenceConf.js should show 'name' in capabilities
When running tests on SauceLabs they all appear as 'Unnamed job'. This can be fixed by specifying 'name' within a capability.

This isn't shown in the referenceConf or https://code.google.com/p/selenium/wiki/DesiredCapabilities - might be nice to add it?

I would like to be able to pass in the build number as well. Do I need to add a node module to read in the command line params, or is there a built-in way of doing this in the config file? (as browser.params isn't available)

sjelin:
Closes #1773

### DIFF
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -107,6 +107,11 @@ exports.config = {
   capabilities: {
     browserName: 'chrome',
 
+    // Name of the process executing this capability.  Not used directly by
+    // protractor or the browser, but instead pass directly to third patries
+    // like SauceLabs as the name of the job running this test
+    name: 'Unnamed Job',
+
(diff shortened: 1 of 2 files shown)


sjelin:
To be honest I'm totally unsure what this property is beyond what the issue mentioned, so if someone more knowledgeable could take a loot that'd be great :smile:

hankduan:
LGTM. Maybe update https://github.com/angular/protractor/blob/master/docs/server-setup.md#remote-selenium-server too, and mention that 'name' is passed directly to third party (i.e. sauce) and not used by protractor for anything.

PR #1844, opened 2015-02-24
feat(a11yPlugin): plugin for integrating with Chrome Accessibility Developer Tools 

marcysutton:
The first pass on an accessibility plugin for Protractor, integrating with the Chrome Accessibility Developer Tools. 

We talked about adding a plugin hook for running after every test, as a way to increase test coverage. 

I'll see how far I can get on the Tenon integration this week!


marcysutton:
Woot! Plugin hook: https://github.com/angular/protractor/issues/1842

sjelin:
Maybe it would be best if we re-used the copy of angular that the testapp already has instead of your importing your own?

marcysutton:
I thought the accessibility testing app should include ngAria, which was introduced in 1.3 and iterated on in releases since (1.4 would be best when it comes out, but 1.3.5 was a bug-free one). Can we upgrade the 1.3-rc1 version in the testapp folder to something newer? That's why I included a new version. Alternatively, we could use newer ngAria with the 1.3-rc1 release of Angular.

juliemr:
Yes, let's upgrade the all of the testapp version to something newer - it's just on 1.3-rc1 because we never had the motivation to update it. I will do this in a separate PR.

plugins/accessibility/index.js
```diff
@@ -0,0 +1,88 @@
...
+var AUDIT_FILE = path.join(__dirname, '../../node_modules/accessibility-developer-tools/dist/js/axs_testing.js');
+
+/**
+ * You enable this plugin in your config file:
+ *
+ *    // Chrome Accessibility Developer Tools:
+ *    exports.config = {
+ *      ...
+ *      plugins: {
+ *        ...
```
juliemr: Update this to the array format and include path:

```
plugins: [{
  path: 'node_modules/protractor.plugins/accessiblity',
  // etc
}]
```

plugins/accessibility/index.js
```diff
@@ -0,0 +1,88 @@
...
+ *        accessibility: {
+ *          developerTools: true,
+ *        }
+ *      }
+ *    }
+ */
+
+/**
+ * Checks the information returned by the accessibility audit and
+ * displays passed/failed results as console output
```
juliemr: style nitpick: end all sentences with .

plugins/accessibility/index.js
```diff
@@ -0,0 +1,88 @@
...
+ */
+function teardown(config) {
+
+  var data = fs.readFileSync(AUDIT_FILE, 'utf-8');
+  data = data + ' return axs.Audit.run();';
+
+  var testOut = {failedCount: 0, specResults: []},
+      elementPromises = [];
+
+  return browser.executeScript_(data, 'axs rules').then(function(results) {
```
juliemr: let's make the description `a11y developer tool rules` since I won't rememberwhat axs means when debugging

plugins/accessibility/index.js
```diff
@@ -0,0 +1,88 @@
...
+
+    var audit = results.map(function (result) {
```
juliemr: style nit: no space after `function`

plugins/accessibility/index.js
```diff
@@ -0,0 +1,88 @@
...
+          // get elements from WebDriver
+          elementPromises.push(
+            elem.getOuterHtml().then(function(text) {
+              return '\n\t\t' + text;
+            })
+          );
+        });
+      }
+
+      if(result.result === 'FAIL') {
```
juliemr: style nit: add a space in `if (result`

plugins/accessibility/index.js
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * You enable this plugin in your config file:
+ *
+ *    // Chrome Accessibility Developer Tools:
+ *    exports.config = {
+ *      ...
+ *      plugins: {
+ *        ...
+ *        accessibility: {
+ *          developerTools: true,
```
juliemr: Can  you add a TODO comment mentioning that developerTools is the only thing that will be run at the moment, since the config isn't actually used yet?

scripts/test.js
```diff
@@ -112,4 +113,16 @@ executor.addCommandlineTest(
       message: 'warning -- Module "xApp" was created but never loaded.'
     }]);
 
+// Check accessibility plugin
+
+executor.addCommandlineTest(
+    'node lib/cli.js plugins/accessibility/spec/failureConfig.js')
+    .expectExitCode(1)
+    .expectErrors([{
+      message: 'warning -- accessibility plugin cannot be run as there is ' +
```
juliemr: I'm not sure we need this test, since it's not adding anything to the plugin test above.

Could we instead have failureConfig.js load up a file that has accessibility errors, and then assert that they're output?

juliemr:
Awesome, just a couple of nits and comments about tests, as you mentioned.

scripts/test.js
marcysutton: Great idea. I'll do that! The rest of the feedback has been incorporated. Thanks!

juliemr:
I've pushed a commit to use angular v1.3.13 and include aria: https://github.com/angular/protractor/commit/64704e031a6c8fe829192506c49388f3b9b98552

marcysutton:
@juliemr I added a failure test (and I'm glad I did! it caught holes in the implementation), but I'm running into an issue where it's difficult to match the exact error message when the results will vary depending on the markup and audit results. 

At the bottom of `scripts/test.js`, I have a placeholder error message of "warning -- Inaccessible markup". I didn't think we would print this to the console since we already have output, but I'm wondering the best way to assert there was a failure without hard-coding the errorMessage from the failing test (which includes HTML strings, urls, and other text). I looked into `test_util.js` and figured you'd have a good idea of how to handle it. Tenon will present a similar situation.

juliemr:
Git seems to have lost the `accessibility/failureConfig` file.

`expectErrors` can take a regexp - can we use that to verify that the output message are sane? That should take care of not needing to hard code URLs - although I don't mind hard coding some error messages int his situation, it'll easily tell us when something unexpected happens and should be quick to update when we need to change our tests.

marcysutton:
Yep, you were right–Git ate some files when I updated from the previous commit. They are back in there now and I have failing tests working appropriately.

juliemr:
Ok, looks good to me! Merging in as a first version.

juliemr:
Merged as 54163dcd22cee27cf16685fbb4d53a2712233d26

I made a small edit - the testapp webserver takes care of adding the version prefix to the angular directory, so I removed the explicit version in the script srcs.

marcysutton:
Thanks @juliemr!

PR #1850, opened 2015-02-25
bug(debugger): breakpoint isn't set properly for windows

hankduan:


juliemr:
Change the commit message to `fix(debugger)` and then LGTM.

hankduan:
Oops, I keep doing that =(. Done and committed https://github.com/angular/protractor/commit/aabdd567ee62d0d48fad499ee5decbb5d7d6b939

alfonso-presa:
It still doesn't seem to work in windows 7 32bits. I'm in protractor 1.8.0 and I've confirmed that this PR patch is applied in it's code. I'm running the following command:
node_modules\.bin\protractor .\test\protractor.conf.js --elementExplorer

It gets stucked after the message "Debugger listening on port 5858" and seems to never enter command Repl mode.

I've tried to debug protractor's code and it seems like it's never reaching line 667 of protractor.js inside pausePromise object. Look's like a failure in webdriver.promise.controlFlow().execute() because I've also tried running require('child_process').fork(debuggerClientPath, [process.debugPort]); outside of pausePromise (without using flow.execute) and the debugger starts, but it's obviously a lot buggy.

I've you give me any clues of where to keep on debugging I can continue trying to figure out the prob in my windows machine.

MystK:
Same exact issue here, although I am using iojs v1.4.3. Are you on node, Alfonso?

hankduan:
iojs 1.4.3 is forked from node 0.10.35. All versions of node (or forks of node) above 0.10.30 won't work because of this issue here (https://github.com/joyent/node/issues/9333)

I have a patch that makes element explorer compatible with higher versions of node (https://github.com/angular/protractor/pull/1898) but until node (or iojs I guess) fixes the bug, it's not going to help much.

MystK:
Thanks for the reply. I didn't know this was due to a bug in Node.

hankduan:
Well, it's partly Node's bug (which makes element explorer extremely laggy to use) and partly element explorer's incompatibility with higher levels of node. 
The latter I fixed in the PR I linked, but there's no point to submit until node fix the lag issue because whatever they do could require me to make further changes to that PR to make element explorer compatible.

alfonso-presa:
Thank you for the replies. 

Yes. I'm in node v0.12.0 indeed.

I'll try the patch and also wait for nodes fix.

Thanks!

PR #1859, opened 2015-02-26
feat(plugins): add postTest hook for plugins

### LINKED ISSUE #1842: Plugins: Add ability to hook in at the end of each test

juliemr:
Additionally, add some tests to make sure that plugins can fail properly.

Closes #1842

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -15,12 +15,9 @@ var log_ = function() {
 };
 
 /**
- * The plugin API for Protractor.  Note that this API is extremely unstable
- * and current consists of only two functions:
- *    <plugin>.setup - called before tests
- *    <plugin>.teardown - called after tests
- *    <plugin>.postResults - called after test results have been processed
- * More information on plugins coming in the future
+ * The plugin API for Protractor.  Note that this API is unstable. See
+ * plugins/README.md for more information.
+ *
  * @constructor
  * @param {Object} config parsed from the config file
  */
@@ -58,8 +55,12 @@ function pluginFunFactory(funName) {
       var pluginConf = this.pluginConfs[name];
       var pluginObj = this.pluginObjs[name];
       names.push(name);
-      promises.push((pluginObj[funName] || noop)(pluginConf));
+      promises.push(
+          (pluginObj[funName] || noop).apply(
+              pluginObj[funName],
+              [pluginConf].concat([].slice.call(arguments))));
     }
+
     return q.all(promises).then(function(results) {
       // Join the results into a single object and output any test results
       var ret = {failedCount: 0};
@@ -130,5 +131,13 @@ Plugins.prototype.teardown = pluginFunFactory('teardown');
  *     torn down.
  */
 Plugins.prototype.postResults = pluginFunFactory('postResults');
+
+/**
+ * Called after each test block completes.
+ *
+ * @return {q.Promise} A promise which resolves when the plugins have all been
+ *     torn down.
+ */
+Plugins.prototype.postTest = pluginFunFactory('postTest');
 
 module.exports = Plugins;
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -311,9 +311,27 @@ Runner.prototype.run = function() {
       self.on('testFail', restartDriver);
     }
 
+    // We need to save these promises to make sure they're run, but we don't
+    // want to delay starting the next test (because we can't, it's just
+    // an event emitter).
+    var pluginPostTestPromises = [];
+
+    self.on('testPass', function() {
+      pluginPostTestPromises.push(plugins.postTest(true));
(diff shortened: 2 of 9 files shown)


juliemr:
@sjelin can you take a look?

lib/plugins.js
```diff
@@ -131,4 +132,9 @@ Plugins.prototype.teardown = pluginFunFactory('teardown');
  */
 Plugins.prototype.postResults = pluginFunFactory('postResults');
 
+/**
+ * Called after each test block completes.
+ */
+Plugins.prototype.postTest = pluginFunFactory('postTest');
```
sjelin: Thee word "test" seems ambiguous.  Is this after each `it()`, `assert()`, or `describe()`?

juliemr: each `it()` (in Jasmine) - another option would be `spec`. Preference?

sjelin: Also please update the comment on line 17

lib/runner.js
```diff
@@ -311,9 +311,27 @@ Runner.prototype.run = function() {
       self.on('testFail', restartDriver);
     }
 
+    // We need to save these promises to make sure they're run, but we don't
+    // want to delay starting the next test (because we can't, it's just
+    // an event emitter).
+    var pluginPostTestPromises = [];
+
+    self.on('testPass', function() {
+      pluginPostTestPromises.push(plugins.postTest());
```
sjelin: Might the plugins care if the test passed or failed?  It would be pretty easy to modify `pluginFunFactory` so that if `plugins.postTest(true)` was called then that `true` would be passed along into the plugin's callback.

plugins/README.md
```diff
@@ -72,6 +76,16 @@ exports.teardown = function(config) {};
...
+
+/**
+ * Called after each test block completes.
+ *
+ * @param {Object} config The plugin configuration object.
+ *
+ * @return Object If an object is returned, it is merged with the Protractor
+ *     result object. May return a promise.
+ */
+exports.postTest = function(config) {};
```
sjelin: Same comment as before - "test"/"test block" is ambiguous.

juliemr: I think the problem is that these terms are overloaded and there's no general agreement on what they mean. "Test case" might mean one 'it' block or a whole file, or maybe even a whole set of things.

Do you prefer `spec`?

lib/plugins.js
sjelin: We already use `spec` to refer to the whole file.  `testcase` might make sense?

plugins/README.md
sjelin: You're right that these terms are overloaded and there's not much agreement.  We already use the word `spec` in our config files.  Maybe the word `test` is the best after all.

sjelin: Just update the comment with something like "(e.g. `it()` block in jasmine)"

lib/plugins.js
juliemr: Updated comments.

lib/runner.js
juliemr: Done.

plugins/README.md
juliemr: Done.

juliemr:
Comments addressed.

lib/plugins.js
```diff
@@ -58,8 +55,10 @@ function pluginFunFactory(funName) {
       var pluginConf = this.pluginConfs[name];
       var pluginObj = this.pluginObjs[name];
       names.push(name);
-      promises.push((pluginObj[funName] || noop)(pluginConf));
+      promises.push(
+          (pluginObj[funName] || noop)(pluginConf, [].slice.call(arguments)));
```
sjelin: I don't think this is consistent with what you put in `plugins/README.md`.  Shouldn't it be `(pluginObj[funName] || noop).apply(this, [pluginConf].concat([].slice.call(arguments)))`?

sjelin: Also maybe it would be good to test this?

juliemr: Done, and added tests.

spec/plugins/failing_plugin.js
```diff
@@ -0,0 +1,45 @@
...
+      return failingResult('from teardown');
+    });
+  },
+
+  postResults: function() {
+    // This function should cause no failures.
+  },
+
+  postTest: function(config, passed) {
+    if (passed) {
```
sjelin: Both blocks of this conditional are identical

juliemr: Nope, one says 'passing' and one says 'failing'

juliemr: Modified to duplicate fewer lines.

sjelin:
LGTM

Issue #1734, opened 2015-01-16
Heads up: Protractor is not working with Firefox 35/36


ericmdantas:
Yesterday things were working just fine, but today Firefox got updated and protractor suddenly wouldn't run the tests anymore. It'd just show the page - like it never requested the whole thing - and, after a while, it shows the ```Jasmine spec timeout``` type of error.

Apparently it's something related to the driver, since when I downgraded Firefox to 33.1 and ran the tests again, they worked like a charm.

When they release the compatible version (firefox/driver), do I have to update the ```webdriver-manager``` or will ```protractor``` have to be updated as well? 

Thanks!

ValentinH:
Same here!

juliemr:
We're taking a look. Here's what I know so far:

This is not an issue just with Protractor. I can do a test using only webdriverJS against FireFox35 and it fails. Here's the test: 

```js

var util = require('util'),
    webdriver = require('selenium-webdriver');

var driver = new webdriver.Builder().
    usingServer('http://localhost:4444/wd/hub').
    withCapabilities({'browserName': 'firefox'}).
    build();

driver.get('http://juliemr.github.io/webdriver-bugs/');

driver.executeAsyncScript(function(cb) {
  window.setTimeout(function() {
    cb();
  }, 30);
});

driver.quit();
```

In the firefox console, I see the error: `Error: Permission denied to access property '0'`

juliemr:
Filed issue: https://code.google.com/p/selenium/issues/detail?id=8398

ericmdantas:
At some point I got that error too, @juliemr.

Looks like we'll just have to wait, then.

Thanks for looking into it =]

ehacke:
:+1: Same issue here.

flegall:
Same issue here.

byaminov:
:+1: same here

juliemr:
My issue was merged into https://code.google.com/p/selenium/issues/detail?id=8390 which is tracking. Note that this is broken on all systems for FireFox 35.

Olgagr:
:+1:

igorshubovych:
+1
http://stackoverflow.com/questions/27645260/protractor-e2e-tests-run-at-chrome-but-halt-at-firefox

pfrandsen:
Temporary "solution" - downgrade to Firefox 34: https://ftp.mozilla.org/pub/mozilla.org/firefox/releases/34.0

juliemr:
[...]

PR #1860, opened 2015-02-26
tests(node version): test on Node 0.12 & io.js instead of unstable Node 0.11

mzgol:
I've also updated Chrome & Firefox in the first commit (I can split them into two PRs if you want). Chrome 38 doesn't work with latest chromedriver which caused smoke tests
to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: node_js
 node_js:
   - "0.10"
-  - "0.11"
+  - "0.12"
+  - "iojs"
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


juliemr:
Sorry, I made and merged 32a4a0d3d0696c73704a36bcf06f81fd6fa023a4 before I saw this.

Can you make a separate PR to update the version of chrome only? Firefox should be at 33 because the latest two versions we're testing against are 34 and 33, due to #1734.

As for io.js, let's test against node 0.12 for now and avoid upping our test time/VMs by adding more engines for the moment.

mzgol:
> Can you make a separate PR to update the version of chrome only? Firefox should be at 33 because the latest two versions we're testing against are 34 and 33, due to #1734.

OK, so I guess it should be Chrome 39 as well since the latest is 40. PR: #1864

PR #1865, opened 2015-02-26
chore(travis): leverage Travis container infrastructure

mzgol:
Adding `sudo: false` to the Travis config makes it run on a container
infrastructure which starts up faster than default. See:
http://docs.travis-ci.com/user/workers/container-based-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: node_js
+
+sudo: false
+
 node_js:
   - "0.10"
   - "0.12"


mzgol:
This may influence what the build can do (e.g. `sudo` is disallowed) so I'd wait to see if nothing gets broken by this PR.

juliemr:
This doesn't actually appear to run any faster (with a sample size of one test run). Is there any other reason for doing it?

mzgol:
Are Protractor tests running on dedicated hardware? If that's the case then maybe it doesn't improve anything. I know in my modules I've seen a massive speed up in startup time (execution time is probably not affected in a major way).

juliemr:
All tests under the 'angular' repo are run on dedicated hardware, (we already get the message `This job is running on container-based infrastructure` on Travis). So, I'm pretty sure this is redundant.

mzgol:
> So, I'm pretty sure this is redundant.

It seems you're right. Sorry! Closing the PR.

Issue #1854, opened 2015-02-25
A11y plugin improvements


juliemr:
- [ ] pending https://github.com/angular/protractor/issues/1842 - audit after each test and merge results
- [ ] Look into adding custom audits specific to Angular applications
- [x] Enable Tenon.io auditing for users of that service
- [x] Run the tests against Protractor's own website
- [x] Add an option to configure whether warnings from Chrome a11y are treated as pass or fail
- [ ] Lower priority: Add meaningful duration numbers to test output
- [x] More sophisticated truncating of long element strings

@marcysutton did I miss anything?

marcysutton:
- [ ] Missing accessibility/successConfig in test.js

Question: will a11y plugin failures fail CI builds? Seems like part of #1842.

juliemr:
Yup, plugin failures will fail CI builds - they change the exit code to non-zero, which is all that Travis cares about.

For merging results, I think it's up to what we feel is most useful. We should probably use the test name to help developers pinpoint where failures actually occurred, but do some sort of deduplication if the same issue is present in a bunch of tests.

marcysutton:
I think I got the "identifying header" part solved by adding text to the Pass/Fail descriptions. It's especially useful when you run both integrations at the same time. PR coming shortly.

marcysutton:
Hey @juliemr I have another improvement to add to the accessibility plugin: make the output color configurable. The red and green colors are very hard to see for some folks.

sjelin:
There's a limited amount we can do with that.  Most of the output comes from tools like jasmine.

dylanb:
@juliemr I would like to integrate the axe-core https://github.com/dequelabs/axe-core ruleset into this functionality - WDYT?

juliemr:
Feature requests should  now go to https://github.com/angular/protractor-accessibility-plugin

PR #1867, opened 2015-02-27
feat(a11yPlugin): add support for Tenon.io

marcysutton:
As part of #1854, this change adds support for the [Tenon](http://www.tenon.io) accessibility testing framework. I removed my API key from the plugin config files, so the Protractor internal tests won't run against Tenon as-is. The Chrome Accessibility Dev Tools will still run, though. 

One thing I noticed was that the `accessibility-developer-tools` package was missing from package.json. I added it back in along with the `entities` package for decoding HTML strings, which has no dependencies.

Console output for accessibility plugin


marcysutton:
One thing we will need to document: in Tenon, scripts have to be accessible from the page source as a string, for example loaded through a CDN. Otherwise Tenon won't be able to test expressions coming from Angular. I ran into this when supplying an alt attribute as an expression. To get around it for now, I just hard-coded a value.

plugins/accessibility/index.js
```diff
@@ -17,12 +17,37 @@ var q = require('q'),
...
+ *            key: 'YOUR_API_KEY_HERE', // only required parameter
+ *            level: 'AA' // WCAG AA OR AAA,
+ *            filter: '0', // one of 0, 20, 40, 60, 80, 100, defaults to 0
+ *            certainty: '0', // one of 0, 20, 40, 60, 80, 100, defaults to 0
+ *            store: '0', // 0 or 1, whether Tenon should store your results, depends on your membership plan
+ *          },
+ *          printAll: false, // whether the plugin should log API response
+ *        },
+ *        chromeA11YDevTools: false,
+ *        path: 'node_modules/protractor.plugins/accessiblity'
```
juliemr: typo: .plugins instead of /plugins

plugins/accessibility/index.js
```diff
@@ -1,14 +1,14 @@
...
+    request = require('request'),
+    Entities = require('html-entities').XmlEntities;
 
 /**
- * You can enable this plugin in your config file:
- *
- *    // The Chrome Accessibility Developer Tools are currently
- *    // the only integration option.
+ * You can audit your website against the Chrome Accessibility Developer Tools,
+ * Tenon.io, or both by enabling this plugin in your config file:
```
juliemr: include a link for Tenon here?

plugins/accessibility/index.js
```diff
@@ -17,12 +17,37 @@ var q = require('q'),
...
+ *
+ *    //  Read about the Tenon.io settings and API requirements:
+ *    //  -http://tenon.io/documentation/overview.php
+ *    //  -http://tenon.io/documentation/understanding-request-parameters.php
+ *
+ *    exports.config = {
+ *      ...
+ *      plugins: [{
+ *        tenonIO: {
+ *          options: {
```
juliemr: Hmm - can we actually just pass the entire options object to tenon, and just modify the 'src'? So here say something like:

tenonIO: {
  options: {} // See http://tenon.io/documentation/understanding-request-parameters.php
                   // The src will be added by the test.

plugins/accessibility/index.js
```diff
@@ -32,76 +57,187 @@ var AUDIT_FILE = path.join(__dirname, '../../node_modules/accessibility-develope
...
+ */
+function runTenonIO(config) {
+  var options = config.tenonIO.options;
+
+  return browser.driver.getPageSource().then(function(source) {
+    // setup response as a deferred promise
+    var deferred = q.defer();
+    request.post({
+      url: TENON_URL,
+      form: {
```
juliemr: As above, can we make all options available by just setting: options.src = source and the passing the whole thing here?

plugins/accessibility/index.js
```diff
@@ -32,76 +57,187 @@ var AUDIT_FILE = path.join(__dirname, '../../node_modules/accessibility-develope
...
+        key: options.key,
+        src: source,
+        level: options.level,
+        filter: options.filter,
+        certainty: options.certainty,
+        store: options.store
+      }
+    },
+    function(err, httpResponse, body) {
+      if (err) return resolve.reject(new Error(err));
```
juliemr: nitpick: even for one line if statements google style guide uses {}

plugins/accessibility/index.js
```diff
@@ -32,76 +57,187 @@ var AUDIT_FILE = path.join(__dirname, '../../node_modules/accessibility-develope
...
+    var testHeader = 'Tenon.io - ';
+
+    if(numResults === 0) {
+      return testOut.specResults.push({
+        description: testHeader + 'All tests passed!',
+        assertions: [{
+          passed: true,
+          errorMsg: ''
+        }],
+        duration: 1
```
juliemr: Not necessary, but in the future it would be nice to actually time the calls to Tenon and insert a real duration here.

testapp/accessibility/badMarkup.html
```diff
@@ -17,5 +17,6 @@
...
...1.2.3</option><option value="object:81" label="v1.2.2">v1.2.2</option><option value="object:82" label="v1.2.1">v1.2.1</option><option value="object:83" label="v1.2.0">v1.2.0</option><option value="object:84" label="v1.2.0-rc.3">v1.2.0-rc.3</option><option value="object:85" label="v1.2.0-rc.2">v1.2.0-rc.2</option></optgroup><optgroup label="v1.1.x"><option value="object:87" label="v1.1.4">v1.1.4</option><option value="object:88" label="v1.1.3">v1.1.3</option><option value="object:89" label="v1.1.2">v1.1.2</option><option value="object:90" label="v1.1.1">v1.1.1</option><option value="object:91" label="v1.1.0">v1.1.0</option></optgroup><optgroup label="v1.0.x"><option value="object:93" label="v1.0.7">v1.0.7</option><option value="object:94" label="v1.0.6">v1.0.6</option><option value="object:95" label="v1.0.5">v1.0.5</option><option value="object:96" label="v1.0.4">v1.0.4</option><option value="object:97" label="v1.0.3">v1.0.3</option><option value="object:98" label="v1.0.2">v1.0.2</option><option value="object:99" label="v1.0.1">v1.0.1</option><option value="object:100" label="v1.0.0">v1.0.0</option><option value="object:101" label="v1.0.0-rc2">v1.0.0-rc2</option></optgroup></select>
```
juliemr: This is testing that a long select gets shortened? I don't see that happening anywhere, but maybe I'm missing it?

juliemr:
Looks good overall, just a couple comments.

plugins/accessibility/index.js
marcysutton: I put it just below in the Tenon section, and the name itself is a URL. Since the Chrome version wasn't linked, I figured we were already covered.

plugins/accessibility/index.js
marcysutton: I like it. I did keep `printAll` outside of that options object because it's related to Protractor's output, not Tenon. I wanted to provide a way for developers to see the entire response for debugging purposes.

plugins/accessibility/index.js
marcysutton: Great suggestion. Thanks.

plugins/accessibility/index.js
marcysutton: In the Tenon response we get some info, can we make use of either of these properties?

```json
 responseExecTime: '0.69',
 responseTime: '2015-02-27T19:27:59.060Z',
```

testapp/accessibility/badMarkup.html
marcysutton: Yep, it was the Chrome Dev Tools that needed shortening. Now it truncates element HTML strings to 200 characters when we return them from WebDriver.

plugins/accessibility/index.js
juliemr: responseExecTime sounds OK - any idea what those units are? 

Alternatively, we could just record a start time before we send the request.

marcysutton: In the Tenon docs it says it's "how long, in seconds, it took for us to test your document". Maybe we could start a timer and add that to it to get an estimation of the total running time?

marcysutton:
@juliemr FYI I pushed all changes except for the duration thing, which I figured could come later. Let me know if there's anything else you'd like to see updated.

juliemr:
Looks good. Merged as 13d34c9192a06634827d89bf356bea33fea75747

marcysutton:
Woooot! :tada: @karlgroves Tenon is merged into Protractor master!

ginader:
:+1:

karlgroves:
w00t!

PR #1868, opened 2015-02-27
feat(plugins): allow plugins to export a name for use in reporting

juliemr:


juliemr:
@sjelin one more for you!

lib/plugins.js
```diff
@@ -22,19 +22,24 @@ var log_ = function() {
...
     } else {
-      path = pluginConf.package || name;
+      path = pluginConf.package;
+    }
+    if (!path) {
+      throw new Error('Plugin configuration did not contain a valid path.');
     }
-    this.pluginObjs[name] = require(path);
-  }
+    pluginConf.name = path;
```
sjelin: Should this be the full path or just the file name?

spec/plugins/failing_plugin.js
```diff
@@ -35,5 +35,5 @@ module.exports = {
     return q.delay(100).then(function() {
       return failingResult('from postTest ' + (passed ? 'passing' : 'failing'));
     });
-  }
+  },
```
sjelin: nit: pretty sure we don't put commas there

lib/plugins.js
juliemr: The file name is usually just 'index.js', which isn't very helpful.

We could use some heuristic to try to trim the last folder + file name so that 'foo/bar/baz/index.js' turns into 'baz/index.js'. Does that seem better?

sjelin: Naw, you've convinced me.  The full path makes sense

sjelin:
LGTM

spec/plugins/failing_plugin.js
juliemr: Done.

juliemr:
Thanks! Merged.

PR #1877, opened 2015-03-02
fix(a11yPlugin): Tenon API daily limit handling

marcysutton:
I hit the Tenon daily API limit in a short amount of testing and found the need for error messaging. This PR provides a "Daily limit exceeded" message or a catch-all error message before exiting.


juliemr:
Looks good, merged as e5570fb976a7277be690dc69b448cb94540ecfe5

PR #1889, opened 2015-03-03
fix(debugger): fix issue where output does not display circular dep and ...

hankduan:
...functions

### DIFF
--- a/lib/debugger/modes/commandRepl.js
+++ b/lib/debugger/modes/commandRepl.js
@@ -30,7 +30,13 @@ var CommandRepl = function(client) {
 CommandRepl.prototype.stepEval = function(expression, callback) {
   expression = expression.replace(/"/g, '\\\"');
   var expr = 'browser.dbgCodeExecutor_.execute("' + expression + '")';
-  this.evaluate_(expr, callback);
+  this.evaluate_(expr, function(err, res) {
+    // Result is a string representation of the evaluation.
+    if (res !== undefined) {
+      console.log(res);
+    }
+    callback(err, undefined);
+  });
 };
 
 /**
@@ -47,7 +53,11 @@ CommandRepl.prototype.complete = function(line, callback) {
   } else {
     line = line.replace(/"/g, '\\\"');
     var expr = 'browser.dbgCodeExecutor_.complete("' + line + '")';
-    this.evaluate_(expr, callback);
+    this.evaluate_(expr, function(err, res) {
+      // Result is a JSON representation of the autocomplete response. 
+      var result = res === undefined ? undefined : JSON.parse(res);
+      callback(err, result);
+    });
   }
 };
 
@@ -77,16 +87,14 @@ CommandRepl.prototype.evaluate_ = function(expression, callback) {
           command: 'evaluate',
           arguments: {
             frame: 0,
-            maxStringLength: 2000,
+            maxStringLength: -1,
             expression: 'browser.dbgCodeExecutor_.getResult()'
           }
         }, function(err, res) {
           try {
-            var result = res.value === undefined ?
-                undefined : JSON.parse(res.value);
-            callback(err, result);
+            callback(err, res.value);
           } catch (e) {
-            callback(e, null);
+            callback(e, undefined);
           }
           self.client.removeListener('break', onbreak_);
         });
--- a/lib/protractor.js
+++ b/lib/protractor.js
@@ -1,3 +1,4 @@
+var util = require('util');
 var url = require('url');
 var webdriver = require('selenium-webdriver');
 var helper = require('./util');
@@ -698,14 +699,7 @@ Protractor.prototype.initDebugger_ = function(debuggerClientPath, opt_debugPort)
       self.execPromiseResult_ = self.execPromiseError_ = undefined;
 
       self.execPromise_ = self.execPromise_.
-          then(function() {
-            var result = execFn_();
-            if (webdriver.promise.isPromise(result)) {
-              return result.then(function(val) {return val;});
-            } else {
-              return result;
-            }
-          }).then(function(result) {
+          then(execFn_).then(function(result) {
             self.execPromiseResult_ = result;
           }, function(err) {
             self.execPromiseError_ = err;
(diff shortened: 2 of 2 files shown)


lib/protractor.js
```diff
@@ -698,14 +699,7 @@ Protractor.prototype.initDebugger_ = function(debuggerClientPath, opt_debugPort)
       self.execPromiseResult_ = self.execPromiseError_ = undefined;
 
       self.execPromise_ = self.execPromise_.
-          then(function() {
-            var result = execFn_();
-            if (webdriver.promise.isPromise(result)) {
-              return result.then(function(val) {return val;});
```
sjelin: What was the reason for this in the past?  It doesn't appear to do anything

lib/protractor.js
```diff
@@ -720,22 +714,49 @@ Protractor.prototype.initDebugger_ = function(debuggerClientPath, opt_debugPort)
     },
 
     // Execute a piece of code.
+    // Result is a string representation of the evaluation.
     execute: function(code) {
       var execFn_ = function() {
+        var deferred = webdriver.promise.defer();
```
sjelin: Instead of having this deferred, you can simply this by using:

```js
  var res = vm_.runInThisContext(code);
  if(!webdriver.promise.isPromise(res)) {
    res = webdriver.promise.fulfilled(res);
  }
  return res.then(...
```

lib/protractor.js
hankduan: `element(by.xyz).getText()` is an elementFinder/promise.
When someone enters `element(by.xyz).getText()`, we need to return the text value of the elementFinder (i.e. by calling then) instead of the text representation of the elementFinder.

lib/protractor.js
hankduan: Thanks, amended.

lib/protractor.js
```diff
@@ -720,22 +714,42 @@ Protractor.prototype.initDebugger_ = function(debuggerClientPath, opt_debugPort)
...
 
     // Execute a piece of code.
+    // Result is a string representation of the evaluation.
     execute: function(code) {
       var execFn_ = function() {
         // Run code through vm so that we can maintain a local scope which is
         // isolated from the rest of the execution.
-        return vm_.runInThisContext(code);
+        var res = vm_.runInThisContext(code);
+        if(!webdriver.promise.isPromise(res)) {
```
juliemr: nit: add a space after `it`

lib/debugger/modes/commandRepl.js
```diff
@@ -30,7 +30,13 @@ var CommandRepl = function(client) {
 CommandRepl.prototype.stepEval = function(expression, callback) {
   expression = expression.replace(/"/g, '\\\"');
   var expr = 'browser.dbgCodeExecutor_.execute("' + expression + '")';
-  this.evaluate_(expr, callback);
+  this.evaluate_(expr, function(err, res) {
+    // Result is a string representation of the evaluation.
+    if (res !== undefined) {
+      console.log(res);
```
juliemr: why this change? Do we need an extra log here?

hankduan: Before I was using JSON.stringify to serialize the data between protractor and the commandRepl, which would parse the string. However, this was not good because you can't serialize functions and objects with circular dependencies. 

The only way to stringify the full object mimic what `console.log` does, which is to use `util.format` (see https://github.com/joyent/node/blob/v0.12.0/lib/console.js#L55). 

So then, lastly, why I can't do `callback(err, res)`, but must instead do `console.log(res); callback(err, undefined)`. It's because during callback, nodejs doesn't console.log the result, but will rather print the result as is. 

Here are some outputs if I do `callback(err, res)`:

```
> browser
'{ controlFlow: [Function],\n  schedule: [Function],\n  getSession: [Function],\n  getCapabilities: [Function],\n  quit: [Function],\n  actions: [Function],\n  executeScript: [Function],\n  executeAsyncScript: [Function],\n  call: [Function],\n  wait: [Function],\n  sleep: [Function],\n  getWindowHandle: [Function],\n  getAllWindowHandles: [Function],\n  getPageSource: [Function],\n  close: [Function],\n  getCurrentUrl: [Function],\n  getTitle: [Function],\n  findElementInternal_: [Function],\n  findDomElement_: [Function],\n  findElementsInternal_: [Function],\n  takeScreenshot: [Function],\n  manage: [Function],\n  switchTo: [Function],\n  driver: \n   { session_: \n      { then: [Function: then],\n        cancel: [Function: cancel],\n        isPending: [Function: isPending] },\n     executor_: { execute: [Function] },\n     flow_: \n      { events_: {},\n        timer: [Object],\n        history_: [],\n        activeFrame_: [Object],\n        schedulingFrame_: [Object],\n        shutdownId_: null,\n        eventLoopId_: [Object],\n        pendingRejections_: 0,\n        numAbortedFrames_: 0 } },\n  element: { [Function] all: [Function] },\n  \'$\': [Function],\n  \'$$\': [Function],\n  baseUrl: \'http://www.angularjs.org\',\n  rootEl: \'body\',\n  ignoreSynchronization: false,\n  getPageTimeout: 10000,\n  params: {},\n  resetUrl: \'data:text/html,<html></html>\',\n  mockModules_: [ { name: \'protractorBaseModule_\', script: [Function], args: [] } ],\n  getProcessedConfig: [Function],\n  forkNewDriverInstance: [Function],\n  dbgCodeExecutor_: \n   { execPromise_: \n      { then: [Function: then],\n        cancel: [Function: cancel],\n        isPending: [Function: isPending] },\n     execPromiseResult_: undefined,\n     execPromiseError_: undefined,\n     replServer_: \n      { domain: null,\n        _events: {},\n        _maxListeners: 10,\n        useGlobal: false,\n        ignoreUndefined: false,\n        eval: [Function],\n        inputStream: [Object],\n        outputStream: [Object],\n        lines: [Object],\n        context: [Object],\n        bufferedCommand: \'\',\n        prompt: \'> \',\n        rli: [Object],\n        commands: [Object],\n        writer: [Object],\n        useColors: false },\n     execute_: [Function],\n     execute: [Function],\n     complete: [Function],\n     resultReady: [Function],\n     getResult: [Function] } }'
> element
'function (locator) {\n    return new ElementArrayFinder(ptor).all(locator).toElementFinder_();\n  }'
```

lib/protractor.js
hankduan: done.

juliemr:
LGTM sorry for the delay

hankduan:
merged in https://github.com/angular/protractor/commit/5d232807f1e32a9a3ba5a5e4f07ace5d535fc3cd

PR #1891, opened 2015-03-04
docs(): add plugin documentation page

marcysutton:
@juliemr I started a documentation page for Plugins. The intro was sort-of a content dump, so it would benefit a lot from your input. I looked at past releases in the Changelog and source code to put everything together in one place.


sjelin:
I think it would make sense bring in the remaining content at [plugins/README.md](../blob/master/plugins/README.md) and then just have that README link to the doc file

marcysutton:
@sjelin that's a great idea. The content in that file does make sense to go on the Plugin documentation page. The routing for pages points to html files generated from the docs folder, how do you generate a page from a README in a different directory?

sjelin:
I was thinking that you'd copy the content from `plugins/README.md` into `docs/plugins.md` and then just replace `plugins.README.md` with something like:

```md
Please see [the plugin documentation](../docs/plugins.md)
```

It is possible to mess with the website generating script to pull in content from other places but that would be confusing

marcysutton:
Ok, I've made the update. The content is now in `docs/plugins.md` with a link directing people there from `plugins/readme.md`.

sjelin:
You need to update [this test](https://github.com/angular/protractor/blob/master/website/test/e2e/navigation_spec.js).  You can run the website tests by doing the following:

1. Open three terminal tabs
2.  In the first, do `webdriver-manager start`
3. In the second, do `npm start` from inside the `/website/` folder
4. In the third, do `npm test` from inside the `/website/` folder

docs/plugins.md
```diff
@@ -0,0 +1,262 @@
...
+```javascript
+// protractor.conf.js
+exports.config = {
+
+  // ... the rest of your config
+
+  plugins: [{
+    // The only required field for each plugin is the path to that
+    // plugin's entry script.
+    path: 'path/to/plugin/index.js',
```
sjelin: Mention that paths are relative to the location of the config file

docs/plugins.md
```diff
@@ -0,0 +1,262 @@
...
+Plugins extend Protractor's base features by using hooks during test
+execution to gather more data and potentially modify the test output.
+
+The Protractor API and available plugins are *BETA* and may change
+without a major version bump.
+
+The `plugins` folder contains default plugins for Protractor.
+
+##In this document:
+* Using Plugins
```
sjelin: These should be links  (e.g. `[Using Plugins](#using-plugins)`)

docs/plugins.md
```diff
@@ -0,0 +1,262 @@
...
+};
+```
+`asTests` specifies if the plugin should generate passed/failed test results
+based on the ngHint output or instead log the results to the console.
+Defaults to true.
+
+`excludeURLs` specifies a list of URLs for which ngHint results should be
+ignored. Defaults to []
+
+The strategy for this plugin is as follows:
```
sjelin: This like though line 238 are about the internal implementation.  Not sure it's relevant to this doc

sjelin:
Should be good once you address all the comments

docs/plugins.md
marcysutton: That's what I thought too, but I wasn't sure how to quickly handle smooth scrolling as links in that format are affected by client-side routing. I started looking into it, but quickly grew beyond the scope of writing documentation. I'd love some help on it!

marcysutton:
@sjelin good to know about the page specs, I'll add one. I was tempted to also make the navigation keyboard accessible, which I could do pretty easily but I was unsure where the logic lived for the main-nav dropdowns. Is it in `test/e2e/menu-partial.js`?

docs/plugins.md
sjelin: I'm confused.  I think you just write the link as `[Using Plugins](#using-plugins)` and github handles the rest?

sjelin:
Everything under `/website/test/` is part of the test code.  Everything under [`/website/js`](https://github.com/angular/protractor/tree/master/website/js) ends up included in the final website's JavaScript.

docs/plugins.md
marcysutton: Links can be created from Markdown, yes. But the Protractor website app has client-side routing which interferes with internal links. A link in that format will currently kick you back to the homepage.

marcysutton:
Yeah I looked in `/website/js` but didn't find anything related to the navigation.

sjelin:
What are you looking for exactly?  All the routing code is in `/website/js`, the templates which are not generated from the docs are in [`/website/partials`](https://github.com/angular/protractor/tree/master/website/partials), and the code which generates templates from markdown files is in the [gulpfile](https://github.com/angular/protractor/blob/master/website/gulpfile.js#L96).

marcysutton:
I was hoping to make the navigation accessible–right now you can't operate the dropdown menus with the keyboard and the links all have `tabindex="-1"`, which should only be set when the dropdowns are collapsed. I was just verifying the location of the code for those dropdowns, as it appears to be in the test folder.

docs/plugins.md
sjelin: Don't worry about that.  I have a [pending PR](https://github.com/angular/protractor/pull/1894) to deal with that.

website/test/e2e/navigation_spec.js
```diff
@@ -161,6 +161,12 @@ describe('Navigation', function() {
       expect($('h1').getText()).toBe('Browser Support');
     });
 
+    it('should go to Plugins', function() {
+      menu.dropdown('Reference').item('Plugins');
+
+      expect($('h1').getText()).toBe('Plugins');
+    });
```
sjelin: Did you run this test?  I think you need to modify the test at line 65

sjelin:
The test folder does not contain any logic which makes it into the actual website.  It's just the logic for testing the website.  The [`/website/test/e2e/menu-partial.js`](https://github.com/angular/protractor/blob/master/website/test/e2e/menu-partial.js) file just provides some wrapper functions for manipulating the dropdowns within [`/website/test/e2e/navigation_spec.js`](https://github.com/angular/protractor/blob/master/website/test/e2e/navigation_spec.js).

sjelin:
If you want to remove `tabindex="-1"` you can do so.  I'm not totally sure why that's there anyway.

sjelin:
@marcysutton Did I scare you away? :worried: If you want I can deal with the parts I commented on myself and merge in this PR, and you can work on the Protractor website's usability in a separate PR.

marcysutton:
Hi @sjelin! I looked a little closer and found that because the nav links already have `display: none` when they are hidden, removing `tabIndex="-1"` helped out. There is more we could do for assistive technologies, but I was unable to locate the logic for those dropdowns. This is probably good for now.

juliemr:
Let's get the plugin docs changes in here and do the rest as a separate PR. I think everything is ready to go besides the website tests, is that correct?

marcysutton:
There's a plugin route-change test, is that sufficient? Should I actually move code out of this PR to create another one or leave it as-is?

juliemr:
No need to do anything else! I've committed the docs part of this as 7c752e51221338ea9f60323f008ad8b981c8d8c0

Let's open a new PR for anything else that needs to be done with the navigation.

PR #1892, opened 2015-03-04
fix(doc): Fix description of hooking Jasmine reporter

tullmann:
The Jasmine reporter-hooking example didn't work.  Some paste-o's when
copying it from jasminewd/index.js, I think.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -81,7 +81,7 @@ See an example ([spec.js](https://github.com/juliemr/protractor-demo/blob/master
 
 How can I get screenshots of failures?
 -------------------------------------------- 
-First, this is how you can take a screenshot:
+First, this is how you can save a screenshot to a file:
 ```javascript
 browser.takeScreenshot().then(function(png) {
   var stream = fs.createWriteStream("/tmp/screenshot.png");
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -95,12 +95,12 @@ The method to take a screenshot automatically on failure would depend on the typ
 
 ```javascript
 // Note: this is using Jasmine 1.3 reporter syntax.
```
sjelin: So is this jasmine 2.0 syntax now?  If so you should update this comment

tullmann: This wasn't intended to be a Jasmine2.0 fix, just to make the code work.  But now that I look more closely, it should work fine as-is.  I thought this code was not working in my local setup, but I didn't really grok the anonymous class that `new function() { ... }` implements.  It actually works fine now that I know how it works.

tullmann:
Never mind.  Sorry for the confusion.

tullmann:
I was misled by this jslint warning on the code:  `Weird construction. Is 'new' unnecessary? (W057)`.

Should anyone else run into that in the future, the [supernew](http://jshint.com/docs/options/#supernew) jshint option will suppress this warning.  The `new` is definitely necessary in this case.

Issue #1890, opened 2015-03-03
Element explorer hangs in protractor >=1.7 if you have node >=0.10.31


hankduan:
Element explorer hangs in versions 1.7 or later. 

NOTE: this behavior only appears if you are on node version 0.10.31 or above

Under closer look, you should see that this delay is not indeterministic. To be exactly:
if you wait n seconds to enter a command, the delay will be (n + time to execute command) seconds. 
In other words, if you execute a simple command like `var x = 3`, which takes no time to execute, the delay is equal to the amount of time you waited before entering the command. 

Try this:
```
$ //start element explorer
> var x = 3  //should return immediately
> // now wait 10 seconds, and then type `var x = 3`
> // should wait ~10s before returning. 
```

An issue has been filed with additional information here: https://github.com/joyent/node/issues/9333

hankduan:
FYI: The only workaround for this issue right now is either:
1) downgrade protractor to 1.6.1
or
2) use node.js to 0.10.30 or lower.

shattenjager:
I tried to downgrade to 0.10.29 but still is not working. I get SyntaxError: evalmachine.<anonymous>:1 on my Win machine. I also tried to install an older version of Protractor but I don't know how, since I use npm. My current version is 2.1.0. Also I get this error not using elementExplorer but the debug mode triggered by a browser.pause() method in my code, then I used "repl" to switch to the interactive mode; from there I can do only a browser.get() but can't manage to interact with any element.
![image](https://cloud.githubusercontent.com/assets/7750660/7850639/e5f84dc2-0499-11e5-8d15-2d03f75bd79e.png)

Win  8.1 64
node 0.12.4
npm 2.10.1
protractor 2.1.0

juliemr:
@shattenjager you do have a syntax error, you're missing a `)` closing the `element` call.

I'm not sure why your error message is so bad, that seems like something to do with how node interacts with windows. On mac, it prompts you to continue the input, since the parentheses haven't been matched.

shattenjager:
Geez, thanks, I feel dumb, I apologize!!

hankduan:
fyi: This will be fixed when node 0.12.5 gets released. https://github.com/joyent/node/issues/9333#event-333488911.

juliemr:
Node 0.12.5 is out, woo! Closing.

PR #1898, opened 2015-03-04
feat(debugger): make element explorer work with node 0.12.0

hankduan:
Node has changed its debugger significantly in 0.12.0, and these changes are necessary to get it to work now.

Specifically here are the key points to take note:

* Before, the process continues running after `process._debugProcess(pid);` is called, but now, the process stops. 
To over come this, the call to `process._debugProcess(pid)` is moved from protractor into the debugger client. Secondly, because the process stops after the call, we call `reqContinue` once the debugger connection is established, so that protractor continues into the first break point (for backwards compatibility, we added an extra empty webdriver command so that in earlier versions of node, protractor doesn't go past the first break point from the reqContinue). 

* Before repl provides '(foobar\n)' when an user inputs 'foobar'. Now it is just 'foobar\n'.
We will parse and strip away both the parenthesis and '\n' to support all versions of node. 

Additionally, this change makes debugger processes fail fast if the port is taken.

### DIFF
--- a/lib/debugger/clients/explorer.js
+++ b/lib/debugger/clients/explorer.js
@@ -1,5 +1,5 @@
 var repl = require('repl');
-var baseDebugger = require('_debugger');
+var debuggerCommons = require('../debuggerCommons');
 var CommandRepl = require('../modes/commandRepl');
 
 /**
@@ -9,7 +9,7 @@ var CommandRepl = require('../modes/commandRepl');
  * @constructor
  */
 var WdRepl = function() {
-  this.client = new baseDebugger.Client();
+  this.client;
 };
 
 /**
@@ -17,20 +17,8 @@ var WdRepl = function() {
  * @private
  */
 WdRepl.prototype.initClient_ = function() {
-  var client = this.client;
-
-  client.once('ready', function() {
-
-    client.setBreakpoint({
-      type: 'scriptRegExp',
-      target: '.*executors\.js', //jshint ignore:line
-      line: 37
-    }, function() {});
-  });
-
-  var host = 'localhost';
-  var port = process.argv[2] || 5858;
-  client.connect(port, host); // TODO - might want to add retries here.
+  this.client = 
+      debuggerCommons.attachDebugger(process.argv[2], process.argv[3]);
 };
 
 /**
@@ -103,7 +91,7 @@ WdRepl.prototype.initRepl_ = function() {
   var stepEval = function(cmd, context, filename, callback) {
     // The command that eval feeds is of the form '(CMD\n)', so we trim the
     // double quotes and new line. 
-    cmd = cmd.slice(1, cmd.length - 2);
+    cmd = debuggerCommons.trimReplCmd(cmd);
     cmdRepl.stepEval(cmd, function(err, res) {
       // Result is a string representation of the evaluation.
       if (res !== undefined) {
@@ -142,7 +130,7 @@ WdRepl.prototype.initReplOrServer_ = function() {
   // advantage is that the process becomes much more realistic because now we're
   // using the normal repl. However, it was not possible to test autocomplete 
   // this way since we cannot immitate the TAB key over the wire. 
-  var debuggerServerPort = process.argv[3];
+  var debuggerServerPort = process.argv[4];
   if (debuggerServerPort) {
     this.initServer_(debuggerServerPort);
   } else {
--- a/lib/debugger/clients/wddebugger.js
+++ b/lib/debugger/clients/wddebugger.js
@@ -1,5 +1,5 @@
 var repl = require('repl');
-var baseDebugger = require('_debugger');
+var debuggerCommons = require('../debuggerCommons');
 var CommandRepl = require('../modes/commandRepl');
 var DebuggerRepl = require('../modes/debuggerRepl');
 
@@ -11,7 +11,7 @@ var DebuggerRepl = require('../modes/debuggerRepl');
  * @constructor
  */
 var WdDebugger = function() {
-  this.client = new baseDebugger.Client();
+  this.client;
   this.replServer;
 
   // repl is broken into 'command repl' and 'debugger repl'.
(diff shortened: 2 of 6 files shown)


lib/debugger/debuggerCommons.js
```diff
@@ -0,0 +1,61 @@
...
+
+  client.once('ready', function() {
+    client.setBreakpoint({
+      type: 'scriptRegExp',
+      target: '.*executors\.js', //jshint ignore:line
+      line: 37
+    }, function() {});
+  });
+
+  process.debugPort = opt_port || 5858
```
tullmann: Missing a trailing ';' here.

lib/debugger/debuggerCommons.js
```diff
@@ -0,0 +1,61 @@
...
+  var connectWithRetry = function(attempts) {
+    var socket = client.connect(process.debugPort, host);
+    socket.on('connect', function() {
+      client.reqContinue(function() {
+        // Intentionally blank.
+      });
+    })
+    socket.on('error', function(e) {
+      if (attempts === 1) {
+        throw e
```
tullmann: Missing a trailing ';' here.

hankduan:
oops, forgot to run through linter. Made some style changes thanks @tullmann

floribon:
That indeed unlocked the `elementExplorer` (and `browser.pause()`) for me.
However, commands take a while and freeze most of the time. I'm using iojs@1.4.3 which forked node 0.10.35

hankduan:
Node versions above 0.10.30 is affected by the nodejs bug described here: https://github.com/angular/protractor/issues/1890

floribon:
Yes but I was experiencing a different bug, the debugger wouldn't work at all, it was impossible to enter any command. With these commits I fall back to that sluggish bug indeed. Using protractor 1.6.1 didn't work for me either, so just saying for any potential iojs user. Thanks for your work.

hankduan:
@juliemr when you review this, please take note of how I check the port for whether it is in use or not, as I'm not sure if that's the best way to do it.

sibeliusseraphini:
+1

lib/debugger/clients/explorer.js
```diff
@@ -9,7 +9,7 @@ var CommandRepl = require('../modes/commandRepl');
  * @constructor
  */
 var WdRepl = function() {
-  this.client = new baseDebugger.Client();
+  this.client;
```
sjelin: This line appears to do nothing and could be removed

bootstraponline:
:+1:

lib/debugger/clients/wddebugger.js
```diff
@@ -103,6 +99,11 @@ WdDebugger.prototype.initRepl_ = function() {
 
   // We want the prompt to show up only after the controlflow text prints.
   this.dbgRepl.printControlFlow_(function() {
+    // Backward compatibility: node version 0.8.14 has a number of built in
```
juliemr: I don't think we need to worry about versions <10 anymore, feel free to remove this.

lib/protractor.js
```diff
@@ -661,34 +661,30 @@ Protractor.prototype.initDebugger_ = function(debuggerClientPath, opt_debugPort)
...
-    if (self.debuggerServerPort_) {
-      args.push(self.debuggerServerPort_);
-    }
-    var nodedebug = require('child_process').fork(debuggerClientPath, args);
-    process.on('exit', function() {
-      nodedebug.kill('SIGTERM');
+  // Invoke fn if port is available.
+  var onPortAvailable = function(port, fn) {
+    var net = require('net');
+    var tester = net.connect({port: port}, function() {
```
juliemr: Can you check that the error is EADDRINUSE?

lib/protractor.js
```diff
@@ -661,34 +661,30 @@ Protractor.prototype.initDebugger_ = function(debuggerClientPath, opt_debugPort)
...
 
   var vm_ = require('vm');
+  var flow = webdriver.promise.controlFlow();
+
   var context = { require: require };
-  for (var key in global) {
-    context[key] = global[key];
-  }
-  context.list = function(locator) {
+  global.list = function(locator) {
```
juliemr: Why this change? I think it's safer to avoid polluting `global.list`

hankduan: otherwise autocomplete doesn't suggest it, since autocomplete doesn't know about context.

juliemr: I see. I don't think we need autocomplete to suggest it, since it's all of 4 characters, but I don't really care either way.

lib/protractor.js
```diff
@@ -700,9 +696,33 @@ Protractor.prototype.initDebugger_ = function(debuggerClientPath, opt_debugPort)
...
+      process.on('exit', function() {
+        nodedebug.kill('SIGTERM');
+      }); 
+    });
+  });
+  
+  var pausePromise = flow.timeout(1000, 'waiting for debugger to attach')
+      .then(function() {
+        // Necessary for backward compatibility with node < 0.12.0
+        browserUnderDebug.executeScript_('', 'empty debugger hook');
```
juliemr: Why wasn't this necessary before?

lib/debugger/clients/wddebugger.js
hankduan: Keeping this, as discussed offline.

juliemr:
Looks good, just a couple questions to make sure I understand the changes.

lib/protractor.js
hankduan: This is not an error. I'm testing that I _can_ connect to it and fail if so. 

I tried something like https://gist.github.com/timoxley/1689041 before, where I'm test creating a server to check for EADDRINUSE. However, that does not fail for some reason, until I try to create the debug process.

juliemr: Ah sorry, I meant something like this: https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/net/portprober.js#L159

Except we don't actually want to create a server, so the method you had is the best I can think of.

lib/protractor.js
hankduan: It's because the way node debugger changed in 0.12.
Before, when you start the debug process, it starts a server but the program continues (that's why you needed the sleep() in there to manually stop execution while waiting for the debugger to attach). However, now, the moment you start debug, the program stops and waits for the reqContinue signal. 

As a result of that, before, we call debug in this process. In the client, we take control once we attach.
Now, we call debug from the client. Because the program is paused, reqContinue is the first call we make. After that, we can take control as before. 

However, if we're using node 0.10, the program isn't paused when issued the debug, so the first reqContinue actually makes the program jump to the next break point.

hankduan:
merged: https://github.com/angular/protractor/commit/de499696eb88721cb073a195025ae48f59cbc905

PR #1900, opened 2015-03-05
fix(accessibility): improve output for long elements

juliemr:
Instead of just printing the first N characters of the element's template,
print the first and last N/2.

See #1854

### DIFF
--- a/plugins/accessibility/index.js
+++ b/plugins/accessibility/index.js
@@ -175,6 +175,15 @@ function runChromeDevTools(config) {
   var elementPromises = [],
       elementStringLength = 200;
 
+  function trimText(text) {
+    if (text.length > elementStringLength) {
+      return text.substring(0, elementStringLength / 2) + ' ... '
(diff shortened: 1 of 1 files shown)


plugins/accessibility/index.js
```diff
@@ -175,6 +175,15 @@ function runChromeDevTools(config) {
   var elementPromises = [],
       elementStringLength = 200;
 
+  function trimText(text) {
+    if (text.length > elementStringLength) {
+      return text.substring(0, elementStringLength / 2) + ' ... '
```
sjelin: This will have a total length of `elementStringLength + 5`.  Is that correct?

juliemr: Yeah (there's also some tabs at the beginning, so it's not the whole output string length anyway...).

sjelin:
LGTM

marcysutton:
:+1:

juliemr:
Thanks! Merged as fb92be6d588e7302989bd171a064739359bb3c74

PR #1907, opened 2015-03-06
fix(element): return not present when an element disappears

hankduan:


lib/element.js
```diff
@@ -353,7 +353,29 @@ ElementArrayFinder.prototype.toElementFinder_ = function() {
  */
 ElementArrayFinder.prototype.count = function() {
   return this.getWebElements().then(function(arr) {
-    return arr.length;
+    var list = [];
+    arr.forEach(function(webElem) {
```
tullmann: A `forEach` that ends with a list `push` is just a `map` call?   Structuring this as `list = arr.map(function(webElem) { ... })` might be cleaner.

hankduan: Changing to use map + reduce instead.

sjelin:
LGTM

hankduan:
committed as https://github.com/angular/protractor/commit/2a765c76e121de13ff86a279fe3f21cb15c17783

PR #1894, opened 2015-03-04
chore(website): add anchors to sections, apply header offset for sections, fix links

sjelin:
Closes #1806

PR #1913, opened 2015-03-08
fix path to referenceConf.js

lookfirst:


docs/server-setup.md
```diff
@@ -66,10 +66,11 @@ To run your tests against a remote Selenium Server, you will need an account wit
 In your config file, set these options:
  - `sauceUser` -  The username for your Sauce Labs account.
  - `sauceKey` -  The key for your Sauce Labs account.
-
+ - `sauceSeleniumAddress` - default: ondemand.saucelabs.com:80/wd/hub ; travis: localhost:4445/wd/hub
+ 
 Please note that if you set `sauceUser` and `sauceKey`, the settings for `seleniumServerJar`, `seleniumPort` and `seleniumArgs` will be ignored.
 
-You can optionally set the [`name` property](referenceConf.js#L113) in a capability in order to give the jobs a name on the server.  Otherwise they will just be called `Unnamed Job`.
+You can optionally set the [`name` property](/docs/referenceConf.js#L113) in a capability in order to give the jobs a name on the server.  Otherwise they will just be called `Unnamed Job`.
```
hankduan: What's the point of this? They're in the same directory.

lookfirst: 1. It matches the value with Line 10: https://github.com/lookfirst/protractor/blame/patch-1/docs/server-setup.md#L10
1. http://take.ms/jh206 <- If I'm on the website and I click the link, it fails.

Happy now?

sjelin: 1. Not a compelling argument.  I'd rather have them both relative links.
2. Thanks for bringing this bug to my attention!  I'll fix it ASAP.  But changing the markdown because the website generation script doesn't work properly isn't really the right way approach these things.

lookfirst: Sure, whatever. Just make it work dude.

docs/server-setup.md
```diff
@@ -66,10 +66,11 @@ To run your tests against a remote Selenium Server, you will need an account wit
 In your config file, set these options:
  - `sauceUser` -  The username for your Sauce Labs account.
  - `sauceKey` -  The key for your Sauce Labs account.
-
+ - `sauceSeleniumAddress` - default: ondemand.saucelabs.com:80/wd/hub ; travis: localhost:4445/wd/hub
```
hankduan: nit: sauceSeleniumAddress is not a mandatory field for users to set. I think it's good to mention this, but it should sound more like a tip for a user.

lookfirst: Which is why I mention the `default`.

docs/server-setup.md
sjelin: PR right here: https://github.com/angular/protractor/pull/1894

docs/server-setup.md
juliemr: Since most users won't need to change this, I hesitate to mention it here - and if so, we should use a full sentence to describe when you would want to change it and what 'travis' means in this context. If you'd like to do this, would you mind making it a new pull request (just so that things are separated)? Thanks!

juliemr:
Now that #1894 is in, I believe this can be closed.

PR #1916, opened 2015-03-10
Console errors and warning plugin

el-davo:
My first attempt at writing a plugin. Its usage is meant to fail when there are errors and warnings on the console of the browser.. Let me know what you guys think


plugins/console/index.js
```diff
@@ -0,0 +1,102 @@
...
+ * Logs messages to the test output
+ *
+ * @param warnings
+ *      The list of warnings detected by the browser log
+ * @param errors
+ *      The list of errors detected by the browser log
+ */
+ConsolePlugin.logMessages = function (warnings, errors) {
+    warnings.map(function (warning) {
+        console.error(warning.level.name + ': ' + warning.message);
```
tullmann: Use "console.warn" here?

plugins/console/index.js
```diff
@@ -0,0 +1,102 @@
+var q = require('q');
```
tullmann: I think you should use protractor.promise instead of q.  The webdriver code has its own promise implementation and you probably want to use the same infrastructure.

.gitignore
```diff
@@ -4,6 +4,7 @@ testapp/inbrowsertest/
 website/bower_components/
 website/build/
 website/docgen/build/
+.idea
```
tullmann: This should probably be in your personal git ignore file not the project one.

plugins/console/index.js
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Parses the log and decides whether to throw an error or not
+ *
+ * @param config
+ *      The config from the protractor config file
+ * @returns {Deferred.promise}
+ */
+ConsolePlugin.parseLog = function (config) {
+    var self = this;
+    var deferred = q.defer();
```
tullmann: I think this should use `protractor.promise.defer()`

plugins/console/index.js
```diff
@@ -0,0 +1,102 @@
...
+    var failOnWarning = config.failOnWarning || this.failOnWarning;
+    var failOnError = config.failOnError || this.failOnError;
+    this.getBrowserLog().then(function (log) {
```
tullmann: I think if you `return` here, you can avoid the explicit deferred.  (You'll be returning the promise returned by `then` which should be the same thing.)

plugins/console/index.js
```diff
@@ -0,0 +1,102 @@
...
+ * Tear-down function used by protractor
+ *
+ * @param config
+ */
+ConsolePlugin.prototype.teardown = function (config) {
+    var audits = [];
+
+    audits.push(ConsolePlugin.parseLog(config));
+
+    return q.all(audits).then(function (result) {
```
tullmann: `protractor.promise.all` is probably safer than `q.all`

tullmann:
This looks handy to me.  We do something similar with scraping through the browser logs.

I'm also not entirely confident about the `q`/`protractor.promise` thing, so you might wait for confirmation from someone with a clue about that.

In my experience Firefox browser often spits out a pile of odd/useless log messages at startup, so I need to suppress the first batch of messages.  (Otherwise the first test gets blamed for a bunch of stuff it didn't do.)

And, I also have a list of messages that I'm not interested in seeing from Firefox.  (We see a lot of "Ruleset ignored due to bad selector" messages, which is probably our app).  So, it might be worth adding the ability to provide the plugin with a list of messages to ignore.

plugins/console/index.js
juliemr: It's super confusing, but Protractor actually uses both `q` and the webdriver promises. I think it's OK to use `q` in this situation, because it has no need to sync up with the webdriver promises - and it's actually probably safer, because it can't accidentally interfere with the webdriver control flow.

juliemr:
I said this above, but I know that 'comments on outdated changes' get hard to see, so copying:

It's super confusing, but Protractor actually uses both q and the webdriver promises. I think it's OK to use q in this situation, because it has no need to sync up with the webdriver promises - and it's actually probably safer, because it can't accidentally interfere with the webdriver control flow.

plugins/console/index.js
```diff
@@ -0,0 +1,102 @@
+var q = require('q');
+
+var testOut = {failedCount: 0, specResults: []};
+
+/**
+ * This plugin scans the log after each test and can fail on error and warning messages
```
juliemr: Clarify that it scans the 'browser log'

plugins/console/index.js
```diff
@@ -0,0 +1,102 @@
...
+ *    exports.config = {
+ *      plugins: [{
+ *        path: 'node_modules/protractor/plugins/console',
+ *        failOnWarning: {Boolean}  (Default - false),
+ *        failOnError: {Boolean}    (Default - true)
+ *      }]
+ *    };
+ */
+var ConsolePlugin = function () {
+    this.failOnWarning = false;
```
juliemr: Style nits: We use 2 spaces for indent inside function blocks.

plugins/console/index.js
```diff
@@ -0,0 +1,102 @@
...
+    this.failOnError = true;
```
juliemr: Style nits: no space between `function` and `()`

plugins/console/index.js
```diff
@@ -0,0 +1,102 @@
...
+
+        var errors = log.filter(function (node) {
+            return (node.level || {}).name === 'SEVERE';
+        });
+
+        if (warnings.length > 0 || errors.length > 0) {
+            self.logMessages(warnings, errors);
+        }
+
+        testOut.failedCount += (warnings.length > 0 && failOnWarning) ? 1 : 0;
```
juliemr: It would be nice to add a message to `testOut.specResults` to show what the error is - if you do this, you might not need to separately log the messages as well. The plugin will automatically log information about failed `specResults`.

juliemr:
Thanks, I think this is generally useful and should definitely be included! I've added a couple comments about the output and some general style nitpicks, could you take a look at those?

el-davo:
Thanks for the great feedback guys. Will get the updates in asap

plugins/console/index.js
el-davo: Will use q for now as per Julies comment

el-davo:
@juliemr 
I updated as per the comments above. The Style nits are updated to 2 spaces indentation and no spaces between functions and (). 
Updated the logging to go to specResults instead of console.out.

@tullmann 
I liked your idea of having a filtering mechanism so I added an exclude parameter which can accept strings and regular expressions.

image

plugins/console/index.js
```diff
@@ -0,0 +1,127 @@
...
+  return q.all(audits).then(function(result) {
+    return testOut;
+  });
+};
+
+var consolePlugin = new ConsolePlugin();
+
+exports.teardown = consolePlugin.teardown.bind(consolePlugin);
+
+exports.ConsolePlugin = ConsolePlugin;
```
sjelin: Is this used anywhere?

plugins/console/index.js
```diff
@@ -0,0 +1,127 @@
...
+ * Tear-down function used by protractor
+ *
+ * @param config
+ */
+ConsolePlugin.prototype.teardown = function(config) {
+  var audits = [];
+
+  audits.push(ConsolePlugin.parseLog(config));
+
+  return q.all(audits).then(function(result) {
```
sjelin: style: `result` variable is unused

plugins/console/index.js
```diff
@@ -0,0 +1,127 @@
...
+  return deferred.promise;
+};
+
+/**
+ * Tear-down function used by protractor
+ *
+ * @param config
+ */
+ConsolePlugin.prototype.teardown = function(config) {
+  var audits = [];
```
sjelin: Is this `audits` variable doing much?  Could this whole function just be:

```js
ConsolePlugin.parseLog(config).then(function() {
  return testOut;
})
```

plugins/console/index.js
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Parses the log and decides whether to throw an error or not
+ *
+ * @param config
+ *      The config from the protractor config file
+ * @returns {Deferred.promise}
+ */
+ConsolePlugin.parseLog = function(config) {
+  var self = this;
+  var deferred = q.defer();
```
sjelin: You don't need this `deferred`.  Just return the promise you're currently resolving it inside.  That is:

```js
return this.getBrowserLog().then(function(log) {
  ...
});
```

plugins/console/index.js
```diff
@@ -0,0 +1,127 @@
...
+      return (e instanceof RegExp) ? logMessage.match(e) : logMessage.indexOf(e) > -1;
+    }).length === 0;
+};
+
+/**
+ * Parses the log and decides whether to throw an error or not
+ *
+ * @param config
+ *      The config from the protractor config file
+ * @returns {Deferred.promise}
```
sjelin: nit: it's `q.Promise`

plugins/console/index.js
```diff
@@ -0,0 +1,127 @@
...
+ */
+ConsolePlugin.parseLog = function(config) {
+  var self = this;
+  var deferred = q.defer();
+  var failOnWarning = config.failOnWarning || false;
+  var failOnError = config.failOnError || true;
```
sjelin: BUG: even if someone explicitly sets `config.failOnError` to `false` then this statement will force `failOnError` to be true anyway.  What you want is `config.failOnError == undefined ? true : config.failOnError`

plugins/console/index.js
el-davo: Hi Sammy

I had initially set it up like this but ran into a strange issue where the test would finish before the parseLog function finished which meant testOut.failedCount was always zero and it would never fail the test. The only way I could get it to work in order was to wrap it in a promise

plugins/console/index.js
el-davo: Hi Sammy

I dont think that is a bug. I am seeing that the tests will pass if I set failOnError to false which is correct. Correct me if i am wrong but || is short hand for undefined null or empty, so the value will be set to true only if config.failOnError is undefined. And if it is initially set to false then it is defined and the value correctly will be set to false?

plugins/console/index.js
```diff
@@ -0,0 +1,127 @@
...
+
+    var warnings = log.filter(function(node) {
+      return (node.level || {}).name === 'WARNING' && self.includeLog(node.message);
+    });
+
+    var errors = log.filter(function(node) {
+      return (node.level || {}).name === 'SEVERE' && self.includeLog(node.message);
+    });
+
+    testOut.failedCount += (warnings.length > 0 && failOnWarning) ? 1 : 0;
```
sjelin: Right now if there are 5 warning that only increases `failedCount` by 1.  I think you want `failOnWarning ? warnings.length : 0`

plugins/console/index.js
```diff
@@ -0,0 +1,127 @@
...
+    testOut.failedCount += (errors.length > 0 && failOnError) ? 1 : 0;
```
sjelin: Same comment as above

plugins/console/index.js
el-davo: Good idea will give it a shot

plugins/console/index.js
```diff
@@ -0,0 +1,127 @@
+var q = require('q');
+
+var testOut = {failedCount: 0, specResults: []};
```
sjelin: You haven't quite conformed to [the spec](https://github.com/angular/protractor/tree/master/lib/frameworks) here. I think what you want is:

```js
var testOut = {failedCount: 0, specResults = [{
   description: "Console output",
   assertions: [],
   duration: 0
}]};
```

And then, in later lines, when you write `testOut.specResults.push(` you want `testOut.specResults.assertions.push(`

plugins/console/index.js
sjelin: `||` is the boolean or operator.  Because `null`, `undefined`, and a couple other things (like `""`) are falsy, and variables default to `undefined`, the operator generally does a good job in a case like this.  However, `false` is also falsy, so it doesn't work well in this case.

I don't know how your tests as passing, but `false || true` is definitely `true`.

plugins/console/index.js
sjelin: I like the fact that it's in a promise!  I'm just not clear why you put the promise in an array

plugins/console/index.js
el-davo: You are correct. I will update this. Good catch

plugins/console/index.js
el-davo: Your correct. A mistake on my part as I originally had 2 promises being pushed onto the array. I will update this

plugins/console/index.js
```diff
@@ -0,0 +1,117 @@
...
+    });
+
+    var errors = log.filter(function(node) {
+      return (node.level || {}).name === 'SEVERE' && self.includeLog(node.message);
+    });
+
+    testOut.failedCount += (warnings.length > 0 && failOnWarning) ? 1 : 0;
+    testOut.failedCount += (errors.length > 0 && failOnError) ? 1 : 0;
+
+    if(testOut.failedCount > 0) {
```
sjelin: I think you want to output the log regardless don't you?

plugins/console/spec/consoleConfig.js
```diff
@@ -0,0 +1,17 @@
+var env = require('../../../spec/environment.js');
```
sjelin: In order to get this test to run on travis, you'll need to make an addition to [`scripts/test.js`](https://github.com/angular/protractor/blob/master/scripts/test.js#L39).  Basically change

```js
passingTests.push(
    'node lib/cli.js plugins/timeline/spec/conf.js',
    'node lib/cli.js plugins/ngHint/spec/successConfig.js',
    'node lib/cli.js plugins/accessibility/spec/successConfig.js');
```

to

```js
passingTests.push(
    'node lib/cli.js plugins/timeline/spec/conf.js',
    'node lib/cli.js plugins/ngHint/spec/successConfig.js',
    'node lib/cli.js plugins/accessibility/spec/successConfig.js',
    'node lib/cli.js plugins/console/spec/consoleConfig.js');
```

el-davo: Thanks I was looking for that.
One quick question for you, because I don't think my tests are very good. I cant think of a way to test that the tests fail without failing the test if you get me :). Any idea if this is possible?

sjelin: It is possible!  If you look at the [tests for the `ngHint` plugin](https://github.com/angular/protractor/tree/master/plugins/ngHint/spec), you'll see both `failureConfig.js` and `successConfig.js`.  Then, in the `scripts/test.js`, you see that `successConfig.js` is pushed onto `passingTests`, but if you scroll down near the bottom you see

```js

// Check ngHint plugin

executor.addCommandlineTest(
    'node lib/cli.js plugins/ngHint/spec/failureConfig.js')
    .expectExitCode(1)
    .expectErrors([{
      message: 'warning -- ngHint plugin cannot be run as ngHint code was ' +
          'never included into the page'
    }, {
      message: 'warning -- ngHint is included on the page, but is not active ' +
          'because there is no `ng-hint` attribute present'
    }, {
      message: 'warning -- Module "xApp" was created but never loaded.'
    }]);
```
This basically means "run `failureConfig.js`, and expect it to fail with these errors"

el-davo: Thats awesome, thanks for the help. Will try getting it to work this evening

plugins/console/index.js
el-davo: If I do that it will output "Console output" onto the console at the end of the test I think. That's what I am seeing. And pushing onto testOut.specResults.assertions causes an error.

However in its current form it seems to work. Maybe I am mistaken?

sjelin:
Nice!  I'll merge this in later today

el-davo:
Awesome and thanks for the help

sjelin:
Merged in at https://github.com/angular/protractor/commit/0351a1018b8ca701b37199c9c5f9c2db666ba037

jameseisenhauer:
Curious if you will be adding documentation for this plugin to the plugin docs.  Also does this work with all browsers protractor works with?

sjelin:
Good point James!  I'll add documentation ASAP.  Not sure if Travis automatically tested this in all browsers, but I imagine that it did and everything works fine.

alecxe:
Thank you very much for the feature.

From what I understand, the plugin would not work on Internet Explorer since `browser.manage().logs()` is not yet supported, see https://code.google.com/p/selenium/issues/detail?id=4925.

I've done smth similar before and I had to skip IE, see [Handling Unknown Errors in protractor](http://stackoverflow.com/questions/27587571/handling-unknown-errors-in-protractor).

alecxe:
Also [How can I get hold of the browser's console?](https://github.com/angular/protractor/blob/master/docs/faq.md#how-can-i-get-hold-of-the-browsers-console) section of the FAQ should be updated and mention the new `console` plugin.

el-davo:
Hi Alex. That's interesting to know that it doesn't work on Internet explorer. I did not realize that. In our company we don't currently run protractor tests against ie, only chrome and firefox.I had a lot of headaches trying to make the tests run stable on ie. Hopefully support will be added soon for i.e. I could put in the suggested fix you mentioned here - http://stackoverflow.com/questions/27587571/handling-unknown-errors-in-protractor But is there any point as you could just leave the plugin out of your internet explorer config?.

alecxe:
@el-davo thanks for the quick response. For us it is not a big deal too at the current project, since currently we are not automating IE, but we'll do that in the future - as a workaround, we can apply the fix suggested in that SO topic, or have a separate config. We should at least document that the plugin would not work on IE, or implicitly skip checking the logs if running against IE, or may be issue a warning..what do you think?

Also, let me know if I should create a separate issue for the following suggestion: is it possible to group the console errors and warnings that have the same text and stack trace? At the moment the test output I get has about a hundred repeated stack trace blocks, since every test I run detects the same error message on the console. Hope it is clear what I'm trying to say. Thanks.

el-davo:
Hi Alex

I like that suggestion of not having the same stack traces. No need to create a seperate issue. I can look into that in the next few days. Maybe we can do what they do in the chrome dev tools and have a counter that increases if the message is the same. For example if i have an error

WARNING (2 times) this is a warning message, line 32, column 4

Will try and add a readme also explicitly stating that i.e. is a no go for now :)

alecxe:
@el-davo exactly! Would definitely make the plugin better. Thank you, let me know if you need any help.

Speaking about the IE logging problems, the `timeline` plugin has it documented [here](https://github.com/angular/protractor/blob/9891d430aff477c5feb80ae01b48356866820132/plugins/timeline/index.js#L35) and [here](https://github.com/angular/protractor/blob/9891d430aff477c5feb80ae01b48356866820132/plugins/timeline/index.js#L250). We can have smth similar for `console` plugin.

cscleison:
From what i could see all tests fail in the event of a console error. Don't you think would be great to check console changes after each test in order to find out which interaction caused the error?

Right now we have a website with a lot of fields not localized yet so we have a lot of INFO entries in console. It gets so huge that I'm getting Java heap space errors when the plugin tries to print it:

```
[launcher] Error: UnknownError: Java heap space
    at new bot.Error (C:\Users\USER\AppData\Roaming\npm\node_modules\protracto
r\node_modules\selenium-webdriver\lib\atoms\error.js:113:18)
    at Object.bot.response.checkResponse (C:\Users\USER\AppData\Roaming\npm\no
de_modules\protractor\node_modules\selenium-webdriver\lib\atoms\response.js:106:
9)
    at C:\Users\USER\AppData\Roaming\npm\node_modules\protractor\node_modules\
selenium-webdriver\lib\webdriver\webdriver.js:377:20
    at [object Object].promise.ControlFlow.runInFrame_ (C:\Users\USER\AppData\
Roaming\npm\node_modules\protractor\node_modules\selenium-webdriver\lib\webdrive
r\promise.js:1877:20)
    at [object Object].promise.Callback_.goog.defineClass.notify (C:\Users\CSTEI
N\AppData\Roaming\npm\node_modules\protractor\node_modules\selenium-webdriver\li
b\webdriver\promise.js:2464:25)
    at [object Object].promise.Promise.notify_ (C:\Users\USER\AppData\Roaming\
npm\node_modules\protractor\node_modules\selenium-webdriver\lib\webdriver\promis
e.js:563:12)
    at Array.forEach (native)
    at Object.goog.array.forEach (C:\Users\USER\AppData\Roaming\npm\node_modul
es\protractor\node_modules\selenium-webdriver\lib\goog\array\array.js:203:43)
    at [object Object].promise.Promise.notifyAll_ (C:\Users\USER\AppData\Roami
ng\npm\node_modules\protractor\node_modules\selenium-webdriver\lib\webdriver\pro
mise.js:552:16)
    at goog.async.run.processWorkQueue (C:\Users\USER\AppData\Roaming\npm\node
_modules\protractor\node_modules\selenium-webdriver\lib\goog\async\run.js:125:21
)
[launcher] Process exited with error code 100
```

PR #1917, opened 2015-03-11
feat(element): Add getter and setter for value of input elements

uglow:

Examples:
```js

// Existing approach

// NEW approach

```

### DIFF
--- a/lib/element.js
+++ b/lib/element.js
@@ -953,6 +953,49 @@ ElementFinder.prototype.isPending = function() {
 };
 
 /**
+ * Shortcut for getting the value of any element that you would normally
+ * use element.getAttribute('value') to get it's value
+ * (e.g. <input type="text|number|...">
+ *
+ * getValue() and setValue() provide API-symmetry when working with
+ * input elements.
+ *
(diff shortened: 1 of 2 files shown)


uglow:
I've changed my git repo's user.email address to u_glow@hotmail.com, but I haven't made another commit yet. What's the best way to fix my CLA status?

sjelin:
Uh, just sign the CLA with both email addresses

uglow:
Signed CLA with other email address which matches commit's user.email.

uglow:
I authored this commit

uglow:
I am the author of this commit

Sent from my iPhone

On 12 Mar 2015, at 2:28 pm, googlebot <notifications@github.com<mailto:notifications@github.com>> wrote:


We found a Contributor License Agreement for you (the sender of this pull request) and all commit authors, but as best as we can tell these commits were authored by someone else. If that's the case, please add them to this pull request and have them confirm that they're okay with these commits being contributed to Google. If we're mistaken and you did author these commits, just reply here to confirm.

PR #1925, opened 2015-03-13
Fix repeater inline documentation.

baseonmars:
Hi, I noticed that the documentation for the column method of by.repeater doesn't match up to it's usage. 

Hope this is now correct, it seems to match the specs.


juliemr:
This looks good, thanks for noticing! Merged as 5f900160026edac6ee80c00eedd21a8c984da2c4

PR #1926, opened 2015-03-14
fix(element): update to selenium-webdriver@2.45.1 and remove element.then

juliemr:
always be a promise.

### DIFF
--- a/lib/element.js
+++ b/lib/element.js
@@ -1,4 +1,3 @@
-var util = require('util');
 var webdriver = require('selenium-webdriver');
 var log = require('./logger.js');
 var clientSideScripts = require('./clientsidescripts.js');
@@ -6,7 +5,7 @@ var clientSideScripts = require('./clientsidescripts.js');
 var WEB_ELEMENT_FUNCTIONS = [
   'click', 'sendKeys', 'getTagName', 'getCssValue', 'getAttribute', 'getText',
   'getSize', 'getLocation', 'isEnabled', 'isSelected', 'submit', 'clear',
-  'isDisplayed', 'getOuterHtml', 'getInnerHtml', 'getId'];
+  'isDisplayed', 'getOuterHtml', 'getInnerHtml', 'getId', 'getRawId'];
 
 /**
  * ElementArrayFinder is used for operations on an array of elements (as opposed
@@ -78,7 +77,6 @@ var ElementArrayFinder = function(ptor, getWebElements, locator, opt_actionResul
     };
   });
 };
-util.inherits(ElementArrayFinder, webdriver.promise.Promise);
 
 exports.ElementArrayFinder = ElementArrayFinder;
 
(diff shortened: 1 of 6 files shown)


juliemr:
I'm pretty sure the failures with `filter` are not due to the specific change here, but are another problem introduced by webdriverJS 2.45. We used to rely on the logic that was changed in https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/CHANGES.md for version 2.45.0, where already-resolved promises `.then` was called immediately.

juliemr:
@hankduan could you take a look at this idea? The only tests that I had to disable were tests which specifically made sure that `element().then` works, not tests which actually did something practical. I can't think of any good reason for doing `element().then`, and honestly I think it's just caused confusion in the past, so I'd be happy to remove it. Can you think of any reason not to?

floribon:
Just my two cents, but many promises are confusing in protractor. Instead of `element(many).then` I'd happily write `element(many).each` to be consistent with other frameworks, and indeed remove `then` as much as possible when not needed.

lib/element.js
```diff
@@ -680,6 +676,28 @@ var ElementFinder = function(ptor, elementArrayFinder) {
   this.ptor_ = ptor;
   this.parentElementArrayFinder = elementArrayFinder;
 
+  // Only have a 'this' function if the parent element array finder
```
hankduan: then function?

spec/basic/elements_spec.js
```diff
@@ -157,7 +157,7 @@ describe('ElementFinder', function() {
     expect(successful).toEqual(false);
   });
 
-  it('then function should be equivalent to itself', function() {
+  xit('then function should be equivalent to itself', function() {
```
hankduan: Can we just remove these `xit`s?

hankduan:
LGTM if we're okay with the notion of removing `then` from `element(by.xyz)`. Even though it is possible to do everything without that then, I feel like many users are going to run into a lot of issues upgrading because they're using the then.

lib/element.js
juliemr: Done.

spec/basic/elements_spec.js
juliemr: Yeah, I'll do that before submitting if this looks good.

hankduan:
Last piece of sanity check:

Is removing the inheritance going to break code that checks for promise, i.e:
https://github.com/angular/protractor/blob/master/lib/protractor.js#L332
https://github.com/angular/jasminewd/blob/master/index.js#L173
https://github.com/angular/protractor/blob/master/lib/protractor.js#L723

If that's all good, do we need this still? https://github.com/angular/protractor/blob/master/lib/element.js#L967

juliemr:
Re: sanity check

Good catch for `isElementPresent` - I added tests and updated the lib.
The jasmineWD wrappers will work because `promise.isPromise` just checks if it has a `then` function.
Same for the last protractor link.

I think we're ok to remove `isPending`. All tests are passing without it.

justusromijn:
Also an idea to update the [protractor documentation](https://angular.github.io/protractor/#/api?view=ElementFinder.prototype.then)? We were getting the `then is not a function` error, and it took us a while to figure out that support has dropped.

However on a side note, a use case for `element.then` would be if you are using the browser.executeScript, with for example the scrollIntoView with a reference to the element. Now, we have some issues with the infamous "Element not clickable at point" because we did this:

```js
    var el = element(by.model('something');
    browser.executeScript('arguments[0].scrollIntoView()', el.getWebElement()).then(function(){
      // this produces sometimes the element not clickable because of a race condition, depending
      // whether the element promise was resolved in time.
    });
```

PR #1932, opened 2015-03-16
chore(ci): move node 0.12 tests out of allowed failures

juliemr:
Test the full suite against node 0.12, and make this not an
allowed failure, now that node 0.12 is roughly stable.


juliemr:
Merged as cc13b031b08c08c84d682fd58804d5e228897ec2

PR #1934, opened 2015-03-17
docs(bower-configuration): Add .bowerrc file to make sure all the static...

Zizzamia:
... assets are going to be in the right folder.
I had the issue that the parent folder contain the file .bowerrc with a different `directory: path`, by adding
a .bowerrc inside the website folder we make sure this kind of conflict will never happen.

### DIFF
--- a/website/.bowerrc
+++ b/website/.bowerrc
@@ -1,0 +1,3 @@
+{
+  "directory" : "bower_components"
+}


sjelin:
Sorry, I'm not super familiar with bower.  Are you saying that it will search up the file tree looking for `.bowerrc` files?

Zizzamia:
Yes!
Looks that if you have `.bowerrc` in one of your parent folder, you will inherit all that configuration.

juliemr:
I think this is good then. Merging!

juliemr:
Merged as fd115c2e30d5ec6674e8039e5fdf2190fc6b671f

PR #1954, opened 2015-03-20
feat(explorer): allow element explorer to start as a server

hankduan:
If element explorer is run with a port (i.e. --debuggerServerPort 1234),
it will start up a server that listens to command from the port instead of
a repl that listens to process.stdin.

Note instead of starting either repl or server, another approach is to 
feed the server socket into the repl as the input/output streams. The 
advantage is that the process becomes much more realistic because now we're
using the normal repl. However, it was not possible to test autocomplete 
this way since we cannot immitate the TAB key over the wire.


lib/debugger/clients/explorer.js
```diff
@@ -36,38 +34,90 @@ WdRepl.prototype.initClient_ = function() {
...
+              sock.write(JSON.stringify(res) + '\r\n');
+            }
+          };
+          cmdRepl.complete(input, callback);
+        } else {
+          // Normal input
+          input = input.trim();
+          var callback = function(err, res) {
+            if (err) {
+              sock.write('ERROR: ' + err + '\r\n');
```
andresdominguez: Return here to reduce indentation:

```javascript
if(err) {
  sock ...
  return;
}
```

or

```javascript
if(err) {
  return sock...
}
```

lib/debugger/clients/explorer.js
```diff
@@ -36,38 +34,90 @@ WdRepl.prototype.initClient_ = function() {
...
+              sock.write(res + '\r\n');
+            }
+          }
+          cmdRepl.stepEval(input, callback);
+        }
+      }
+    });
+  }).listen(port);
+
+  console.log('Server listening on 127.0.0.1:'+ port);
```
andresdominguez: use comma:

```javascript
console.log('Server listening on 127.0.0.1:', port);
```

lib/debugger/clients/explorer.js
```diff
@@ -36,38 +34,90 @@ WdRepl.prototype.initClient_ = function() {
...
 };
 
 /**
- * Instantiate all repl objects, and debuggerRepl as current and start repl.
+ * Instantiate a repl to handle IO.
```
andresdominguez: Add param

```javascript
@param {string} cmdRepl The command to execute in the protractor session.
```

lib/debugger/clients/explorer.js
```diff
@@ -36,38 +34,90 @@ WdRepl.prototype.initClient_ = function() {
...
-WdRepl.prototype.initRepl_ = function() {
+WdRepl.prototype.initRepl_ = function(cmdRepl) {
   var self = this;
-  this.cmdRepl = new CommandRepl(this.client);
 
-  self.replServer = repl.start({
-    prompt: self.cmdRepl.prompt,
+  // Eval function for processing a single step in repl.
+  var stepEval = function(cmd, context, filename, callback) {
+    cmd = cmd.slice(1, cmd.length - 2);
```
andresdominguez: Can you add some comments here?

lib/debugger/clients/explorer.js
```diff
@@ -36,38 +34,90 @@ WdRepl.prototype.initClient_ = function() {
...
+        received = received.substring(i + 2);
+        if (data[0] === 0x1D) {
+          // '^]': term command
+          self.client.req({command: 'disconnect'}, function() {
+            // Intentionally blank.
+          });
+          sock.end();
+        } else if (input[input.length - 1] === '\t') {
+          // Autocomplete
+          input = input.substring(0, input.length - 1);
```
andresdominguez: Add comment here. Is the last character the ENTER key?

lib/debugger/clients/explorer.js
```diff
@@ -36,38 +34,90 @@ WdRepl.prototype.initClient_ = function() {
 };
 
 /**
- * Eval function for processing a single step in repl.
+ * Instantiate a server to handle IO.
```
andresdominguez: Add `@param` for cmdRepl, port

lib/debugger/clients/explorer.js
```diff
@@ -36,38 +34,90 @@ WdRepl.prototype.initClient_ = function() {
...
-  this.cmdRepl.stepEval(cmd, callback);
+WdRepl.prototype.initServer_ = function(cmdRepl, port) {
+  var net = require('net');
+  var self = this;
+  var received = '';
+  net.createServer(function(sock) {
+    sock.on('data', function(data) {
+      received += data.toString();
+      var i = received.indexOf('\r\n');
+      if (i >= 0) {
```
andresdominguez: flip the if to reduce indentation:

```javascript
if(i === 0) {
  return;
}
```

lib/debugger/clients/wddebugger.js
```diff
@@ -75,8 +75,16 @@ WdDebugger.prototype.stepEval_ = function(cmd, context, filename, callback) {
     this.replServer.prompt = this.currentRepl.prompt;
     this.replServer.complete = this.currentRepl.complete.bind(this.currentRepl);
     callback();
+  } else if (this.currentRepl === this.cmdRepl) {
```
andresdominguez: Add a comment here.

lib/debugger/clients/explorer.js
hankduan: using ',' puts an extra space there

andresdominguez: Then add space after ':

' + port

lib/debugger/clients/explorer.js
hankduan: done

lib/debugger/clients/explorer.js
hankduan: done

lib/debugger/clients/explorer.js
hankduan: removed param

lib/debugger/clients/explorer.js
hankduan: done

lib/debugger/clients/explorer.js
hankduan: done

lib/debugger/clients/explorer.js
hankduan: done for port. Removed cmdRepl

lib/debugger/clients/explorer.js
hankduan: done

lib/debugger/clients/wddebugger.js
hankduan: done

hankduan:
all comments addressed. PTAL

lib/debugger/clients/explorer.js
```diff
@@ -36,38 +34,100 @@ WdRepl.prototype.initClient_ = function() {
...
+            res = '';
+          }
+          sock.write(res + '\r\n');
+        }
+        cmdRepl.stepEval(input, callback);
+      }
+    });
+  }).listen(port);
+
+  console.log('Server listening on 127.0.0.1:' + port);
```
andresdominguez: I just discovered this:

```javascript
console.log('Server listening on 127.0.0.1:%s', port);
```

http://expressjs.com/starter/hello-world.html

hankduan: I'm reluctant to change the style of console.log because it's inconsistent with how we've been using it so far. Are there any functional reasons to do this?

andresdominguez: not really, but sometimes it is more readable when you have something like this:

```javascript
'foo ' + bar + ' baz'
```

lib/debugger/clients/explorer.js
```diff
@@ -36,38 +34,100 @@ WdRepl.prototype.initClient_ = function() {
...
+WdRepl.prototype.initServer_ = function(port) {
+  var net = require('net');
+  var self = this;
+  var cmdRepl = new CommandRepl(this.client);
+
+  var received = '';
+  net.createServer(function(sock) {
+    sock.on('data', function(data) {
+      received += data.toString();
+      var i = received.indexOf('\r\n');
```
juliemr: Can we change `i` to something more meaningful?

lib/debugger/clients/explorer.js
juliemr: Let's stick with our consistent style for now.

lib/runner.js
```diff
@@ -46,6 +46,10 @@ var Runner = function(config) {
     flow.timeout(1000, 'waiting for debugger to attach');
   }
 
+  if (config.debuggerServerPort) {
+    process.debuggerServerPort = config.debuggerServerPort;
```
juliemr: Is there a reason that we need to attach this to the `process` object? I know we do that within `protractor.js` for `process.debugPort`, but I don't remember if there's a compelling reason.

hankduan: Protractor doesn't have access to the config object. Alternatively, we can create a setter in Protractor and set it in runner or something, but this seems very roundabout for a non-production feature.

scripts/interactive_tests/interactive_test1.js
```diff
@@ -0,0 +1,33 @@
...
+// Check promises complete.
+test.addCommandExpectation('browser.driver.getCurrentUrl()', 'data:,');
+test.addCommandExpectation('browser.get("http://localhost:8081")');
+test.addCommandExpectation('browser.getCurrentUrl()', 
+    'http://localhost:8081/#/form'); 
+
+// Check promises are resolved before being returned.
+test.addCommandExpectation('var greetings = element(by.binding("greeting"))');
+test.addCommandExpectation('greetings.getText()', 'Hiya');
+
```
juliemr: Can we add a test that `element(by.binding("nonexistent"))` gives a reasonable error message instead of blowing up?

scripts/interactive_tests/interactive_test2.js
```diff
@@ -0,0 +1,9 @@
+var InteractiveTest = require('./interactive_test_util').InteractiveTest;
```
juliemr: Rename this file `scripts/interactive_tests/with_base_url.js`

lib/runner.js
juliemr: I'd prefer to do something a bit roundabout, but explicit, instead of adding properties to `process` randomly. Can we add a member in Protractor and just change it directly (e.g. `browser.debugPort = 8798`)

juliemr:
LG after a couple small comments. I'm very excited to have these tests!!

lib/debugger/clients/explorer.js
hankduan: done

lib/runner.js
hankduan: `process.debugPort` is actually used by node debugger to get the port to create the debugger at, so that can't be changed. 
For debuggerServerPort though, I created `browser.debuggerServerPort_`.

scripts/interactive_tests/interactive_test2.js
hankduan: done

scripts/interactive_tests/interactive_test1.js
hankduan: this actually doesn't blow up since protractor doesn't retrieve it yet. I've added this as well as `element(by.binding("nonexistent")).getText()`

hankduan:
PTAL

juliemr:
LGTM

hankduan:
merged via https://github.com/angular/protractor/commit/45341c944bf60537849776c7f362ee0442b219e6

PR #1957, opened 2015-03-23
feat(api): Add get/setValue() to element object 

uglow:

Examples:
```js

// Existing approach

// NEW approach

```


uglow:
Sorry... ignore this. Struggling with the gits and interwebs today. :(

PR #1958, opened 2015-03-23
feat(element): Add getter and setter for input elements

uglow:

Examples:
```js

// Existing approach

// NEW approach

```

[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/1958)


uglow:
googlebot, I love you, but you are giving me a lot of grief. I authored this commit.

juliemr:
@uglow noted, no need to worry about it anymore :)

lib/element.js
```diff
@@ -955,6 +955,49 @@ ElementFinder.prototype.allowAnimations = function(value) {
...
+/**
+ * Shortcut for setting the value of any element that you would normally
+ * use sendKeys() to set it's value (e.g. <input type="text|number|...">
+ *
+ * getValue() and setValue() provide API-symmetry when working with
+ * input elements.
+ *
+ * @alias element(locator).setValue('value') element(locator).clear().sendKeys('value');
+ *
+ * @param {...(string|!webdriver.promise.Promise<string>)} var_args The sequence
```
juliemr: This comment seems off, this is just treated as a string.

uglow:
@juliemr Update JSDoc to match code, and thanks for taming the googlebot :)

elgalu:
Don't see how this PR solves the inconsistency mentioned:

> This is a webdriver quirk. input and textarea elements always have empty getText values

I would expect a patch to `getText()` so that if the element is a kind of input then returns `elm.getAttribute('value')` else fallback to regular `elm.getText()`

Regarding `setValue` I think the name is misleading since internally performs a `sendKeys` .. maybe `setText` would be better? Else first user thinking might be that is doing a client side `foundElm.setAttribute("value", "some value");`

uglow:
@elgalu - you raise an interesting point, which I did consider when naming these functions. 

The reasons for choosing  get/set value instead of get/set text are:
- I wanted to have different methods for working with input & textarea elements, because they ARE different from the other kinds of elements. I see no virtue in pretending they are the same. 
- typically developers (who are the target audience of Protractor, rather then end-users) talk about the "value" of an input element, rather than the "text" of an input element. This is probably due to there being a "value" attribute on input elements.

uglow:
Any updates?

sjelin:
@juliemr We should talk about this PR before you go on sabbatical

juliemr:
Thanks for all of the work on this PR. I think that right now, this is adding some small convenience, but not enough to justify the addition to the API surface in the core Protractor library. I'd love to see this available as an additional library!

PR #1963, opened 2015-03-24
Minor improvement to the Contributing guidelines page

alecxe:


juliemr:
I actually have it as code that way so that we can highlight and copy-paste easily as a response.

alecxe:
@juliemr sorry, never mind then, it's just mine perfectionism issue. Thanks.

PR #1980, opened 2015-03-27
docs(server-setup): add info when sauce is ignored

marcin-wosinek:
It's already in the docs; but place is not obvious when we come with the question
> Why sauceUser is ignored?

### DIFF
--- a/docs/server-setup.md
+++ b/docs/server-setup.md
@@ -61,7 +61,7 @@ Please note that if you set seleniumAddress, the settings for `seleniumServerJar
 Remote Selenium Server
 ----------------------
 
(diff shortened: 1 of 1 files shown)


juliemr:
The paragraph right above says "Please note that if you set seleniumAddress, the settings for `seleniumServerJar`, `seleniumPort` and `seleniumArgs` will be ignored (it will also override the sauce options listed below)." Is this not clear enough?

marcin-wosinek:
When reading doc as a whole it's enough, but it wasn't enough for me to quickly find answer for 'why `sauceUser` is ignored?'.

Replacing 'sauce options listed below' with 'sauceUser & sauceKey' will be more search friendly.

PR #2004, opened 2015-04-03
link to api is broken

lookfirst:
not sure if this is the right fix, but the one on the site is broken right now.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -239,7 +239,7 @@ Since the Calculator reports the oldest result at the bottom, the oldest additio
 
 Fix the test so that it correctly expects the first history entry to contain the text "3 + 4".
 
-ElementArrayFinder also has methods `each`, `map`, `filter`, and `reduce` which are analogous to JavaScript Array methods. [Read the API for more details](#/api?view=ElementArrayFinder).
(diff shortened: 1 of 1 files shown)


docs/tutorial.md
```diff
@@ -239,7 +239,7 @@ Since the Calculator reports the oldest result at the bottom, the oldest additio
 
 Fix the test so that it correctly expects the first history entry to contain the text "3 + 4".
 
-ElementArrayFinder also has methods `each`, `map`, `filter`, and `reduce` which are analogous to JavaScript Array methods. [Read the API for more details](#/api?view=ElementArrayFinder).
+ElementArrayFinder also has methods `each`, `map`, `filter`, and `reduce` which are analogous to JavaScript Array methods. [Read the API for more details](/#/api?view=ElementArrayFinder).
```
sjelin: Nice catch!  We probably should just link to the absolute URL here (i.e. `http://angular.github.io/protractor/#/api?view=ElementArrayFinder`) though.  The issue is that the link needs to work in both github's markdown and the website, and since the API docs don't exist on github, a relative link isn't possible.

juliemr:
I've fixed this up - thanks for submitting the pr!

PR #2010, opened 2015-04-07
test(config): use env.webServerDefaultPort and env.interactiveTestPort

elgalu:
* Keep using `env.webServerDefaultPort` instead of hardcoded port **8081**
* Extract port **6969** into `env.interactiveTestPort`


juliemr:
Thanks! merged as 4a75aca21349cb2025edac87452fbe546c526bd9

PR #2015, opened 2015-04-07
Updates docs

scottsword:
I found "If set" to be too vague. As a reader / user of this tool I had to use trial error to determine how to use this. Now it is explicit. Thanks!


sjelin:
That's how we phrase everything else in the config, so I think it's actually more confusing to elaborate.  Makes you think there's some other variable out there that you have to hunt down and this one will interact with

PR #2024, opened 2015-04-09
Document how to use jasmine2

pbrianmackey:
Documentation should show how to use the newer jasmine v2;


juliemr:
I rewrote the entire section a bit, incorporating this advice. Thank  you!

Issue #1051, opened 2014-07-08
SafariDriver occasionally times out when loading a page


juliemr:
Using `browser.get` on SafariDriver occasionally (<1%) causes a timeout (this manifests as a message 'Timed out waiting for page to load'). 

So far, I have only observed this when running on SauceLabs.

juliemr:
As of Safari 10, SafariDriver is now shipped by Apple with Safari (see the post at https://webkit.org/blog/6900/webdriver-support-in-safari-10/). I'm closing old external safari issues because they are now obsolete.

PR #2028, opened 2015-04-09
fix(navigation): fix timeout using browser.get with safari driver

eduardomb:
```Protractor.get``` was not working on latest Safari (8.04) on latest OS X (10.10.2). Not sure if this is related but I had to manually install SafariDriver 2.45 downloaded at http://selenium-release.storage.googleapis.com/2.45/SafariDriver.safariextz

I figured the problem was safari could not execute script ```window.location.replace(...)``` after calling ```driver.get('about:blank')```. It used to work in the past and I think it might be an issue with the latest SafariDriver.

In order to get this working I've refactored ```Protractor.get```. I've removed the initial ```driver.get``` call to **resetUrl** (*about:blank* or *data:text/html,\<html\>\</html\>*). I believe the **resetUrl** was being called only to help detecting the "page load event" by the checking the moment when ```window.location.href``` becomes different from **resetUrl**. In order to detect this moment I now keep a reference to an old element on the page and try to interact with it until *webdriver* throws a *StaleElementReferenceError*, which means the element no longer exists and the new page is loaded. I've read about this new page load approach on http://www.obeythetestinggoat.com/how-to-get-selenium-to-wait-for-page-load-after-a-click.html

It's working on Safari, Chrome and Firefox. No idea about IE.

### DIFF
--- a/lib/protractor.js
+++ b/lib/protractor.js
@@ -15,7 +15,6 @@ var ExpectedConditions = require('./expectedConditions.js');
 /* global angular */
 
 var DEFER_LABEL = 'NG_DEFER_BOOTSTRAP!';
-var DEFAULT_RESET_URL = 'data:text/html,<html></html>';
 var DEFAULT_GET_PAGE_TIMEOUT = 10000;
 
 /*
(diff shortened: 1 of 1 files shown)


eduardomb:
CLA signed

eduardomb:
This might fix https://github.com/angular/protractor/issues/1051

eduardomb:
I guess my approach does not work when angular's html5 mode is disabled and the test tries to get the same URL as the previous one. I'll fix it and amend the commit soon

eduardomb:
It turned out I couldn't find a solution that works for safari with html5mode disabled.

I'll close this PR but will keep this fork alive because is a temp fix that enables test in safari for my specific app, which uses html5mode. It might help others too, but be aware the tests will fail in any browser if the app uses html5Mode.

I wonder if anyone else is able to test in safari with the latest stable protractor. Is there any open issue for this?

PR #2029, opened 2015-04-09
tests(element): add a test that fails when finding subelements within ea...

juliemr:
...ch

[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/2029)

### DIFF
--- a/spec/basic/elements_spec.js
+++ b/spec/basic/elements_spec.js
@@ -400,6 +400,21 @@ describe('ElementArrayFinder', function() {
     });
   });
 
+  it('should allow accessing subelements from within each', function() {
+    browser.get('index.html#/form');
+    var rows = element.all(by.css('.rowlike'));
+
+    rows.each(function(row) {
+      var input = row.element(by.css('.input'));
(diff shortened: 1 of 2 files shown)


juliemr:
WIP - this does not yet fix the problem, just shows it.

juliemr:
Should be fixed when #2245 is done.

hankduan:
FYI: in your original example (from the PR), if you replace the protractor-defined locator i.e. `by.binding('username')` by a webdriver-native locator i.e. `by.css('#animals')`, then the example will hang for both selenium-webdriver 2.46.1 and lower versions.

hankduan:
Reported reduced case here: https://github.com/SeleniumHQ/selenium/issues/715

juliemr:
Updating this, now that we're using selenium-webdriver 2.47

hankduan:
Is this still an issue with 2.48.2?

juliemr:
Sorry for the delay - yes, this is working now. I'll submit to make sure we're testing for it in the future.

juliemr:
Merged with 586d634bcd306a3226fd4c55bef43155294bad4e

PR #2034, opened 2015-04-10
Documentation: Homepage: fix example script

hppycoder:
Updated the homepage example with one that will work given the changes on https://angularjs.org


juliemr:
Thanks! Merged as e89db5976397101c089b76536d286209d601763a

Issue #2026, opened 2015-04-09
Commands executed out of order  in `each` sometimes


juliemr:
So far, this is the minimal case we've been able to reproduce on: #2025

```js
describe('failing each order', function() {
  it('should get each', function() {
    console.log('');
    // This fails.
    browser.get('http://localhost:8081');

    var stuffs = element.all(by.css('.rowlike'));
    stuffs.each(function(stuff) {
      var input = stuff.element(by.css('.teststuff'));
      input.getAttribute('value').then(function(value) {
        // This fails for the first value, with value='1030'
        console.log('in expect, value = ' + value);
      });
      input.sendKeys('30').then(function() {
          console.log('After sendKeys');
        });
    }).then(function() {
      console.log('Done with each');
    });
  });
});
```

The first 'sendKeys' happens before the first 'getAttribute'.

juliemr:
@hankduan if you have time, do you have any ideas for this one? I think it's something to do with the way the elementfinder/elementarrayfinder is interacting with each, but haven't been able to dig in enough to find more info yet.

hankduan:
I'll play with this either tonight or tomorrow.

hankduan:
@juliemr Can you paste your output?
I'm getting the following which means everything is working as expected

```
/workspace/protractor on master*
$ node lib/cli.js /workspace/protractor/spec/hankConf.js --suite hank 
Using the selenium server at http://localhost:4444/wd/hub
[launcher] Running 1 instances of WebDriver
Started

in expect, value = 10
After sendKeys
in expect, value = 20
After sendKeys
Done with each
.


1 spec, 0 failures
Finished in 1.128 seconds
[launcher] 0 instance(s) of WebDriver still running
[launcher] chrome #1 passed
```

juliemr:
Output:

```js
[each] - arr.forEach 0
[elementArrayFinder] + getAttribute
[elementArrayFinder] + sendKeys
[each] - arr.forEach 1
[elementArrayFinder] + getAttribute
[elementArrayFinder] + sendKeys
Done with each
[elementArrayFinder] + getAttribute inner

[elementArrayFinder] + sendKeys inner

After sendKeys
in expect, value = 1030
[elementArrayFinder] + getAttribute inner

in expect, value = 10
[elementArrayFinder] + sendKeys inner

After sendKeys
```
This is the exact code in pr #2025 right now, with some of the extra logging cut out.

juliemr:
[...]

PR #2038, opened 2015-04-10
refactor(element): make each just forward to map

juliemr:
ElementArrayFinder.each is just a call to ElementArrayFinder.map
without gathering the return values - reduce logic by making it so.

See https://github.com/angular/protractor/issues/2026


juliemr:
Merged as ca39eebf02229ce0fe09eb21a90a9f834eb40851

Issue #124, opened 2013-09-28
Is there a way to specify a proxy server in the Config?


sheldonhall:

jonniespratley:
This needs to be re-opened, here at GE Software, setting the http_proxy and https_proxy in the configuration still does not allow Protractor to send Saucelabs the final exit code of the test results, making Saucelabs fail the test due to inactivity.

hankduan:
@jonniespratley does `protractor config --proxy=http://user:password@proxy_url:port` not work for you?

chitwant82:
If I use Sauce labs sauce-connect script, before running scripts I am able to run localhost url on Sauce lab machine

Successful case Logs:

23:48:48.669 INFO - Launching a standalone server
Setting system property webdriver.firefox.logfile to c:\log\firefox.log
Setting system property webdriver.server.session.timeout to 0
Setting system property http.proxyHost to maki79213.miso
Setting system property http.proxyPort to 33128
Setting system property http.nonProxyHosts to "localhost|127.0.0.1"
Setting system property https.proxyHost to maki79213.miso
Setting system property https.proxyPort to 33128
Setting system property https.nonProxyHosts to "localhost|127.0.0.1"
23:48:48.887 INFO - Java: Oracle Corporation 25.65-b01
23:48:48.887 INFO - OS: Windows Server 2008 R2 6.1 x86
23:48:48.908 INFO - v2.45.0, with Core v2.45.0. Built from revision 5017cb8
23:48:49.153 INFO - RemoteWebDriver instances should connect to: http://127.0.0.1:4443/wd/hub
23:48:49.155 INFO - Version Jetty/5.1.x
23:48:49.158 INFO - Started HttpContext[/selenium-server,/selenium-server]
23:48:49.254 INFO - Started org.openqa.jetty.jetty.servlet.ServletHandler@c6a292
23:48:49.254 INFO - Started HttpContext[/wd,/wd]
23:48:49.254 INFO - Started HttpContext[/selenium-server/driver,/selenium-server/driver]
23:48:49.254 INFO - Started HttpContext[/,/]
23:48:49.260 INFO - Started SocketListener on 0.0.0.0:4443
23:48:49.260 INFO - Started org.openqa.jetty.jetty.Server@11bc7ed
[...]

PR #2040, opened 2015-04-13
feat(sauce-lab): Pass custom agent to SauceLabs from protractor config

naumovs:
Hello,

This PR for the saucelabs driver running behind proxy.

#124 does not work for this, because SauceLab.js issues `https.request()`, which does not use environment proxy variables and issues a call ti saucelab.com directly, leading to `Sauce pass/fail status: 'Could not send request: connect ETIMEDOUT'` error.

With this PR now in protractor configuration file I can specify

```js
var HttpsProxyAgent = require("https-proxy-agent");
 
var agent = new HttpsProxyAgent('http://<user>:<password>@<proxy host>:<port>');
 
exports.config = {
    agent: agent,
    // other things
};
```

And last request to saucelabs.com will not cause tests to fail.

based on guide by [Amber Kaplan](http://sauceio.com/index.php/2015/03/repost-angular-protractor-sauce-connect-launched-from-gulp-all-behind-a-corporate-firewall/)

### DIFF
--- a/lib/driverProviders/sauce.js
+++ b/lib/driverProviders/sauce.js
@@ -56,7 +56,8 @@ SauceDriverProvider.prototype.setupEnv = function() {
   var deferred = q.defer();
   this.sauceServer_ = new SauceLabs({
     username: this.config_.sauceUser,
-    password: this.config_.sauceKey
+    password: this.config_.sauceKey,
+    agent: this.config_.sauceAgent
   });
   this.config_.capabilities.username = this.config_.sauceUser;
   this.config_.capabilities.accessKey = this.config_.sauceKey;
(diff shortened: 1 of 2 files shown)


naumovs:
I signed it!

naumovs:
I signed it!

Seemed like I had different full name in my github account

lib/driverProviders/sauce.js
```diff
@@ -56,7 +56,8 @@ SauceDriverProvider.prototype.setupEnv = function() {
   var deferred = q.defer();
   this.sauceServer_ = new SauceLabs({
     username: this.config_.sauceUser,
-    password: this.config_.sauceKey
+    password: this.config_.sauceKey,
+    agent: this.config_.agent
```
juliemr: Can you switch this variable name to `sauceAgent`? Many settings are shared in the config so just having something called 'agent' could be confusing.

Also, please add documentation in `docs/referenceConf.js` for the new field.

naumovs:
Thanks for the feedback. I've renamed property and documented it in `docs/referenceConf.js`

Issue #2042, opened 2015-04-14
taskId using extended ascii char which is invisible when 


quiz1991:
![screen shot 2015-04-13 at 5 58 49 pm](https://cloud.githubusercontent.com/assets/5345051/7128444/1399f056-e207-11e4-8483-e99d6461713a.png)
I have more than 128-97 = 31 spec files to run for single capabilities, in which case the following code
```javascript
if (queue.specLists.length > 1) {
    taskId += String.fromCharCode(97 + queue.specsIndex); //ascii 97 is 'a'
}
```
would create a taskId with extended ascii chars and these chars are invisible in our screwdriver which causes lots of confusion.

Maybe create taskId like "#1aa" instead?

juliemr:
I knew supporting sharding was a bad idea :)

quiz1991:
Thanks @juliemr !

juliemr:
2.1 is out.

PR #2054, opened 2015-04-16
fix(taskscheduler): label sharded tasks with numbers instead of letters

juliemr:
Letters run into a problem with a maximum of 26.
See #2042


juliemr:
Closed via e599cf301d1ff0fe705b2362fa6b4b17859ab8da

Issue #2051, opened 2015-04-15
[Jasmine 2] Results of assertions are not getting added in protractor json output file


Raghulraj:
Hi,

I am using "resultJsonOutputFile" to write my test results in json format. Recently i have migrated to Jasmine2.0, but after that, assertions results are always empty in the json output.

eg below:

[
  {
    "description": "my_test",
    "assertions": [],
    "duration": 6811
  }
]

I have tried upgrading jasmine-expect 2.0.0-beta1 , but still getting same result.

juliemr:
We have many tests that use jasmine2 and resultJsonOutputFile, so this is nota  general issue. I'm not familiar with the `jasmine-expect` module, but that may be causing your issues.

If you can create a reproducible example that shows your problem, please feel free to open up a new issue. Otherwise, I'm closing as not reproducible.

Raghulraj:
Hi Julie Ralph,

The issue was with assertions for only passed test cases. failed cases are perfectly fine. Before adding Jasmine2 as framework in config, i used to get assertions results for passed ones too. eg: below
```
[
    {
        "description": "test1",
        "assertions": [
            {
                "passed": true
            }
        ],
        "duration": 43041
    }
]
```
But after adding framework: jasmine2, the resultjsonoutput looks like this.
```
[
    {
        "description": "test1",
        "assertions": [],
        "duration": 12251
    },
    {
        "description": "test2",
        "assertions": [
            {
                "passed": false,
                "errorMsg": "Failed: No element found using locator: By.cssSelector(\".search-box-desktop1\")",
                "stackTrace": "Error: Failed: No element found using locator: By.cssSelector"
            }
        ],
        "duration": 14724
    }
]
```
Only changes I have added is framework change in config. So I am confused with output now.

juliemr:
That's true, there is an inconsistency here. Jasmine2 has a different system for what gets passed to reporters, so the output has changed slightly. I'll make an update, but is there any reason this is a problem? No assertions means everything passes.

Raghulraj:
Thanks @juliemr for the quick fix. I had written an custom reporter for my test executions that relies on assertions. I have considered no assertions  as invalid test cases. So i need to have assertion results in the result output.

PR #2055, opened 2015-04-16
fix(jasmine2): be consistent about passing assertions in output JSON

juliemr:
See #2051


sjelin:
Merged in at https://github.com/angular/protractor/commit/815ff5d092c8dda88e2a1d4c5c80f66bf20892fa

PR #2067, opened 2015-04-20
Return a promise that handles errors in protractor.get

wiggzz:
If an error occurs during the protractor.get() call, that error cannot be handled.  This is an issue if you want to do cleanup after a failed test, for example.

This change ensures that on failure of any of the calls in protractor.get(), the returned promise will be resolved with an error, rather than the error not being handled.


wiggzz:
CLA is now signed.

sjelin:
Merged in at 7b96db0

PR #2070, opened 2015-04-20
(cucumber) process no-snippets param for cucumber framework

karthiktv006:


karthiktv006:
I signed it!

karthiktv006:
I signed again.

sjelin:
The issue is that the email address of author of the commits is not the same as that of your github account.  If you could sign with whatever email the commits were authored with that would be ideal, but just confirming that you are, in fact, the author would be sufficient.

sjelin:
Merged in at e18d499

PR #2089, opened 2015-04-27
chore(docs): specify that the console plugin only works on chrome

### LINKED ISSUE #2058: Console plugin does not fail on error on Firefox
I'm trying to use the console plugin with Firefox and it seems to work only partially. The only thing that works is logs filtering (the `exclude` option), but none of `failOnError`, `failOnWarning` doesn't seem to work properly, they work when I tested on Chrome.

sjelin:
Closes #2058


juliemr:
LGTM

sjelin:
Merged in at b57f242

PR #2090, opened 2015-04-27
bug(locators): excapt query in byExactBinding

### LINKED ISSUE #1918: exactBinding fails to locate bindings which contain certain characters
bindings which use characters that have special meaning in regular expressions will fail to work in By.exactBinding - the binding needs to be escaped before matching. It is deceiving because of the way you have the argument described:

```javascript
@param {string} binding The binding, e.g. {{cat.name}}.
```

Example: 
```html
<span>{{ vehicle.garagingTerritory.name() }}</span>
```

passing in 'vehicle.garagingTerritory.name()' to By.exactBinding will not match the element. You must at least escape the parenthesis. I would suggest escaping all of the input automatically, or adjusting the documentation to describe when the user needs to escape the input themselves.

Thanks

sjelin:
See http://stackoverflow.com/q/3561711

Closes #1918

### DIFF
--- a/lib/clientsidescripts.js
+++ b/lib/clientsidescripts.js
@@ -72,7 +72,10 @@ functions.findBindings = function(binding, exactMatch, using, rootSelector) {
     if (dataBinding) {
       var bindingName = dataBinding.exp || dataBinding[0].exp || dataBinding;
       if (exactMatch) {
-        var matcher = new RegExp('({|\\s|^|\\|)' + binding + '(}|\\s|$|\\|)');
+        var matcher = new RegExp('({|\\s|^|\\|)' +
+            // See http://stackoverflow.com/q/3561711
+            binding.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&") +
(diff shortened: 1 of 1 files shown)


bootstraponline:
:+1:

lib/clientsidescripts.js
```diff
@@ -72,7 +72,10 @@ functions.findBindings = function(binding, exactMatch, using, rootSelector) {
     if (dataBinding) {
       var bindingName = dataBinding.exp || dataBinding[0].exp || dataBinding;
       if (exactMatch) {
-        var matcher = new RegExp('({|\\s|^|\\|)' + binding + '(}|\\s|$|\\|)');
+        var matcher = new RegExp('({|\\s|^|\\|)' +
+            // See http://stackoverflow.com/q/3561711
```
juliemr: Single line comments inside client side scripts can screw things up - see the note at the top of this file. Make this `/* */` style.

juliemr:
One comment, then LGTM.

sjelin:
Merged in at 3cded9b

PR #2128, opened 2015-05-11
fix: Adapt test to changes in angularjs homepage

mattiash:


mattiash:
This became way too complicated. I edited the file in githubs web-interface, which means that it got my github-specific mail-address, which is not included in the cla I signed earlier and I don't want to add it to my Google account. I'll just close the pull-request. Feel free to steal the patch.

PR #2129, opened 2015-05-11
Page-object sample should not expose page internals

soygul:
Page objects should wrap all page elements and should not expose element objects directly. This PR fixes the documentation on that to prevent confusion.

Nicely summarized here: https://code.google.com/p/selenium/wiki/PageObjects#Summary


sjelin:
Merged in at 4b55091

juliemr:
Not a problem, but FYI I actually kind of liked this the way it was before. I think it's fine to do it either way - from my point of view, adding a `getter` unnecessarily complicates things. I think this is a style choice that should be up to what your team prefers.

PR #2147, opened 2015-05-16
Add additional configuration options to example/conf.js

itsryandrake:
I added these two options into the example as I thought they made more sense if someone was looking at immediately setting up Protractor to test multiple browsers, and configure a different framework.

### DIFF
--- a/example/conf.js
+++ b/example/conf.js
@@ -7,6 +7,14 @@ exports.config = {
     'browserName': 'chrome'
   },
 
+  // If you would like to test against multiple browsers, use the multiCapabilities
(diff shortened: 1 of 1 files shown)


itsryandrake:
I signed it.

deveshmaheshwari:
:+1:

sjelin:
The `example/` folder is supposed to be a minimal example of some protractor tests.  But in real protractor tests you wouldn't use `capabilities` and `multiCapabilities` at the same time.

`docs/referenceConf.js` is where we have more complete documentation of everything you can have in a config file.

PR #2161, opened 2015-05-19
fix(test): fix chat app race condition

hankduan:
If two users send a message within 100ms of each other, messages
could be dropped


juliemr:
Merged as b58b29681048466258f0250db30bd9c4617d63d1

PR #2164, opened 2015-05-20
chore(website): fix API section of website

### LINKED ISSUE #2163: Website is failing with new version
Failing to parse JSON. This has something to do with the by.xpath description:

```
  {
    "description": "<code ng-non-bindable>\&quot;&#x2F;&#x2F;div\&quot;</code>",
    "other": "bar"
  }
```
is not valid JSON.

juliemr:
Do not try to replace special character quotes in JSON.

Closes #2163

### DIFF
--- a/website/docgen/inline_tags/code.js
+++ b/website/docgen/inline_tags/code.js
@@ -2,8 +2,6 @@ var htmlSpecialCharMap = {
   "&": "&amp;",
   "<": "&lt;",
   ">": "&gt;",
-  '"': '&quot;',
-  "'": '&#39;',
   "/": '&#x2F;'
 };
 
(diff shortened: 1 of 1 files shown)


juliemr:
@sjelin was there a reason we needed to escape quotes in the first place?

sjelin:
I guess not.  I think I was coping code from stack overflow

sjelin:
(Sorry, closed accidentally)

On a different note, I think this will work, but I don't think it's the "right" solution.  It looks like what's going on is that before the `@code` inline processor gets to it, the `"`s are escaped to `\"`s, resulting in `\&quot;`s, which is the problem.  The core issue here seems like the order of escaping, and I'm a little bit concerned that this might be more masking the problem than solving it, but I guess it's fine.

juliemr:
Thanks for taking a look. @sjelin can you open up a new issue to take a look at the order of escaping?

PR #2179, opened 2015-05-26
To support grep option at suite level.

ammanyer:
Hello, 
My team at Fidelity want to have a filler on the suite level as it already exist for spec level. I already did a pull request for the jasmine changes.
I also did a change to apply case insensitive mach on regExp.

Thanks,
Ammanuel

[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/2179)

### DIFF
--- a/lib/frameworks/jasmine2.js
+++ b/lib/frameworks/jasmine2.js
@@ -74,17 +74,45 @@ exports.run = function(runner, specs) {
   var reporter = new RunnerReporter(runner);
   jasmine.getEnv().addReporter(reporter);
 
-  // Filter specs to run based on jasmineNodeOpts.grep and jasmineNodeOpts.invert.
-  jasmine.getEnv().specFilter = function(spec) {
+  // Filter specs/suite to run based on jasmineNodeOpts.grep and jasmineNodeOpts.invert.
+  jasmine.getEnv().specFilter = function(spec, suite) {
+    var suiteMatched = isSuiteMatched(suite);
+
+    //If it matched at suite level and invert is true, all the specs will not be included.
+    if(suiteMatched && jasmineNodeOpts.invertGrep){
+      spec.pend();
+      return true;
+    }
+
+    //If it matched at suite level and invert is false, all the specs will be included.
+    if(suiteMatched && !jasmineNodeOpts.invertGrep){
+      return true;
+    }
+
+    var specWithSuiteFullName = spec.getFullName();
+    if(suite){
+      specWithSuiteFullName += suite.getFullName();
+    }
+
     var grepMatch = !jasmineNodeOpts ||
         !jasmineNodeOpts.grep ||
-        spec.getFullName().match(new RegExp(jasmineNodeOpts.grep)) != null;
+        specWithSuiteFullName.match(new RegExp(jasmineNodeOpts.grep, 'i')) != null;
     var invertGrep = !!(jasmineNodeOpts && jasmineNodeOpts.invertGrep);
     if (grepMatch == invertGrep) {
       spec.pend();
     }
     return true;
   };
+
+  function isSuiteMatched(suite) {
+    if (!suite) {
+      return false;
+    }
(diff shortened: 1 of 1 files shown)


ammanyer:
I already submitted the CLA.

lib/frameworks/jasmine2.js
```diff
@@ -74,11 +74,29 @@ exports.run = function(runner, specs) {
...
+
+    var specWithSuiteFullName = spec.getFullName();
+    if(suite){
+      specWithSuiteFullName += suite.getFullName();
+    }
+
     var grepMatch = !jasmineNodeOpts ||
         !jasmineNodeOpts.grep ||
-        spec.getFullName().match(new RegExp(jasmineNodeOpts.grep)) != null;
+        specWithSuiteFullName.match(new RegExp(jasmineNodeOpts.grep, 'i')) != null;
```
sjelin: This (both adding the suite name and changing to case-insensitivity) is a breaking change.

lib/frameworks/jasmine2.js
```diff
@@ -74,11 +74,29 @@ exports.run = function(runner, specs) {
   var reporter = new RunnerReporter(runner);
   jasmine.getEnv().addReporter(reporter);
 
-  // Filter specs to run based on jasmineNodeOpts.grep and jasmineNodeOpts.invert.
-  jasmine.getEnv().specFilter = function(spec) {
+  // Filter specs/suite to run based on jasmineNodeOpts.grep and jasmineNodeOpts.invert.
+  jasmine.getEnv().specFilter = function(spec, suite) {
+    var suiteMatched = isSuiteMatched(suite);
```
sjelin: This whole section is also a breaking chance.  And what if people want to filter by spec but not suite?

lib/frameworks/jasmine2.js
```diff
@@ -74,11 +74,29 @@ exports.run = function(runner, specs) {
...
+
+    //If it matched at suite level and invert is true, all the specs will not be included.
+    if(suiteMatched && jasmineNodeOpts.invertGrep){
```
sjelin: This invert logic is just wrong.  If the suite matches but `invertGrep` is `false` then we don't check the spec?  If the suite doesn't match, we don't do any filtering even if `invertGrep` is `false`?

lib/frameworks/jasmine2.js
```diff
@@ -86,6 +104,16 @@ exports.run = function(runner, specs) {
     return true;
   };
 
+  function isSuiteMatched(suite) {
+    if (!suite) {
```
sjelin: Should this `if` be here?  Shouldn't we just skip all suite checks if `suite == undefined`?

sjelin:
We want the interface for the `grep` option to match that [of Mocha's `grep` option](http://mochajs.org/#grep-option).  Our current implementation falls short of Mocha's because it doesn't check suite names.  If you could change your implementation so that our suite-level checks worked like theirs, that would be ideal.

sjelin:
(also: you need at least one test)

lib/frameworks/jasmine2.js
ammanyer: I will remove the change for the case-insensitivity.

lib/frameworks/jasmine2.js
ammanyer: we can additional option like jasmineNodeOpts.ignoreSuite, jasmineNodeOpts.ignoreSpec. What is your thought to support this?

lib/frameworks/jasmine2.js
ammanyer: if the filter is applied at suite level, it means that it will apply for all specs under the Suite.

lib/frameworks/jasmine2.js
ammanyer: Yes. Its purpose is to skip suite checks and we have to return false then the spec filter will not consider the suite check.

sjelin:
Closing as inactive.  Moved to issue #2337

Issue #2112, opened 2015-05-05
Customizing the name when running parallel tests


willchen90:
I have been using Protractor to run my test suite in parallel for various languages. For example, I will run 3 instances of Chrome with the browser language set to English, Spanish, or French. Currently, in the output from Protractor it states `"Chrome #1"`, `"Chrome #2"`, or so on. I would like to customize the output and pass it in as a config property like "Chrome - French Canadian". 

This process name is defined by the variable `shortName` in the reportSummary function. Please see the source code here:
https://github.com/angular/protractor/blob/9891d430aff477c5feb80ae01b48356866820132/lib/launcher.js#L68

I have included a screenshot of this (I am using the Cucumber options with Protractor):
![screen shot 2015-04-30 at 3 24 32 pm](https://cloud.githubusercontent.com/assets/7344640/7484854/a29aedfe-f342-11e4-8a16-3d3e78c24f3b.png)

I would be happy to work on adding this feature (and related testing) but wanted to check with the core team and see if there is interest in adding this functionality before I spent time to create a PR.

sjelin:
Sounds good to me

ghu21st:
Customized name for parallel processes would be a useful feature for Protractor test. 
For my project, I need to run a test with 15 Chrome instances concurrent run parallel, and each with different DOM manipulation test. It would be nice to give each instances with specific name instead of 'Chrome #1', "Chrome #2"... etc.

eduncan26:
@sjelin I don't see any unit tests for the launcher or taskLogger. Do you need tests with this PR?

PR #2195, opened 2015-06-01
feat(capabilites): custom instance name

eduncan26:
Resolves angular/protractor#2112.

Naming of instances is now supported by leveraging the "name" option of the capability. This was previously not used by protractor itself but was used to pass to SauceLabs.

Documents updated.

Tested manually


eduncan26:
I have signed the CLA

PR #2196, opened 2015-06-01
feat(capabilites): custom instance name

eduncan26:
Resolves angular/protractor#2112.

Naming of instances is now supported by leveraging the "name" option of the capability. This was previously not used by protractor itself but was used to pass to SauceLabs.

Documents updated.

Tested manually

### DIFF
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -108,9 +108,9 @@ exports.config = {
   capabilities: {
     browserName: 'chrome',
 
-    // Name of the process executing this capability.  Not used directly by
-    // protractor or the browser, but instead pass directly to third parties
-    // like SauceLabs as the name of the job running this test
+    // Name of the process executing this capability. Primarily used by third
+    // parties like SauceLabs, but is also used by protractor to give a
+    // capability instance a user defined name.
     name: 'Unnamed Job',
 
     // Number of times to run this set of capabilities (in parallel, unless
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -114,7 +114,8 @@ exports.config = {
     name: 'Unnamed Job',
 
     // User defined name for the capability that will display in the results log
-    logName: 'Chrome - French Canadian'
+    // Default is 'Chrome #1', 'Firefox #2', etc.
+    logName: 'Chrome - French Canadian',
 
     // Number of times to run this set of capabilities (in parallel, unless
     // limited by maxSessions). Default is 1.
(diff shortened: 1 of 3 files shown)


sjelin:
I don't like overloading the meaning of the `name` property.  Why not create a new property for this purpose?

eduncan26:
I originally did create a new property called instanceName, but I thought
the name property was a relevant property that already exists that
protractor doesn't really use.

If you would prefer, I can create a new property and utilize that instead
of potentially making the name property too ambiguous.

sjelin:
I'd prefer a new property.  Someone might want their job on Sauce Labs with a custom, but want to keep Protractor's console output the same.  Plus, people who are already using the `name` property would be confused if suddenly their output changed.

docs/referenceConf.js
```diff
@@ -113,6 +113,9 @@ exports.config = {
     // like SauceLabs as the name of the job running this test
     name: 'Unnamed Job',
 
+    // User defined name for the capability that will display in the results log
```
sjelin: Mention the default

lib/launcher.js
```diff
@@ -71,8 +71,10 @@ var taskResults_ = {
     this.results_.forEach(function(result) {
       var capabilities = result.capabilities;
       var shortName = (capabilities.browserName) ? capabilities.browserName : '';
+      shortName = (capabilities.logName) ? capabilities.logName
+        : (capabilities.browserName) ? capabilities.browserName : '';
       shortName += (capabilities.version) ? capabilities.version : '';
-      shortName += (' #' + result.taskId);
+      shortName += (capabilities.logName && capabilities.count < 2) ? '' : ' #' + result.taskId;
```
sjelin: nit: this line is too long

lib/taskLogger.js
```diff
@@ -55,13 +55,11 @@ TaskLogger.prototype.flush = function() {
...
-  tag += (capabilities.version) ?
-    (' ' + capabilities.version) : '';
-  tag += (capabilities.platform) ?
-    (' ' + capabilities.platform) : '';
-  tag += (' #' + this.task.taskId);
+  tag += (capabilities.logName) ? capabilities.logName
+    : (capabilities.browserName) ? capabilities.browserName : '';
+  tag += (capabilities.version) ? (' ' + capabilities.version) : '';
+  tag += (capabilities.platform) ? (' ' + capabilities.platform) : '';
+  tag += (capabilities.logName && capabilities.count < 2) ? '' : ' #' + this.task.taskId;
```
sjelin: nit: this line is too long

sjelin:
Looks good!  Just some small issues

sjelin:
Merged in at https://github.com/angular/protractor/commit/b76bfc832cfb1123063733b88a7c56fa83d6bf6c

PR #2202, opened 2015-06-03
Adding an example of "executeScript" passing in a regular function

### LINKED ISSUE #2143: Add an example of "executeScript" passing in a regular function
[How can I interact directly with the JavaScript running in my app?](https://github.com/angular/protractor/blob/master/docs/faq.md#how-can-i-interact-directly-with-the-javascript-running-in-my-app) section of the FAQ provides an example of `executeScript()` with a js code passed in as a string:

    browser.executeScript('your script as a string')

But, it is possible to pass in a regular function also, e.g. getting all element's attributes:

```javascript
function getAllAttributes (arguments) {
    var items = {}; 
    for (index = 0; index < arguments[0].attributes.length; ++index) { 
        items[arguments[0].attributes[index].name] = arguments[0].attributes[index].value; 
    }
    return items;
}

browser.executeScript(getAllAttributes, elm).then(function (attrs) {
    console.log(attrs);
});
```

With the absence of multi-line strings in js (as e.g. triple quotes in Python) - this becomes important when it comes to multi-line js code needed to be executed.

It would great if protractor docs would mention it. What do you think?

alecxe:
This closes #2143.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -61,9 +61,21 @@ by.css('.foo').evaluate('bar')
 ```
 would return whatever `{{bar}}` is in the scope of the element with class 'foo'.
 
-You can also execute arbitrary JavaScript in the browser with
+You can also execute arbitrary JavaScript in the browser with:
 ```javascript
-browser.executeScript('your script as a string')
+browser.executeScript('your script as a string');
+```
+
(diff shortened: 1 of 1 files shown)


sjelin:
Merged in at d2543c9187966d36595b815bfd9a5d6f8fcff6f7

PR #2213, opened 2015-06-05
fix(cucumber): fix beforeFeature event handler call guard

jbpros:
Fixes the run failures reported in https://github.com/cucumber/cucumber-js/issues/342.


jbpros:
I signed it!

sjelin:
Merged in at 0262268

PR #2219, opened 2015-06-06
feat (plugins): reporting loaded plugins

### LINKED ISSUE #1961: Protractor to report loaded plugins
For debugging and informational purposes, could protractor report which plugins are currently loaded?

For example:

    [plugins] Plugin "node_modules/protractor/plugins/console/index.js" loaded.
    [plugins] Plugin "node_modules/protractor/plugins/timeline/index.js" loaded.

Cross-referencing: [Output plugins used by protractor](http://stackoverflow.com/questions/29214140/output-plugins-used-by-protractor).

I can send a PR in case you would find it useful. Thanks.

alecxe:
This should close #1961, but let me know if I've made the change in the right place and, whether I needed to attach a test to the change or not. Thanks.

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -44,6 +44,8 @@ var Plugins = function(config) {
       throw new Error('Plugin configuration did not contain a valid path or ' +
           'inline definition.');
     }
+
+    log.debug('Plugin "' + path + '" loaded.');
   });
 };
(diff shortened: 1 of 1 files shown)


sjelin:
Merged in at 7aeebd6

PR #2222, opened 2015-06-08
Update the versions

NicoPennec:


juliemr:
I'd like to upgrade both the selenium standalone and the `selenium-webdriver` node.js dependency together, and unfortunately that is causing some issues so it's not as simple as this. See #2245. Rest assured, updates will happen soon though!

PR #2224, opened 2015-06-08
chore(website): add a message at the top of WebDriver docs explaining…

### LINKED ISSUE #1567: sendKeys example uses webdriver.Key instead of protractor.Key
On the [sendKeys API page](http://angular.github.io/protractor/#/api?view=webdriver.WebElement.prototype.sendKeys), the example reads " element.sendKeys("text was", webdriver.Key.CONTROL, ...."   When I tried something like that, I got an error that webdriver was not defined.  It took me a while before I found that I could use protractor.Key.CONTROL instead.

sjelin:
… the syntax

Closes #1567

[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/2224)

### DIFF
--- a/website/css/protractor.less
+++ b/website/css/protractor.less
@@ -65,6 +65,12 @@
   margin-top: @api-padding-top;
 }
 
+.api-webdriver {
+  border: 1px solid #F0F0F0;
+  padding: 0.5em;
+  font-style: italic;
+}
+
 .api-left-nav {
   .title {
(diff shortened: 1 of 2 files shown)


sjelin:
For reference: 
screenshot from 2015-06-08 12 22 27

juliemr:
LGTM

sjelin:
Closed with 9747966

PR #2234, opened 2015-06-11
feat(webdriver-manager): update download logic with streaming

juliemr:
This seems to fix up some unreliable download issues that we were
previously having.

### DIFF
--- a/bin/webdriver-manager
+++ b/bin/webdriver-manager
@@ -133,6 +133,7 @@ var httpGetFile = function(fileUrl, fileName, outputDir, callback) {
   var filePath = path.join(outputDir, fileName);
   var file = fs.createWriteStream(filePath);
   var ignoreSSL = argv.ignore_ssl;
+  var length = 0;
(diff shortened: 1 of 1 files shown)


bin/webdriver-manager
```diff
@@ -133,6 +133,7 @@ var httpGetFile = function(fileUrl, fileName, outputDir, callback) {
   var filePath = path.join(outputDir, fileName);
   var file = fs.createWriteStream(filePath);
   var ignoreSSL = argv.ignore_ssl;
+  var length = 0;
```
hankduan: nit: maybe name this to contentLength

hankduan:
LGTM

bin/webdriver-manager
juliemr: done.

juliemr:
Thanks. Merged with aded26bc9ee6172d6f64361207f6a8b04da09b0d

PR #2243, opened 2015-06-12
chore(docs): Correctly mark member functions set in the constructor o…

sjelin:
…f an object

Currently if you have a memeber function set like this:
```js
var Class = function() {
  this.fun = function() {...};
}
```
Then dgeni will not grasp that `fun` is a member of the `Class` class.  This fixes that

[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/2243)

### DIFF
--- a/website/docgen/processors/these-children.js
+++ b/website/docgen/processors/these-children.js
@@ -1,0 +1,23 @@
+/**
+ * Set the parents of member functions set in the constructor of an object.
+ */
+module.exports = function theseChildren() {
+  return {
+    $runAfter: ['extracting-tags'],
+    $runBefore: ['tags-extracted'],
+    $process: function(docs) {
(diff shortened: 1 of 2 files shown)


website/docgen/processors/these-children.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+module.exports = function theseChildren() {
+  return {
+    $runAfter: ['extracting-tags'],
+    $runBefore: ['tags-extracted'],
+    $process: function(docs) {
+      var parentDoc = docs[0];
+      for (var i = 0; i < docs.length; i++) {
+        doc = docs[i];
+        if(doc.codeNode && doc.codeNode.expression &&
```
juliemr: nit: space after if

juliemr:
LGTM

PR #2261, opened 2015-06-17
Add support to allow Protractor to test an Angular application which …

gkohen:
…has debugging info disabled

Currently Angular application which for performance reasons, have debug information turn off cannot be tested. 
Per the following issue I've added the Angular debug logging flag to the Protractor run.

[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/2261)

### DIFF
--- a/lib/protractor.js
+++ b/lib/protractor.js
@@ -16,6 +16,7 @@ var ExpectedConditions = require('./expectedConditions.js');
 /* global angular */
 
 var DEFER_LABEL = 'NG_DEFER_BOOTSTRAP!';
+var ENABLE_DEBUG_INFO_LABEL = 'NG_ENABLE_DEBUG_INFO!';
 var DEFAULT_RESET_URL = 'data:text/html,<html></html>';
 var DEFAULT_GET_PAGE_TIMEOUT = 10000;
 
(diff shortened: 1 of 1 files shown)


gkohen:
I signed it!

sjelin:
@juliemr Thoughts on this one?

hankduan:
Is this still relevant? If so, can you point me to when NG_ENABLE_DEBUG_INFO! is added?

gkohen:
@hankduan this is quite relevant, still. If you, meant when is the variable value is added in Protractor, you can find it in line 532. In Angular itself you can find in [this file](https://github.com/angular/angular.js/blob/f13852c179ffd9ec18b7a94df27dec39eb5f19fc/src/Angular.js), that the variable is being evaluated and added or removed to flag if we're in debug mode or not. The Goal is to be able to run Protractor on application that have in their bootstrap, debugInfoEnabled=false.

hankduan:
When does this affect you (i.e. when would the test fail that it wouldn't after this PR)? Protractor already adds debugging information to Angular tests here https://github.com/angular/protractor/blob/master/lib/protractor.js#L488, and it doesn't seem like this PR does anything on top of it.

gkohen:
Let's say I have a code that is production ready. As such in my application
bootstrap I can debugInfoEnabled(false). However, I still want to be able
to run Protractor as part of my continuous integration. I might be wrong,
but we need something to be able to turn on the debug information on in
Angular. What I see in your reference, is that the debug is being turned on
only if $compileProvider.debugInfoEnabled === true. The debug info needs to
be turned on regardless of that flag for tests to work.

justechn:
I really need this for my testing. Is it going to be merged?

sjelin:
@hankduan assigning this to you because I think you have a better grasp of this and Julie and I have a bunch of other PRs

hankduan:
Sorry for the delay. It is merged here: 148f020bf4bbd71e17326581a2f7ed6f4ff55832

PR #2265, opened 2015-06-18
puts interfare with console output

chauhansudhir:
e.g cucumber json format generate json which is getting mixed with "puts" logs

[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/2265)


juliemr:
But that's not the intention of the logs - they should be visible when log level is not set to debug.

PR #2268, opened 2015-06-19
fix(browser.refresh): use timeout in milliseconds

tiago:
When invoked without arguments, browser.refresh uses a 10-millisecond timeout, wrongly documented as seconds. Changed it to delegate timeout handling to browser.get, which defaults to DEFAULT_GET_PAGE_TIMEOUT (10 seconds).

[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/2268)


juliemr:
Thanks! Merged as e60dc0ff4fc4a6d22e877d4182605913fae5d5cb

PR #2284, opened 2015-06-30
docs(faq): add faq about file upload

### LINKED ISSUE #2153: Using file uploads is confusing. Protractor needs documentation/helpers
At the moment, Protractor has no special support for file uploads. Everything is available through the wrapped instance of `selenium-webdriver`, but this is not well documented anywhere and causes lots of confusion, especially with remote file uploads.

Should document/helper/make clearer.

juliemr:
Closes #2153

[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/2284)

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -143,6 +143,17 @@ WebDriver throws errors when commands cannot be completed - e.g. not being able 
 elm.click().then(function() { /* passing case */}, function(err) { /* error handling here */})
 ```
 
+How can I test file uploads?
+----------------------------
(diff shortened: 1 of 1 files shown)


juliemr:
@sjelin can you see if this makes any sense to a reader?

juliemr:
Closed as aadc68e7e9e68fac6922c1f3d1196839374aa1f6

evilaliv3:
@juliemr the current documented solution would work only if users are loading jquery. i all different situations users will get an error such:Failed: $ is not defined

i would suggest to document a solution using native angularjs code.

evilaliv3:
e.g. i would suggest instead of suggesting:

    '$('input[type="file"]') .....

to suggest:

    angular.element(document.querySelector('input[type="file"]')) .....


the latter is more standard, atop of angular and more important not require jquery

PR #2285, opened 2015-06-30
chore(website): fix links in plugins docs

juliemr:
Also, change the `npm run` commands to use gulp - they're
smart enough to grab the right thing now.

[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/2285)


sjelin:
Changing the link URLs like this shouldn't be necessary pending https://github.com/angular/protractor/pull/2288, but we can still do it if you want.  The other part looks fine.

Issue #2169, opened 2015-05-20
The webdriver/protractor split in docs API is confusing


juliemr:
Users don't care that some functions are added by Protractor, and some are wrapped. It's confusing that locators, for example, live in two different locations. We should merge and present this more clearly.

andresdominguez:
We need examples for all the locators. The webdriver jsdocs do not have examples. Probably we need special annotations to extend examples. 

We can have an empty file that contains a bunch of annotations for examples. What about something like this:

```javascript
/**
 * @extendsDoc {webdriver.By.linkText}
 * @replaceDoc
 * [optional] Text that will replace the original jsdoc.
 * @appendDoc
 * [optional] Text that will be appended to the original webdriver jsdoc
 * @view
 * <a>Julie</a>
 * <a>Hank</a>
 *
 * @example
 * expect(element(by.linkText('Julie')).isPresent()).toBe(true);
 */
```

juliemr:
Assigning @sjelin as he's working on it.

PR #2287, opened 2015-06-30
chore(website): API Page improvements

sjelin:
Should close https://github.com/angular/protractor/pull/2169

[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/2287)

### DIFF
--- a/website/css/protractor.less
+++ b/website/css/protractor.less
@@ -84,8 +84,30 @@ a code {
 
   .child {
     color: @child-color;
+  }
+
+  .depth-1 {
     margin-left: 20px;
   }
+
+  .depth-2 {
+    margin-left: 40px;
+  }
+
+  .depth-3 {
+    margin-left: 60px;
+  }
+}
+
+.api-left-nav a, .mobile-menu a {
+  display: inline-block;
+}
+
+.api-left-nav .extension, .mobile-menu .extension {
+  &:before {
+    content: 'Inherited from ';
+  }
+  font-style: italic;
 }
 
 // The grid is too wide.
--- a/website/css/protractor.less
+++ b/website/css/protractor.less
@@ -8,6 +8,10 @@
 @child-color: #777;
 @mobile-menu-margin: 15px;
 
+a code {
+  color: inherit;
+}
+
 .protractor-container {
   margin-bottom: 50px;
   padding-top: @padding-top;
--- a/website/js/api-controller.js
+++ b/website/js/api-controller.js
@@ -98,7 +98,8 @@
       });
 
       self.setViewProperties(list);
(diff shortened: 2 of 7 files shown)


website/js/api-controller.js
```diff
@@ -121,8 +121,9 @@
    * Set the extra properties used by the view to show the docs.
    * @param list
    */
+  var itemsByName = {};
```
juliemr: Make this a property of the controller (`this.itemsByName`)

website/js/api-controller.js
```diff
@@ -230,6 +265,8 @@
           return item.name && item.name.match(nameExpr);
         })
       };
+
+      (itemsByName[parentName] || {}).extension = true;
```
juliemr: maybe
```
if (itemsByName[parentName]) {
  itemsByName[parentName].extension = true
}
```
it's longer but clearer.

website/partials/api.html
```diff
@@ -32,16 +32,17 @@ <h3 id="title">Protractor API <small>{{::version}}</small></h3>
         <span ng-repeat="item in items | filter:searchTerm"
             ng-switch="item.type">
             <div ng-switch-when="title"
-                class="title"
+                class="title {{item.treeClasses}}"
```
juliemr: Optional:
`class="title" ng-class="item.treeClasses"`

I'm not sure which I like better, up to you.

website/css/protractor.less
```diff
@@ -119,6 +138,10 @@
     font-size: 1.1em;
     padding-right: 10px;
   }
+
```
juliemr: Can we just do a non-nested .extension definition once?

juliemr:
Link to the closing issue just links to itself.

sjelin:
All comments addressed

sjelin:
(accidentally closes and re-opened)

juliemr:
Style: I find the indentation somewhat confusing. I think it would be better if the `inherited from` blocks were at the same indentation level as the methods... e.g. (with dots for spaces)

element(locator)
..then
..clone
..<i>Inherited from webdriver.WebElement</i>
..equals
..getDriver
..getId

That makes it clearer to me that `getId` is a method on `element`.

juliemr:
We should do something with the whole `webdriver` section at the bottom (maybe `protractor` and then `inherits from webdrivder`), but I think that should be a separate PR.

sjelin:
Fixed the indentation.  I agree that it's better now.  I donno what you have in mind with the webdriver section at the bottom.  At least it's smaller now :P

juliemr:
LGTM.

sjelin:
Merged in at 9267f45

PR #2288, opened 2015-06-30
chore(website): parse relative links correctly

sjelin:
Before it would add two forward slashes after the hash mark, thus causing the navigation to fail

[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/2288)


sjelin:
Merged in at https://github.com/angular/protractor/commit/4bff24a703d71d88222d38d306b9e95067b798c1

PR #2289, opened 2015-06-30
chore(website): more protractor vs vanilla wd documentation

sjelin:
[Review on Reviewable](https://reviewable.io/reviews/angular/protractor/2289)

### DIFF
--- a/docs/webdriver-vs-protractor.md
+++ b/docs/webdriver-vs-protractor.md
@@ -27,3 +27,26 @@ tests.
 
 If you need the vanilla WebDriver `browser` object, you can access it via
 `browser.driver`
+
+Using ElementFinders
+--------------------
+
+In Protractor, you use the `element` function to find and interact with elements
+through an `ElementFinder` object. This extends a WebDriver `WebElement` by
+adding chaining and utilities for dealing with lists. See
+[locators#actions](/docs/locators) for details.
+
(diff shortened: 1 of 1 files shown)


docs/webdriver-vs-protractor.md
```diff
@@ -27,3 +27,31 @@ tests.
 
 If you need the vanilla WebDriver `browser` object, you can access it via
 `browser.driver`
+
+Chaining WebDriver Actions
```
juliemr: Can we send readers to http://angular.github.io/protractor/#/locators instead, and maybe add a note about chaining actions there?

sjelin: Done.  Let me know if this is what you had in mind.

juliemr: Let's change the text to be a bit more general:

```
Using ElementFinders
------------------------------
In Protractor, use the `element` function to find and interact with elements through an `ElementFinder` object. This extends a WebDriver `WebElement` by adding chaining and utilities for dealing with lists. See
[locators#actions](/docs/locators) for details.
```

sjelin:
Should be good now

juliemr:
Looks like it's missing a push, I don't see the latest changes.

sjelin:
Sorry, I forgot to `-f`

PR #2308, opened 2015-07-07
chore(website): fix old tests and add new ones for API page upgrades

sjelin:


website/test/e2e/api_spec.js
```diff
@@ -133,4 +133,11 @@ describe('Api', function() {
     expect(browser.getCurrentUrl()).
         toMatch(/api\?view=webdriver.WebDriver.prototype.executeAsyncScript/);
   });
+
+  it('should sort the menu to put webdriver docs next to the relavant ' +
```
hankduan: nit: relevant.

Why don't we just assert expect(apiPage.getAdultNmaes()).toBe(['a', 'b', ..., 'browser', 'webdriver.WebDriver', ..., 'y', 'z'])
So we can assert all the orders?

hankduan:
LGTM

website/test/e2e/api_spec.js
sjelin: If the files were arranged differently the location of `browser` and `by` in the docs might be switched, but that shouldn't break the tests.  This test is designed to make sure that `webdriver.WebDriver` comes directly after `browser`, since that's how the inheritance works, but not care about the absolute order of everything.

sjelin:
Merged in at 822749bdc91cc851e4a6348cf88742bfac8438e8

Issue #85, opened 2013-09-12
"Detected a page unload event" error in phantom.js


alexgorbatchev:
Trying to run "onMocha.js" example with phantomjs or chromedriver doesn't seem to work. 

Here's the code, I only changed `webdriver.Capabilities.chrome()` to `webdriver.Capabilities.phantomjs()` and `expect.js` to `chai`

```
/**
 * This example shows how to use the protractor library in a Mocha test.
 * It assumes that a selenium server is running at localhost:4444.
 * Run this test with:
 *   mocha onMocha.js
 */

var util = require('util');
var expect = require('chai').expect;
var webdriver = require('selenium-webdriver');
var protractor = require('../lib/protractor.js');

describe('angularjs.org homepage', function() {
  this.timeout(80000);

  var driver, ptor;

  before(function() {
    driver = new webdriver.Builder().
        withCapabilities(webdriver.Capabilities.phantomjs()).build();

    driver.manage().timeouts().setScriptTimeout(10000);
    ptor = protractor.wrapDriver(driver);
  });

  after(function(done) {
    driver.quit().then(function() {done()});
  })

  it('should greet using binding', function(done) {
    ptor.get('http://www.angularjs.org');

    ptor.findElement(protractor.By.input('yourName')).sendKeys('Julie');

    ptor.findElement(protractor.By.binding('{{yourName}}')).
        getText().then(function(text) {
          expect(text).to.eql('Hello Julie!');
          done();
        });
  });

  ... snip
});
```

Here's the error

```
Uncaught UnknownError: Error Message => 'Detected a page unload event; asynchronous script execution does not work across page loads.'
[...]

PR #2310, opened 2015-07-07
Reset URL cannot be a data url for PhantomJS

carlos-granados:
When trying to use the lates version of Angular with PhantomJS we get a message complaining about "Detected a page unload event". This was fixed in earlier versions of Angular, see issue #85, but reappeared now. The problem is that using data urls to reset the page causes this issue, so we have to do as we do with Safari and use "about:blank" instead


carlos-granados:
I signed it!

2015-07-07 16:19 GMT+02:00 googlebot <notifications@github.com>:

sjelin:
merged in at d15ccdc

PR #2329, opened 2015-07-10
fix for locators.md

justechn:
fixes the link in locators.md to point to the correct resource


justechn:
I signed it

sjelin:
Merged in at 6582f81

PR #2332, opened 2015-07-10
fix(Chrome Accessibility Plugin) No error context

### LINKED ISSUE #2331: StaleElementReferenceError handled poorly in Chrome Accessibility Plugin
If the dom changes as the Chrome Accessibility Plugin generates its results then you can get StaleElementReferenceError without any context to tell you where this error came from.

I have spent the past 4 hours debugging this problem and finally have a working solution that at least gives you more useful information than the silent, context lacking failure that is currently returned.

The failure that I was experiencing happened here: https://github.com/angular/protractor/blob/master/plugins/accessibility/index.js#L176

A better way to handle this would be to pass the error back into the failure reporter.

```
elementPromises.push(
  elem.getOuterHtml().then(function(text) {
    return {
      code: result.rule.code,
      list: trimText(text)
    };
  }, function(reason){
    return {
      code: result.rule.code,
      list: reason
    };
  })
```

JLLeitschuh:
If your tests fail because of StaleElementReferenceError then there is no context about where this is coming from.
By having the failure be handled inside of the plugin then grunt can fail gracefully.
Additionally, this provides context about where the error originated from.

Fixes #2331

### DIFF
--- a/plugins/accessibility/index.js
+++ b/plugins/accessibility/index.js
@@ -178,6 +178,12 @@ function runChromeDevTools(context) {
                 code: result.rule.code,
                 list: trimText(text)
               };
+            },
+            function(reason){
+              return {
+                code: result.rule.code,
+                list: reason
+              };
             })
           );
         });


JLLeitschuh:
I signed it!

sjelin:
Merged in at 4c9886b

PR #2336, opened 2015-07-13
fix(by.exactRepeater): should split by "="

### LINKED ISSUE #2335: Should by.exactRepeater also split by "="?
Following up http://stackoverflow.com/questions/31389521/protractor-by-repeater-not-finding-anything.

There is the following repeater:

    item in filteredItems = (items | filter:'abc')

Currently, it is not possible to locate it by:

    element.all(by.exactRepeater("item in filteredItems"));

We had to include everything before the pipe:

    element.all(by.exactRepeater("item in filteredItems = (items"));

This is because the [`findAllRepeaterRows()`](https://github.com/angular/protractor/blob/e7873093ada02ad6c9d6e190b7184775ab3686ce/lib/clientsidescripts.js#L168) "client side" function splits by "track by", "|" and "as" only:

    return ngRepeat.split(' track by ')[0].split(' as ')[0].split('|')[0].trim() == repeater;

Should it also split by the equal sign `=`?

Thank you.

sjelin:
Closes #2335

### DIFF
--- a/lib/clientsidescripts.js
+++ b/lib/clientsidescripts.js
@@ -107,7 +107,7 @@ functions.findBindings = function(binding, exactMatch, using, rootSelector) {
   function repeaterMatch(ngRepeat, repeater, exact) {
     if (exact) {
       return ngRepeat.split(' track by ')[0].split(' as ')[0].split('|')[0].
-          trim() == repeater;
+          split('=')[0].trim() == repeater;
     } else {
       return ngRepeat.indexOf(repeater) != -1;
     }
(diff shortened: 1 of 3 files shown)


lib/clientsidescripts.js
```diff
@@ -107,7 +107,7 @@ functions.findBindings = function(binding, exactMatch, using, rootSelector) {
   function repeaterMatch(ngRepeat, repeater, exact) {
     if (exact) {
       return ngRepeat.split(' track by ')[0].split(' as ')[0].split('|')[0].
```
hankduan: Can we refactor this into a helper function instead of repeating it 4 times?

hankduan:
Small suggestion + can we add a test to highlight the fix?

lib/clientsidescripts.js
sjelin: It's complicated because these are client side scripts.  Maybe in another CL?  I did work out how to reduce code redundancy in this file a couple weeks ago, but eventually decided it wasn't worth the effort.

hankduan: hmm okay, but can we at least add a test for exactRepeater (and/or a comment stating all the forms that the repeater can take) so they know why these splits are necessary?

LGTM otherwise

sjelin: Working on tests

sjelin:
@hankduan added tests.

sjelin:
Merged in at b147033

PR #2350, opened 2015-07-15
chore(clientsidescripts): reduce code redundancy by allowing for help…

sjelin:
…er functions.

If we decided this was an idea worth using, we could use `wrapWithHelpers` much more and probably reduce the size of the file by a significant margin.


sjelin:
Closed with 19738d3

PR #2356, opened 2015-07-18
Update cucumber.js to run with json formatter

joehua87:
I don't know the meaning inside originalHandleBeforeFeatureEvent.apply(formatter, arguments) but it produce bug when use json formatter.
When I change the code like this, it run normally.


joehua87:
I signed it!

sjelin:
If you use `originalHandleBeforeFeatureEvent.apply(this, arguments)`, does that also solve your bug?

joehua87:
@sjelin :) It's same. I use originalHandleBeforeFeatureEvent(event, callback) to look like the line below

sjelin:
Ok, it sounds like everything will be fixed when we release the next version then?

joehua87:
@sjelin  yes :)

Issue #2359, opened 2015-07-19
ElementArrayFinder.prototype.count is slow since 2.0.0


Droogans:
Since https://github.com/angular/protractor/commit/2a765c76e121de13ff86a279fe3f21cb15c17783 was included, I've seen a test that used to take 2 seconds now take almost 11.

This is the test.

```js 
it('should have 182 rows in the table', function () {
    expect(element.all(by.repeater('r in rows')).count()).to.eventually.equal(182);
});
```

The output from the Selenium server in version 2.1.0.

```
15:54:13.774 INFO - Executing: [is enabled: 0 [org.openqa.selenium.remote.RemoteWebElement@8dadb1bd -> unknown locator]])
15:54:13.782 INFO - Done: [is enabled: 0 [org.openqa.selenium.remote.RemoteWebElement@8dadb1bd -> unknown locator]]
15:54:13.786 INFO - Executing: [is enabled: 1 [org.openqa.selenium.remote.RemoteWebElement@f05f4f94 -> unknown locator]])
15:54:13.854 INFO - Done: [is enabled: 1 [org.openqa.selenium.remote.RemoteWebElement@f05f4f94 -> unknown locator]]
15:54:13.857 INFO - Executing: [is enabled: 2 [org.openqa.selenium.remote.RemoteWebElement@f2f33543 -> unknown locator]])
15:54:13.912 INFO - Done: [is enabled: 2 [org.openqa.selenium.remote.RemoteWebElement@f2f33543 -> unknown locator]]
15:54:13.915 INFO - Executing: [is enabled: 3 [org.openqa.selenium.remote.RemoteWebElement@cfff877 -> unknown locator]])
15:54:13.965 INFO - Done: [is enabled: 3 [org.openqa.selenium.remote.RemoteWebElement@cfff877 -> unknown locator]]
15:54:13.968 INFO - Executing: [is enabled: 4 [org.openqa.selenium.remote.RemoteWebElement@f8780d1a -> unknown locator]])
15:54:14.027 INFO - Done: [is enabled: 4 [org.openqa.selenium.remote.RemoteWebElement@f8780d1a -> unknown locator]]
15:54:14.030 INFO - Executing: [is enabled: 5 [org.openqa.selenium.remote.RemoteWebElement@e87b795b -> unknown locator]])
15:54:14.077 INFO - Done: [is enabled: 5 [org.openqa.selenium.remote.RemoteWebElement@e87b795b -> unknown locator]]
[...]

Issue #1903, opened 2015-03-05
"Reusing" element promises leads to StaleElementReferenceError


tullmann:
I'm trying to test how a page changes in response to user input.  As such, the DOM is mutating, and I believe these mutations make any `element` promise from before the change suspect.  Given that mutating the DOM is the Angular Way, and ExpectedConditions are supposed to help test for an element's visibility/presence, it seems like EC should be hiding this problem?

However, I'm not clear if holding on to such promises across DOM mutations is just "bad form" (and I should always re-"locate" any elements that I think may have changed/moved) or if Protractor should implicitly re-locate elements for me.  (Those both seem problematic, so I think this is an EC issue?)

The Selenium page on the exception is useful/frustrating reading: http://docs.seleniumhq.org/exceptions/stale_element_reference.jsp
It really seems like the case of navigating to a different page should be separated from the case of an element being destroyed and re-created which should also be separated from the case of an element being removed from the DOM.  (So maybe this is a Selenium bug?)

Here is a simplified test case that demonstrates a variant on my problem using the protractor API page.  Note that the test catches exceptions from the ExpectedConditions, but I'm just doing that to count the failures (I don't think every use of ExpectedConditions should do this).

```javascript
describe('demo EC bug', function() {

  var inputbox = $('#searchInput');
  var title = $('.api-title');

  it('force a StaleElementReferenceError', function() {
    browser.get('/protractor/#/api', 90 * 1000);

    expect(title.getText()).toBe('Protractor API Docs');

    // Get an array of element promises, one for each "<li>" on the page (the
    // doc links on the left side).
    var deferred = protractor.promise.defer();
    element.all(by.tagName('li')).then(function(elements) {
      var elProms = [];
      elements.forEach(function(elProm) {
        elProms.push(elProm);
      });
      deferred.fulfill(elProms);
    });

    // Tell controlFlow about the array generation promise
    protractor.promise.controlFlow().await(deferred);

    var failureCt = 0;

    // Once the array is done, put some text into the search box.  This
    // will cause most of the "<li>" elements to disappear, then when the
    // ExpectedConditions operate on the element, a
    // StaleElementReferenceError will get thrown (on some elements).
    deferred.promise.then(function(elements) {
      console.log('ELEMENTS:', elements.length);

      // This should cause all of the <li> elements on the page to
      // disappear: I believe this is effectively invalidating many of the
      // element promises in the "elements" array.
      inputbox.sendKeys('something unique and a bit long');

[...]

PR #2364, opened 2015-07-20
fix(element): ElementArrayFinder.count was slow

hankduan:
The bug fix for #1903 (https://github.com/angular/protractor/commit/2a765c76e121de13ff86a279fe3f21cb15c17783)
was causing ElementArrayFinder.prototype.count to be slow because of
the extranous checks that must be performed. However, the checks could be
delayed into the isPresent check, which will mitigate this issue

fixes(#2359)

### DIFF
--- a/lib/element.js
+++ b/lib/element.js
@@ -350,23 +350,7 @@ ElementArrayFinder.prototype.toElementFinder_ = function() {
  */
 ElementArrayFinder.prototype.count = function() {
   return this.getWebElements().then(function(arr) {
-    var list = arr.map(function(webElem) {
-      // Calling any method forces a staleness check
-      return webElem.isEnabled().then(function() {
-        return 1; // is present
-      }, function(err) {
-        if (err.code == webdriver.error.ErrorCode.STALE_ELEMENT_REFERENCE) {
-          return 0; // not present
-        } else {
-          throw err;
-        }
-      });
-    });
-    return webdriver.promise.all(list).then(function(presenceArr) {
-      return presenceArr.reduce(function(acc, isPresent) {
-        return acc + isPresent;
-      }, 0);
-    });
+    return arr.length;
   }, function(err) {
     if (err.code == webdriver.error.ErrorCode.NO_SUCH_ELEMENT) {
       return 0;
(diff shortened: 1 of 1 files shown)


lib/element.js
```diff
@@ -350,23 +350,7 @@ ElementArrayFinder.prototype.toElementFinder_ = function() {
...
-          throw err;
-        }
-      });
-    });
-    return webdriver.promise.all(list).then(function(presenceArr) {
-      return presenceArr.reduce(function(acc, isPresent) {
-        return acc + isPresent;
-      }, 0);
-    });
+    return arr.length;
```
sjelin: The only issue here is that the this number we return is not necessarily correct anymore.  I feel like this should be documented somewhere

hankduan: I thought about this. I don't think there is anyway for users to build an ElementArrayFinder from raw webelements (it is possible for ElementFinder). So, that means the webelements would never be stale calling from ElementArrayFinder. I could be wrong though. Can you think of an example?

lib/element.js
```diff
@@ -922,8 +906,25 @@ ElementFinder.prototype.$ = function(selector) {
...
+      } else {
+        throw err;
+      }
+    });
+  }, function(err) {
+    if (err.code == webdriver.error.ErrorCode.NO_SUCH_ELEMENT) {
+      return false;
+    } else {
+      throw err;
+    }
```
Droogans: Is this any better than using https://code.google.com/p/selenium/source/browse/javascript/webdriver/webdriver.js#959?

lib/element.js
juliemr: After reviewing the code, I agree with Hank - I don't see a way that `this.getWebElements` doesn't immediately send webdriver find elements calls, so I don't think they can be stale.

lib/element.js
hankduan: webdriver.WebDriver.prototype.isElementPresent doesn't work because it doesn't work with protractor's locator overrides.

lib/element.js
sjelin: You're right.  I hadn't looked at the implementation of `getWebElements`

hankduan:
merged with https://github.com/angular/protractor/commit/3508335199fee5dad74c66d9ee19b8bf448c2e62

PR #2367, opened 2015-07-21
Give some older android platforms a way to proceed

mark-bradshaw:
Some older android platforms don't end up with a resumeBootstrap method despite the defer tag in window.name.  This gives the tests a way to proceed despite that.  I noticed this being an issue while testing at SauceLabs on some android 4.x versions.  It's unclear to me why the resumeBootstrap method never gets applied to angular, but it doesn't.  This change at least allows things to move forward.

### DIFF
--- a/lib/protractor.js
+++ b/lib/protractor.js
@@ -640,7 +640,7 @@ Protractor.prototype.get = function(destination, opt_timeout) {
   }
 
   this.executeScript_(
-      'angular.resumeBootstrap(arguments[0]);',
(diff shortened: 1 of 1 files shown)


sjelin:
I think you're still going to run into problems at https://github.com/angular/protractor/blob/master/lib/clientsidescripts.js#L570.

hankduan:
You should set `ignoreSynchronization` (https://github.com/angular/protractor/pull/2367/files#diff-21a8203aceda03e1b1a0008f789dcf50R567) if there's no resumeBootstrap.

Issue #2365, opened 2015-07-20
  protractor can't find nested elements when using by.repeater and ng-repeat-start


jjulicher:
When you use ng-repeat-start with nested objects protractor isn't exhibiting the expected behaviour. 

I have create a test application to simulate this scenario with the following code:

Controller: 
```javascript
   .controller('View1Ctrl', ['$scope',function ($scope) {
        $scope.tests = [{
                name: 'test1',
                nestedElements: [{name: 'test1.subtest1'}, {name: 'test1.subtest2'}]
            },
            {
                name: 'test2',
                nestedElements: [{name: 'test2.subtest1'}, {name: 'test2.subtest2'}, {name: 'test2.subtest3'}]
            }
        ];
    }]);
```

View:
```html

<table>
<tr ng-repeat-start="t in tests">
    <td>{{t.name}}</td>
</tr>
<tr ng-repeat-start="t2 in t.nestedElements">
    <td>{{t2.name}}</td>
</tr>
<tr ng-if="false" ng-repeat-end></tr>
<tr ng-if="false" ng-repeat-end></tr>
</table>

```
Protractor Test:
```javascript
  it('should get all the tests', function(){
        var tests = element.all(by.repeater('t in tests'));
        tests.then(function(e){
            console.log('found ' + e.length + ' tests');

            // expect(e.length).toBe(2);
            for (var i = 0; i < e.length; i++) {
                e[i].getInnerHtml().then(function(r){console.log(r);});
                var subtasks =  e[i].all(by.repeater('t2 in t.nestedElements'));
                subtasks.getInnerHtml().then(function(r){console.log(r);});
                expect(subtasks).toBeDefined();
            }
        });
    });
```

I would expect that it would only return the 2 test objects not those plus the other nested ng-repeat objects. 

Here is a full repo with all the code and tests: https://github.com/jjulicher/NestedRepeaterTest

@sjelin  Please don't close this if you don't understand nested repeat elements.

As per the docs:

Special repeat start and end points
To repeat a series of elements instead of just one parent element, ngRepeat (as well as other ng directives) supports extending the range of the repeater by defining explicit start and end points by using ng-repeat-start and ng-repeat-end respectively. The ng-repeat-start directive works the same as ng-repeat, but will repeat all the HTML code (including the tag it's defined on) up to and including the ending HTML tag where ng-repeat-end is placed.

sjelin:
Gotta love being lectured at :)

[...]

PR #2368, opened 2015-07-22
fix(ng-repeat): properly detect the end of an ng-repeat-start block

sjelin:
This bug was introduced when by.exactRepeater was

Discovered while investigating issue #2365


hankduan:
Can you describe what this is for?

sjelin:
Uh, a bug!  Basically, way back in 324f69d6aa7c23ad77f1d50e26e0a56bade40132, I replaced all the `attr.indexOf(repeater) != -1` checks with `repeaterMatch(attr, repeater, exact)`, which is stricter if `exact` is on.  Some some of those `indexOf(..) != -1` checks shouldn't have been changed.  This essentially reverts my change in the places where is it inappropriate.

PR #2379, opened 2015-07-27
Switch protractor to minijasminenode2

ns11:


ns11:
I signed it!

hankduan:
Why do you need this? We're using "jasmine": "2.3.1" to support jasmine2, and 'minijasminenode2' is essentially obsolete. See http://www.protractortest.org/#/jasmine-upgrade.

Closing for now. Please open an issue if there is any lingering problems.

Issue #2245, opened 2015-06-13
Update to WebDriver 2.47.1


juliemr:
This will fix several issues, but is currently blocked because `selenium-webdriver` node module @ 2.46.0 is causing issues with `ElementArrayFinder.each`, probably due to control flow changes.

elifa:
I would appreciate if you could add the proxy support to Protractor that comes with WebDriver 2.46 as well.
I suggest adding this row to the `driverProvider.js` file:

```javascript
DriverProvider.prototype.getNewDriver = function() {
    // ...
    usingWebDriverProxy(this.config_.seleniumProxy).
    // ...
}
```

juliemr:
See also https://github.com/angular/protractor/pull/2263

hankduan:
Here's a test which hangs with webdriver 46, but not 45.

```javascript
var webdriver = require('selenium-webdriver');
var driver = new webdriver.Builder().
  usingServer('http://localhost:4444/wd/hub').
  withCapabilities(webdriver.Capabilities.chrome()).
  build();
var flow = webdriver.promise.controlFlow();

driver.get('http://www.w3schools.com/html/html_lists.asp');

var printTexts = function(webElems) {
  var attrVals = webElems.then(function(arr) {
    var results = arr.map(function(webElem) {
      return webElem.getText();
    });
    return webdriver.promise.all(results);
  });

  flow.execute(function() {
    return attrVals.then(function(res) {
      console.log(res[0]);
    });
  });

  return true;
};

driver.findElements(webdriver.By.css('.example_result li')).then(function(arr) {
  var result = arr.map(function(webElem) {
    var webElemsPromise = webdriver.promise.fulfilled(webElem).then(function(webElem) {
      return [webElem];
    });
    return webdriver.promise.fullyResolved(printTexts(webElemsPromise));
  });
  return webdriver.promise.all(result);
})

driver.quit();
```

hankduan:
fyi, this is pending https://github.com/SeleniumHQ/selenium/issues/796

juliemr:
NOTE - the pending issue is still pending, but with WebDriverJS 2.47 out and other issues beginning to crop up because of the stale version of WebDriverJS (such as install warnings with node.js 4.1 and errors with new browsers when using `directConnect`) we are going to go ahead and do the upgrade anyway. We will attempt to outline potential new issues clearly in the breaking changes section of the release notes. Stay tuned!

juliemr:
Update:

[...]

PR #2426, opened 2015-08-13
feat(webdriver): bump selenium, chromedriver, and iedriver versions

gempesaw:
Hullo, I noticed that the versions that `webdriver-manager` references are slightly outdated. I've seen that there were a few issues/PRs with webdriver @ 2.46.0, so if there are external concerns that mean y'all would like to keep these versions in `webdriver-manager` at those specific ones, feel free to reject.

Cheers & thanks for protractor!

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,7 +1,7 @@
 {
   "webdriverVersions": {
-    "selenium": "2.45.0",
-    "chromedriver": "2.15",
-    "iedriver": "2.45.0"
+    "selenium": "2.47.1",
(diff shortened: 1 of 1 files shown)


hankduan:
Thanks for the contribution. In fact, there is a blocker on this (https://github.com/angular/protractor/issues/2245). I'm going to close this PR for now, but will upgrade selenium once the issue is resolved.

gempesaw:
ah yep, figured I'd bring it up in case it happened to have been missed. thanks!

PR #2429, opened 2015-08-17
chore: fix jshint errors blocking npm test

marcysutton:
To get `npm test` to run again, some cleanup was necessary. Note this includes a change to the jshint config allowing `eval`, rather than rewriting the `wrapWithHelpers` function in `lib/clientsidescripts.js`.

Refs #2398

PR #2155, opened 2015-05-18
Fix linting error in client side scripts

NickTomlin:
This is extremely minor, but it was a blocker for me trying to follow the contributing instructions. It might be nice to include linting as part of the travis build. I can PR that if that sounds acceptable.

Before: 

screen shot 2015-05-18 at 5 12 37 pm

After: 

screen shot 2015-05-18 at 5 16 07 pm

PR #2435, opened 2015-08-17
chore(travis): Lint project files in CI

NickTomlin:
This lints test in CI, which will hopefully prevent issues like #2429 and #2155  from creeping into the codebase. At the moment i'm only aware of the need for jshint, but let me know if there are additional scripts that need to be run.

**Note**: this PR is going to fail in CI until the fixes in #2429 are merged in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ matrix:
 
 
 before_script:
+  - npm run pretest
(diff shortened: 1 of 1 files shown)


hankduan:
Thanks! Merged at 94859efc0c89174d20ad9c06283146f1f7e1bd63

PR #2445, opened 2015-08-20
Spec with line number

### LINKED ISSUE #2413: Unable to run single scenario by passing line number and spec (or feature) file name
In cucumber framework, a particular scenario can be executed by passing the line number and feature file name. like this
>cucumber features/smoke.feature:4

So, only the scenario at line number 4 gets executed. When cucumber framework is used with protractor it throws error saying no file pattern matched since the line number was not stripped out while looking for the existence of the spec file.
Error:
WARNING - pattern test/e2e/features/smoke.feature:4 did not match any files.

karthiktv006:
Resolve spec with line number so, cucumber framework can run scenarios based on line number
example:
    specs: [
        'cucumber/lib.feature:7'
      ]

Fixes #2413


jlin412:
+1

karthiktv006:
I signed it!

lib/configParser.js
```diff
@@ -96,12 +96,18 @@ ConfigParser.resolveFilePatterns =
 
   if (patterns) {
     for (var i = 0; i < patterns.length; ++i) {
-      var matches = glob.sync(patterns[i], {cwd: cwd});
+      var fileName = patterns[i].split(':')[0];
+      var lineNumber = patterns[i].split(':')[1];
```
hankduan: Leave a message referencing your issue (https://github.com/angular/protractor/issues/2413)

lib/configParser.js
```diff
@@ -96,12 +96,18 @@ ConfigParser.resolveFilePatterns =
...
+      var matches = glob.sync(fileName, {cwd: cwd});
       if (!matches.length && !opt_omitWarnings) {
         log.warn('pattern ' + patterns[i] + ' did not match any files.');
       }
       for (var j = 0; j < matches.length; ++j) {
-        resolvedFiles.push(path.resolve(cwd, matches[j]));
+        if(lineNumber) {
```
hankduan: nit: clear the redundancy a little, and it makes it more obvious that lineNumber is only for cucumber. 

var resolvedPath = path.resolve(cwd, matches[j]);
if (lineNumber) {
  resolvedPath += ':' + lineNumber;
}
resolvedFiles.push(resolvedPath);

hankduan:
minor comments. Looks good otherwise

karthiktv006:
@hankduan - its updated based on your comments. Let me know if any other changes needed.

hankduan:
Thanks! Merged via https://github.com/angular/protractor/commit/ff88e969d55585cc4267d75c12c0cafc78a01895

darrinholst:
This got yanked out with [3.0](https://github.com/angular/protractor/commit/ddb8584a59343284904676ef6d8db5c1c996b900) so we can't send line number for cucumber scenarios anymore. I didn't see a way to handle this in cucumber-framework-protractor. Can we get this added back in if I send a PR or figure out a way to handle this in the framework lib?

juliemr:
Hm, the idea was supposed to be that cucumber-framework-protractor would take care of it. What does it need in order to handle it there?

darrinholst:
Specifically it needs the line number passed in the list of specs....`[features/foo.feature:13, features/bar.feature:27]`. If we want to keep cucumber specific stuff out of the core then we would need some way to call into a `resolveFilePatterns` provided by the framework. I haven't been that far into the internals of protractor to know how easy/hard that is.

Another way to do it would be just to pass non-glob patterns through verbatim. Untested and not sure if not resolving the path would have any downstream ramifications...

```javascript
  if (patterns) {
    for (var i = 0; i < patterns.length; ++i) {
      var fileName = patterns[i];

      if (glob.hasMagic(fileName)) {
        var matches = glob.sync(fileName, {cwd: cwd});

        if (!matches.length && !opt_omitWarnings) {
          log.warn('pattern ' + patterns[i] + ' did not match any files.');
        }
        for (var j = 0; j < matches.length; ++j) {
          var resolvedPath = path.resolve(cwd, matches[j]);
          resolvedFiles.push(resolvedPath);
        }
      } else {
        resolvedFiles.push(filename);
      }
    }
  }
```
Or yet another option would to be just ignore the `specs` that protractor tells us to run all together and handle it in the framework with our own `features` config option. (I think. Again, just another untested idea.)

darrinholst:
I guess `path.resolve` doesn't care if the file actually exists so going back to the second option this _should_ work...

```javascript
  if (patterns) {
    for (var i = 0; i < patterns.length; ++i) {
      var fileName = patterns[i];
      var matches = glob.hasMagic(filename) ? glob.sync(fileName, {cwd: cwd}) : [fileName];

      if (!matches.length && !opt_omitWarnings) {
        log.warn('pattern ' + patterns[i] + ' did not match any files.');
      }
      for (var j = 0; j < matches.length; ++j) {
        var resolvedPath = path.resolve(cwd, matches[j]);
        resolvedFiles.push(resolvedPath);
      }
    }
  }
```

which would only change [line 98](https://github.com/angular/protractor/blob/master/lib/configParser.js#L98) to only do globing if it's a glob pattern

Galileo1:
@cnishina seems like version 4.0.11 configParser doesn't have this feature- 
image


I was trying to run single scenario from a feature using 'e2e-tests\\features\\**\\Feature Editor.feature:5' expression but i doesn't work config parser throws error and runner is not able to pick this up because spec =[] is empty.

Can you please advise?

Galileo1:
for those who have been seeking the answer for this passing cucumberOpts.name=['name of scenario'] will solve this issue.

PR #2437, opened 2015-08-18
Site - Mocha/cucumber spec links don't work

nwolverson:
Not sure if the intention was that they existed in `gh-pages` branch but currently these `../spec` links are broken.

PR #2455, opened 2015-08-27
Mocha/cucumber spec links don't work

nwolverson:
Fix as per #2437 - but actually in the source this time...


hankduan:
Does it work without the 'https://github.com/angular/protractor/blob/master'? The website is hosted in protractortest.org and not github. @sjelin can you take a look? I'm not familiar with website generation.

nwolverson:
The file in question is not hosted with the website, neither are the other `.feature` or `.js` links due to a rewrite in the gulp file - could alternatively add a rule for `.md here but didn't want to break anything - https://github.com/nwolverson/protractor/blob/master/website/gulpfile.js#L111

sjelin:
This is a real bug but the solution isn't great.  I'll fix this tomorrow.

PR #2457, opened 2015-08-27
resolving the location of AUDIT_FILE

mstrutt:
In my case it installed to the root `node_modules` folder instead of the `protractor/node_modules` folder. Using `require.resolve` should handle both eventualities.

Not sure how to add a test for this, but it passed the existing one


juliemr:
Thanks! Merged as 2ff7a0771b6695dc49566ed81548b3fe2cebf11c

PR #2462, opened 2015-08-31
docs(plugins): Add a section for community plugins

### LINKED ISSUE #2246: Decoupling runtime frameworks (angularJs) and e2e testing framewoks (protractor)
## TL;DR 
This issue proposes to build an spec to enable decoupling between testing frameworks and the runtime frameworks, libraries and applications.

## Intro
I've been working for more than a year with protractor and angular and I love both tools, but I've come to some issues when trying to make protractor work when building some modules and applications for angular (specially when playing with the router and when doing lazyloading).

In this time I've learned that in order to make protractor work (wait natively for everything to be ready in the page without aditional waits for this and for that in the test spec) you should use angular API "correctly" in your development. The best pattern to do so I've come with is using $browser.defer(...).cancel(...) (please if there's a better one please let me know).

Any way I think it's weird that protractor needs angular so badly to work. I understand that it provides some added value like model or directive based selectors, but IMHO the "waitForAngular" service should be something more abstract like "waitForBrowserToBeReady".

## Proposal

I think it will be better for the OpenSource community if there was an abstraction layer (hopefully an spec) between the runtime libraries like angular and the testing frameworks like protractor. This will have the following advantages:
- Testing frameworks will not be coupled to specific framework or versions.
- Runtime frameworks will be testable by any testing framework, present or future.
- Last but not least, applications and libraries will have a mechanism to create aditional testable asynchronous services outside of the frameworks (like websockets, webworkers, lazyloading, custom rendering algorithms, etc).

## Sample implementation
I've made a first reference implementation for this future spec for you to check the viability of the proposal: https://github.com/alfonso-presa/testability.js

alfonso-presa:
Now that the plugins API is much more powerful, I though it might be worth including a section in the docs for community plugins for reference. I've included mine :-).

- Includes in the list the protractor-testability-plugin

Closes #2246

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -17,6 +17,7 @@ The `plugins` folder contains default plugins for Protractor.
   * [ngHint Plugin](/docs/plugins.md#nghint-plugin)
   * [Timeline Plugin](/docs/plugins.md#timeline-plugin)
   * [Console Plugin](/docs/plugins.md#console-plugin-chrome-only)
+* [Community Plugins](/docs/plugins.md#community-plugins)
 
 Using Plugins
 -------------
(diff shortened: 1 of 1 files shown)


sjelin:
Closed with 056986d

PR #2470, opened 2015-09-03
Feat(accessibility plugin): added option "showUnsupportedRulesWarning…

faceleg:
…" to plugins.chromeA11YDevTools

Defaults to "true", which is the vanilla behaviour.

This change is desirable because without it, there is no way to use both the A11Y plugin **and** the console plugin with errors turned on - the A11Y plugin outputs errors (see below) that cause a test run to fail. 

When the "showUnsupportedRulesWarning" is set to "false", the following console output will not be emitted:

```TEXT
 	 Fail:  Console output
		WARNING: console-api 1753:141 Some rules cannot be checked using the axs.Audit.run() method call. Use the Chrome plugin to check these rules: AX_FOCUS_02
		WARNING: console-api 1753:286 To remove this message, pass an AuditConfiguration object to axs.Audit.run() and set configuration.showUnsupportedRulesWarning = false.
```

### DIFF
--- a/plugins/accessibility/index.js
+++ b/plugins/accessibility/index.js
@@ -148,7 +148,14 @@ function runTenonIO(context) {
 function runChromeDevTools(context) {
 
   var data = fs.readFileSync(AUDIT_FILE, 'utf-8');
-  data = data + ' return axs.Audit.run();';
+  data += 'var configuration = new axs.AuditConfiguration();';
+
+  if (typeof context.config.chromeA11YDevTools.showUnsupportedRulesWarning !== 'undefined' &&
+      !context.config.chromeA11YDevTools.showUnsupportedRulesWarning) {
+    data += 'configuration.showUnsupportedRulesWarning = false;';
+  }
+
+  data = data + ' return axs.Audit.run(configuration);';
 
   var elementPromises = [],
       elementStringLength = 200;


faceleg:
@googlebot signed it!

faceleg:
Maintainers: I would like to add tests for this, but 

a) it was midnight
b) my cursory glances through the code didn't reveal any existing tests for features that manipulate the console output of a plugin

Please let me know what else is required on my part to get this in.

faceleg:
Bumping

sjelin:
Hey, sorry, I'm going back through our PR backlog.  We're splitting plugins into separate repos, so could you move this PR to https://github.com/angular/protractor-accessibility-plugin?

faceleg:
I'll put it on the list! Thanks for getting back to me on this.

faceleg:
@sjelin would the new PR need tests? See my comment https://github.com/angular/protractor/pull/2470#issuecomment-137566522

sjelin:
The PR should have at least one test.  You could mock out `console.log` maybe?  Or you could pipe the output into a separate program which validates it.  Something like adding

```js
  'node node_modules/protractor/lib/cli.js spec/unsupported.js | spec/unsupportedValidator.sh',
```

to [this array](https://github.com/angular/protractor-accessibility-plugin/blob/master/test.js#L6)

faceleg:
Good tips, thanks

PR #2483, opened 2015-09-09
updated saucelabs dependency to 1.0.1 to support proxy

theandrewlane:


juliemr:
Thanks! Merged as 96def81dc7d364e789fc290e97aee0f898648a10

Note that your PR contained an unrelated commit changing the selenium version, which has not been merged.

PR #2508, opened 2015-09-19
feat(webdriver-manager): Configurable output directory

blabno:
Allows to change binaries cache directory, so that you don't need to download them each time you remove node_modules and run npm install and webdriver-manager update.

### DIFF
--- a/bin/webdriver-manager
+++ b/bin/webdriver-manager
@@ -100,7 +100,7 @@ var cli = optimist.
         '  update: install or update selected binaries\n' +
         '  start: start up the selenium server\n' +
         '  status: list the current available drivers').
-    describe('out_dir', 'Location to output/expect binaries, can be modified with WEBDRIVER_MANAGER_SELENIUM_DIR evn variable').
+    describe('out_dir', 'Location to output/expect binaries, can be modified with WEBDRIVER_MANAGER_SELENIUM_DIR env variable').
     default('out_dir', SELENIUM_DIR).
     describe('seleniumPort', 'Optional port for the selenium standalone server').
     describe('ignore_ssl', 'Ignore SSL certificates').boolean('ignore_ssl').
--- a/bin/webdriver-manager
+++ b/bin/webdriver-manager
@@ -13,7 +13,7 @@ var childProcess = require('child_process');
 /**
  * Download the requested binaries to node_modules/protractor/selenium/
  */
-var SELENIUM_DIR = path.resolve(__dirname, '../selenium');
+var SELENIUM_DIR = process.env.WEBDRIVER_MANAGER_SELENIUM_DIR || path.resolve(__dirname, '../selenium');
 
 var versions = require('../config.json').webdriverVersions;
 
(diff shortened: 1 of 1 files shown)


blabno:
I signed it!

sjelin:
Why an env variable instead of a flag?

blabno:
For the same reason as http_proxy, so that we can setup env and keep build
descriptors env agnostic.
23 wrz 2015 19:53 "Sammy Jelin" <notifications@github.com> napisał(a):

blabno:
Particularly I have platform where I run tasks, mini projects, completed by
other people remotely on their PCs. Most of the dependencies are the same
so I want my Jenkins to reuse dependencies between build to save bandwidth.

If I was to put path to out-dir in package.jso n then it would be very
tightly coupled to my env, and may conflict with other people config.

Besides there's already a flag. This PR adds additional flexibility.
23 wrz 2015 20:00 "Bernard Łabno" <s4237@pjwstk.edu.pl> napisał(a):

bin/webdriver-manager
```diff
@@ -100,7 +100,7 @@ var cli = optimist.
         '  update: install or update selected binaries\n' +
         '  start: start up the selenium server\n' +
         '  status: list the current available drivers').
-    describe('out_dir', 'Location to output/expect ').
+    describe('out_dir', 'Location to output/expect binaries, can be modified with WEBDRIVER_MANAGER_SELENIUM_DIR evn variable').
```
juliemr: typo: evn/env

juliemr:
My gut reaction is that there must be a better way to do this - why do you need to change your `package.json` to couple a command line flag?

The reason we support HTTP_PROXY as an environment variable is because this is an existing convention that users expect to work for all node projects on their machine. I'm wary of introducing new environment variables.

blabno:
Ok, lets imagine that my package.json looks like this:
```
"scripts": {
  "pretest":"webdriver-manager update --standalone --out_dir=/home/bernard/.protractor-cache"
}
```

Now Mikey, won't be too happy because he would like to store the binaries in `/home/mikey`.
Env variable solves this nicely.

blabno:
btw. the reason why I put `webdriver-manager update` into package.json is because I want to limit number of commands my students have to type before running tests to minimum.

Of course I could write for each task that I give them:
install npm
install grunt
install bower
install protractor
update webdriver
etc.

But isn't this nicer `npm install && npm test`?

juliemr:
But, you still have to tell them to set the environment variable! 

I think the better way to solve this is:

```
  "config": {
    "webdriver_manager_selenium_dir": "/abc/def/ghi"
  },
  "scripts": {
    "pretest":"webdriver-manager update --standalone --out_dir=$npm_package_config_webdriver_manager_selenium_dir"
  }
```

juliemr:
I'm closing this as I still don't think we should be adding new, non-standard environment variables, and I'm convinced that npm already provides the tools to do this. Thanks for your work (and for using Protractor in your teaching!).

blabno:
Pity. I'll have to force users to use my fork.
10 lis 2015 22:31 "Julie Ralph" <notifications@github.com> napisał(a):

PR #2537, opened 2015-09-24
refactor(Array#forEach): Replaced Array#forEach with more suitable

stalniy:
Replaced `Array#forEach` with more suitable `Array` methods


sjelin:
Merged with 1fa8f5b

PR #2541, opened 2015-09-25
chore(website): fix parsing in function list

### LINKED ISSUE #2486: Documentation display errors
screen shot 2015-09-10 at 8 40 41 pm

For some reason there seem to be a lot of dead tags floating around in the docs. Almost looks like it's not rendering fully.

Google Chrome Version 45.0.2454.85 (64-bit) on Mac OSX

sjelin:
fixes #2486 and #2672

### DIFF
--- a/website/js/shared.js
+++ b/website/js/shared.js
@@ -1,0 +1,25 @@
+/*
+ * Use the $sce service to trust the html rendered in the view.
+ * Also parse links
+ *
+ * @param $sce The $sce service from Angular
+ * @param {String} html The HTML to trust
+ * @return {*} An object that can be passed to $sce.getTrustedHtml(value) to
+ *   obtain the original value
+ */
+function trustHTML($sce, html) {
+  if (!html) {
+    return;
+  }
+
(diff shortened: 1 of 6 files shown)


juliemr:
Don't we have to include `ngSanitize` in the app to use `ng-bind-html`?

sjelin:
Oh, probably.  I was coping code from another part of the website but I guess that was foolish

sjelin:
Fixed

sjelin:
Now also fixes #2672!  Also, we should talk about XSS

website/js/shared.js
```diff
@@ -0,0 +1,23 @@
+/*
+ * Use the $sce service to trust the html rendered in the view.
+ * Also parse links
+ *
+ * @param $sce
+ * @param html
```
juliemr: add `@return` doc

juliemr:
Can you add a test for `trust` to https://github.com/angular/protractor/blob/master/website/test/unit/api-controller-spec.js?

sjelin:
All comments addressed

PR #2510, opened 2015-09-21
updated deps

donaldpipowitch:
Please update `selenium-webdriver` to `2.47.0` which has updated `ws` to `0.8.0` and release a new protractor version after that. This should fix https://github.com/angular/protractor/issues/2489 and the problems we see with npm3 and Node4.

PR #2553, opened 2015-09-29
Update selenium-webdriver.

panuhorsmalahti:
This updates selenium-webdriver, which will use a newer version of ws, which will use a newer version of bufferutil finally fixing node 4.1.0 compatibility issue. The older bufferutil didn't support newest nan package, which caused a build failure for me on node 4.1.0.

I verified that this change fixed my issue.


panuhorsmalahti:
Never mind, identical to #2510.

PR #2559, opened 2015-09-29
chore(dependencies): update selenium-webdriver to 2.47.1

juliemr:
Along with it, update `jasminewd2` to avoid situations where the
control flow gets locked up and hangs.

Potential Breaking Change:

This is passing all existing Protractor tests, but there is a
possibility that the changes to the control flow will cause some
test flows to hang. If this causes issues, please revisit your
use of functions affecting the control flow, such as
`flow.execute`.


benkroeger:
+1

juliemr:
I'm seeing a consistent issue with the polling spec, which only occurs on Travis. Cannot yet reproduce locally or running from my machine to sauce.

PR #2563, opened 2015-09-30
Enable dry-run option of cucumber

karthiktv006:


sjelin:
Merged in with 93abf53

PR #2564, opened 2015-10-01
fix(config): Fixes absolute path parsing in windows

NickTomlin:
This allows absolute paths absolute paths in to be properly parsed in windows.
This should maintain the line-number feature introduced in [ff88e](https://github.com/angular/protractor/blob/ff88e969d55585cc4267d75c12c0cafc78a01895/lib/configParser.js#L101) without breakage.

I decided to use `parse` and `format` from [`path`](https://nodejs.org/api/path.html#path_path) instead of a just a regex because it feels a little safer to me, but please let me know if there's a better/cooler way to do this. 

Caveat: I do not use cucumber, or windows, so I'd appreciate a second pass from someone that does use those tools :)

### DIFF
--- a/lib/configParser.js
+++ b/lib/configParser.js
@@ -98,17 +98,23 @@ ConfigParser.resolveFilePatterns =
     for (var i = 0; i < patterns.length; ++i) {
       // Cucumber allows running a spec given a line number. See
       // https://github.com/angular/protractor/issues/2413
-      var fileName = patterns[i].split(':')[0],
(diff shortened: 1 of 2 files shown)


juliemr:
This looks good to me, but I haven't had a chance to check with a windows machine yet. Let me go open up the windows laptop...

juliemr:
Thanks! This is merged as d6aebbad6e9b191fef141472887637ee4318438e

PR #2575, opened 2015-10-02
refactor(element.prototype): Moved webdriver element specific methods…

### LINKED ISSUE #2574: Add ElementFinder's methods to prototype
Instead of adding methods to each instance you can add them only once to prototype. That will reduce memory usage and will speed up ElementFinder creation. I'm talking about this part:
```js
  var self = this;
  WEB_ELEMENT_FUNCTIONS.forEach(function(fnName) {
    self[fnName] = function() {
      var args = arguments;
      var actionFn = function(webElem) {
        return webElem[fnName].apply(webElem, args);
      };
      return self.applyAction_(actionFn);
    };
  });
```
Similar code inside `ElementFinder` class

stalniy:
… into prototype. Fixes #2574

### DIFF
--- a/lib/element.js
+++ b/lib/element.js
@@ -65,20 +65,19 @@ var ElementArrayFinder = function(ptor, getWebElements, locator, opt_actionResul
   this.getWebElements = getWebElements || null;
   this.actionResults_ = opt_actionResults;
   this.locator_ = locator;
-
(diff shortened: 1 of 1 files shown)


sjelin:
Merged in with deba67d

PR #2578, opened 2015-10-03
aligned synchronization with controlFlow

martin770:


martin770:
Since I'm just a QA guy, and not a real javascript developer, feel free to modify these changes as needed to follow best practices.

In the project I am currently working on, we have a non-angular login page, an angular app, and then within the angular app we have some legacy code that is not angular. When I overwrite the .get and .waitForAngular methods with these, they work awesomely! I am able to use browser.get and browser.element functions instead of browser.driver.get and browser.driver.findElement, even on non-angular pages/elements, so it helps to make my pageObject code cleaner and more uniform.

In practice, it looks something like this:

    var pageObject = function(elementFinder) {
        return browser.getSynchronization().then(function(sync) {
            if (sync !== 'protractor') {
                return browser.wait(function() {
                    return elementFinder.isDisplayed();
                });
            }
            return elementFinder;
        });
    }

    onPrepare: {
      // nothing needed here
    }

    beforeEach: {
        browser.setSynchronization(false);
        browser.get(homePageUrl);
        element(login_button).isDisplayed().then(function(isLoginPage) {
            if (isLoginPage) {
                pageObject(element(login_button)).click(); // sync is false, does not waitForAngular
            }
            browser.setSynchronization(true);
        }
    }
    it('should use the legacy non-angular thing', function() {
        pageObject(element(launch_legacy_dialog_button)).click(); // sync is true, waits for Angular
        expect(element(legacy_app).isDisplayed()).toBe(true);
        browser.setSynchronization(false);
        pageObject(element(legacy_dialog_exit_button)).click(); // sync is false, waits to be displayed
        // because waitForAngular doesn't know this dialog is closing, we have to manually wait
        browser.wait(function() {
            return element(legacy_app).isDisplayed().then(function(isDisplayed) {
                return !isDisplayed;
            });
        });
        browser.setSynchronization(true);
        pageObject(element(sign_out_button)).click(); // sync is true, waits for Angular
        browser.setSynchronization(false); // turn off sync because we go to the non-angular page
        expect(element(login_button).isDisplayed()).toBe(true);
    }


So my actual implementation might be a little more elegant, but essentially wrapping the elementFinder in a pageObject that knows to either 'waitForAngular' or 'waitReady' (based on the synchronization within the context of the controlFlow) before interacting with the element seems to bridge the gap between the angular and non-angular stuff quite nicely.

sjelin:
I'm not sure about this exact implementation, but getting `ignoreSynchronization` to align with the control flow is probably a good idea.  Though it would be a huge breaking change.

juliemr:
I think the way that we will deal this will be when we change to using a proxy for determining stability. That will be done in a later breaking change. In the meantime, we've improved the situation slightly with https://github.com/angular/protractor/commit/f034e010156a85cf1826b95eb7f41f50ef5a1791

I'm going to close this since it's not the direction we're going to go, but thanks for your PR!

Issue #2460, opened 2015-08-31
Angular dependency to be removed


opichals:
As Protractor is not meant to be used exclusively with Angular it would be nice to avoid any 'Angular' dependency and rather let angular implement a Protractor API to become properly Protractor testable and utilize the waitStable testability feature.

We have been successfully using Protractor to test a ReactJS application which implements the necessary stuff from the window.angular for Protractor to work. The following code is the one that works and simulates the [whenStable()](https://github.com/angular/protractor/blob/4a97370608694343cefdfcfd3f93ca4526d44161/lib/clientsidescripts.js#L49-L69) functionality:
```
window.angular = {
    // dummy for Protractor boot
    resumeBootstrap: function() {},

    // dummy for Protractor boot
    module: function() {
        var module = {
            config: function() {
                return module;
            }
        }
        return module;
    },

    // hook for `github.com/angular/protractor/lib/clientsidescripts.js: waitForAngular`
    getTestability: function() {
        return {
            whenStable: function(fn) {
                ...
            }
        }
    }
};
```

The Protractor specific API could look something like (not completely accurate but should be enough to get what I mean):
```
window.protractor = {
    // ...
    resumeBootstrap: function() {},

    // Simplified config to avoid any proractor -> angular dependency
    config: function(opts) {
    },

    getTestability: function() {
        return {
            whenStable: function(fn) {
                ...
            }
        }
    }
}

**Comment by stalniy** [comment]
Probably better idea is to implement an ability to create Framework/App specific waiter.
```js
var react = {
   connect: function() {
      // here is dance with angular.resumeBootstrap and similar stuff
   },

   wait: ..., // similar to waitForAngular implementation here
 
   addMockModule: ....

   removeMockModule: ....
};
```
PROS:
* no extra global variable in app scope
* no need to change app
* clear interface
* can be created for any framework/app on Nodejs side

And by default use waiter for `angular.js`, first one and last 2 methods are optional

juliemr:
The way that we'd like to implement this is with a separate process that proxies commands to webdriver. Then, that process will take care of all waiting commands, and will have the option to swap them out. I think this will clean up a lot of things in Protractor both with Angular apps and in general.

This is on our timeline for end of 2015.

opichals:
@stalniy @juliemr Could you please elaborate on your ideas of the framework spacific waiter API and the separate process workflow and API?

It would be clean for the protractor core not to contain any angular specific waiting and be generic in API terms.

[...]

PR #2586, opened 2015-10-07
feat(lib): add support for waiting for angular2

### LINKED ISSUE #2396: Update waitForAngular to use ng2's whenStable
This will allow Protractor to wait for ng2 apps

juliemr:
Use Angular2's testability API, if present, when waiting
for stability or loading a page.

Closes #2396

### DIFF
--- a/lib/clientsidescripts.js
+++ b/lib/clientsidescripts.js
@@ -50,6 +50,10 @@ functions.waitForAngular = function(rootSelector, callback) {
   var el = document.querySelector(rootSelector);
 
   try {
+    if (window.getAngularTestability) {
+      window.getAngularTestability(el).whenStable(callback);
+      return;
+    }
     if (!window.angular) {
       throw new Error('angular could not be found on the window');
     }
@@ -555,6 +559,8 @@ functions.findByCssContainingText = function(cssSelector, searchText, using) {
  *
  * @param {number} attempts Number of times to retry.
  * @param {function} asyncCallback callback
+ *
+ * @return {{version: ?number, message: ?string}}
  */
 functions.testForAngular = function(attempts, asyncCallback) {
   var callback = function(args) {
@@ -564,19 +570,21 @@ functions.testForAngular = function(attempts, asyncCallback) {
   };
   var check = function(n) {
     try {
-      if (window.angular && window.angular.resumeBootstrap) {
-        callback([true, null]);
+      if (window.getAllAngularTestabilities) {
+        callback({ver: 2});
+      } else if (window.angular && window.angular.resumeBootstrap) {
+        callback({ver: 1});
       } else if (n < 1) {
         if (window.angular) {
-          callback([false, 'angular never provided resumeBootstrap']);
+          callback({message: 'angular never provided resumeBootstrap'});
         } else {
-          callback([false, 'retries looking for angular exceeded']);
+          callback({message: 'retries looking for angular exceeded'});
         }
       } else {
         window.setTimeout(function() {check(n - 1);}, 1000);
       }
     } catch (e) {
-      callback([false, e]);
+      callback({message: e});
     }
   };
   check(attempts);
--- a/lib/clientsidescripts.js
+++ b/lib/clientsidescripts.js
@@ -38,7 +38,8 @@ function wrapWithHelpers(fun) {
 
 /**
  * Wait until Angular has finished rendering and has
- * no outstanding $http calls before continuing.
+ * no outstanding $http calls before continuing. The specific Angular app
+ * is determined by the rootSelector.
  *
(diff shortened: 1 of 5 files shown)


juliemr:
@hankduan could you take a look at this one?

The current implementation only waits for one Angular2 application per page - the one found using `config.rootElement`. If it finds an angular2 app, it will not try to find or wait for an angular1 application. An alternative would be to wait for all angular apps, or to add the behavior as a configuration option, but this seemed like the simplest initial setup to me - let me know if you disagree.

juliemr:
Hm, actually after thinking about this for a minute longer - at least when testing the angular2 example applications, there's a lot of cases where the default behavior that we want is waiting for all ng2 apps. That way, we don't have to specify a different selector every time. I'm working on a modification to do that instead.

juliemr:
OK, the branch has been updated with the addition of a new configuration option to wait for all angular2 applications on the page. This requires only one change to the config file. Let me know what you think of option name, it's kinda long.

lib/clientsidescripts.js
```diff
@@ -47,24 +47,46 @@ function wrapWithHelpers(fun) {
  *     be passed as a parameter.
  */
 functions.waitForAngular = function(rootSelector, callback) {
-  var el = document.querySelector(rootSelector);
-
-  try {
-    if (!window.angular) {
-      throw new Error('angular could not be found on the window');
+  if (!rootSelector) {
```
hankduan: Maybe I'm missing something, but wouldn't all ng1 tests fail right now? I guess by default the rootSelector is 'body' or something, but someone could have easily accidentally called 'useAllAngularAppRoots' for ng1. Maybe a safeguard?

juliemr: Yup, the root selector gets defaulted to 'body' in the constructor for a `Protractor` instance. How would someone set `useAllAngularAppRoots` to true accidentally?

lib/clientsidescripts.js
```diff
@@ -47,24 +47,46 @@ function wrapWithHelpers(fun) {
  *     be passed as a parameter.
  */
 functions.waitForAngular = function(rootSelector, callback) {
```
hankduan: Do you think the logic would be easier to understand if we split this up?

```
var version = testForAngular()
...
if (version == 1) {
  waitForAngular1(...)
} else {
  waitForAngular2(...)
}
```

If someone less knowledgeable of angular2 (i.e. contributor) reads this, he/she would need to understand that angular1 contains `window.angular`, and angular2 contains `window.getAllAngularTestabilities()` and `window.getAngularTestability()`. This code is really hard to understand unless you know that.

lib/clientsidescripts.js
```diff
@@ -555,6 +577,8 @@ functions.findByCssContainingText = function(cssSelector, searchText, using) {
  *
  * @param {number} attempts Number of times to retry.
  * @param {function} asyncCallback callback
+ *
+ * @return {{version: ?number, message: ?string}}
```
hankduan: This is the callback for the asyncCallback, not a return

lib/clientsidescripts.js
hankduan: By accidentally I mean if they specify that in their config. The config name itself doesn't mention anything about ng2, and a lot of people don't read the comments.

hankduan:
Anyway, the PR itself looks fine to me functionally. Just left some small suggestion.

stalniy:
why don't you want to implement #2460? Instead of adding `if-s`  you can create a simple and clear interface using Strategy pattern which will allow to test any app. 

In case dev want to test both Angular1 and Angular2 apps he can use Decorator pattern or smth like composition between 2 strategies (Angular1 and Angular2).

Angular2 haven't been released yet (just alpha), so we have time to implement this in transparent way.

juliemr:
@stalniy I've added a note at #2460 with my ideas on that issue. It's a slightly longer term goal. Angular2 has not yet been released, but it is already using Protractor to run tests, and many people are beginning to experiment with it and we need to start growing an e2e testing solution there.

lib/clientsidescripts.js
juliemr: We can't use `testForAngular` unless we modify the client side scripts strategy (as you suggested separately), so I'm punting on that one for now.

I split it into two different client side scripts, since it doesn't make sense to have it as an if/else case.

lib/clientsidescripts.js
juliemr: Good catch, done.

lib/clientsidescripts.js
juliemr: Fair point. Changed the name so that it's harder to mess up.

PR #2591, opened 2015-10-07
fix(configParser): Remove path.parse so protractor works with node < …

### LINKED ISSUE #2588: Latest release requires node v0.12 with 'no method 'parse''
The latest release (2.5.0) calls path.parse in configParser.js and that function does not exist in node versions prior to v0.12.

I don't see anything that indicates protractor requires v0.12, so I'm wondering if this was just an accident.

sjelin:
…v0.12

Closes #2588

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 sudo: false
 node_js:
+  - "0.10"
   - "0.12"
   - "4"
 
@@ -20,6 +21,8 @@ matrix:
   exclude:
     - env: JOB=smoke
       node_js: "0.12"
+    - env: JOB=smoke
+      node_js: "0.10"
 
 
 before_script:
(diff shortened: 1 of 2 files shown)


NickTomlin:
This handles windows/unix paths. LGTM :)

.travis.yml
```diff
@@ -20,6 +21,9 @@ matrix:
   exclude:
     - env: JOB=smoke
       node_js: "0.12"
```
juliemr: It looks like this is no longer excluding smoke jobs with nodejs 0.12 :(

That's not the end of the world - can we try it with:

```
exclude:
  - env: JOB-smoke
    node_js: "0.12"
  - env: JOB-smoke
    node_js "0.10"

juliemr:
One change to the yml, then lgtm!

PR #2611, opened 2015-10-15
Docs: onPrepare should return a promise

konrad-garus:


konrad-garus:
CLA signed as konrad.garus@gmail.com

martin770:
Cool, I've had to use this in my onPrepare as well, returning the promise that adds my jasmine reporters.

I started using browser.getCapabilities() to get the browser name and sessionId to use as a unique filename for protractor-jasmine2-screenshot-reporter. If I do not return this promise, I see occasional failures where it seems like the tests are starting or finishing before the reporter can get hooked up properly.

Before using browser.getCapabilities(), I would just add the reporter using its default settings, and it worked without having to return anything in onPrepare.

I wasn't sure why this worked exactly, but I found information about it in the jasmine-reporters github project issues: https://github.com/larrymyers/jasmine-reporters/issues/94

docs/system-setup.md
```diff
@@ -8,4 +8,8 @@ There are a couple of things to watch out for!
 
 **If your page uses `$timeout` for polling** Protractor will not be able to tell when your page is ready. Consider using `$interval` instead of `$timeout`.
 
-If you need to do global preparation for your tests (for example, logging in), you can put this into the config in the `onPrepare` property. This property can be either a function or a filename. If a filename, Protractor will load that file with Node.js and run its contents. See the [login tests](https://github.com/angular/protractor/blob/master/spec/withLoginConf.js) for an example.
+**If you need to do global preparation for your tests* (for example, logging in), you can put this into the config in the `onPrepare` property. This property can be either a function or a filename. If a filename, Protractor will load that file with Node.js and run its contents.
```
sjelin: You need a second `*` to close the `**`.  Also, why did you split this into three separate paragraphs?

konrad-garus: Sorry about closing the bold, will fix.

The reason for splitting is the regular English readability.

docs/referenceConf.js
```diff
@@ -206,7 +210,7 @@ exports.config = {
     //
     // If you need access back to the current configuration object,
     // use a pattern like the following:
-    //     browser.getProcessedConfig().then(function(config) {
+    //     return browser.getProcessedConfig().then(function(config) {
```
hankduan: Why do you need to return here? It should be up to the user.

docs/referenceConf.js
```diff
@@ -197,6 +197,10 @@ exports.config = {
   // capability.
   // You can specify a file containing code to run by setting onPrepare to
   // the filename string.
+  // If the preparation involves any asynchronous calls (e.g. interacting with
+  // the browser), it should return a promise. Otherwise Protractor will be
```
hankduan: Maybe something along this line:
OnPrepare can optionally return a promise, which Protractor will wait for before continuing execution. If the preparation involves any asynchronous calls (e.g. interacting with the browser), notify Protractor to wait by returning a promise.

hankduan:
Thanks for the PR! Please fix the comments, and add the '**'. I'll merge it in once it's done.

docs/referenceConf.js
konrad-garus: Shouldn't this be the default? It indicates that you can return and it will be used for something.

It seems like a better idea than not returning anything here and having the user scratch their head, spend a long time googling, searching docs and reading source code. That's exactly why I came up with this PR.

Besides, leaving a loose promise more often than not is not what you want.

konrad-garus:
I've just pushed an update, closer to what @hankduan suggested.

docs/referenceConf.js
hankduan: Ok, that's fine. I only mentioned that because there are multiple ways of using this information without returning it, but you could be right that returning the value serves as an important use case.

hankduan:
Thanks for the PR and the changes. It's merged here: https://github.com/angular/protractor/commit/827554148d2af8e3dc1fdeb4481bfb906a98d116

PR #2615, opened 2015-10-17
bump source-map-support

charlierudolph:
In order to integrate with the a new version of cucumber-js, need source-map-support to be up to date.
cucumber-js listens for uncaughtException, as does source-map-support. With 0.3.2, source-map-support will not exit on uncaughtException, if there are other listeners.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "optimist": "~0.6.0",
     "q": "1.0.0",
     "lodash": "~2.4.1",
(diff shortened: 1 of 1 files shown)


charlierudolph:
I signed it!

hankduan:
Thanks @charlierudolph for looking into the issue. Merged with https://github.com/angular/protractor/commit/372263f93acf94334eee0547a4ad4152dd8a5842

PR #2618, opened 2015-10-19
feat(plugins): Add config option to disable logging of warnings in co…

paraplexed:
…nsole plugin

Running tests in multiple browsers ends up printing out a lot of useless warnings I'm already aware of. To make skimming through logs in the case of an actual failure easier, I want to be able to disable the logging of warnings in the console plugin.

### DIFF
--- a/plugins/console/index.js
+++ b/plugins/console/index.js
@@ -5,13 +5,14 @@ var q = require('q');
  * It can be configured to fail a test if either is detected.  There is also an
  * optional exclude parameter which accepts both regex and strings.  Any log
  * matching the exclude parameter will not fail the test or be logged to the
- * console.
+ * console. A false setting to logWarnings also overrides the failOnWarning setting.
  *
  *    exports.config = {
  *      plugins: [{
  *        path: 'node_modules/protractor/plugins/console',
  *        failOnWarning: {Boolean}                (Default - false),
- *        failOnError: {Boolean}                  (Default - true)
+ *        failOnError: {Boolean}                  (Default - true),
+ *        logWarnings: {Boolean}                  (Default - true),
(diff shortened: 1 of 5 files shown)


paraplexed:
I signed it!

plugins/console/spec/consoleFailLogWarnings.js
```diff
@@ -0,0 +1,14 @@
+var env = require('../../../spec/environment.js');
```
sjelin: These tests should be added to `scripts/test.js`

plugins/console/index.js
```diff
@@ -76,14 +79,18 @@ ConsolePlugin.parseLog = function(context) {
...
+      context.config.logWarnings;
   ConsolePlugin.exclude = context.config.exclude || [];
 
   return ConsolePlugin.getBrowserLog().then(function(log) {
-
-    var warnings = log.filter(function(node) {
-      return (node.level || {}).name === 'WARNING' &&
-          ConsolePlugin.includeLog(node.message);
-    });
+    var warnings = undefined;
```
sjelin: Better than making an edge case where `warnings` is `undefined`, why not use `var warnings = []`?  That way you won't have to add complexity on line 43

paraplexed: Very good point. Fixing that now.

plugins/console/spec/consoleFailLogWarnings.js
paraplexed: Ah I didn't realize that that's where they were run from. I'll get them added to that and update the PR in a little bit.

Thanks for the feedback!

paraplexed:
Both concerns should be addressed now. Let me know if there are others. Thanks!

sjelin:
Merged in at a54c0e0

PR #2652, opened 2015-10-30
chore(tests): test doesn't work without network

hankduan:


hankduan:
merged with https://github.com/angular/protractor/commit/31d1b72aba8ec809195c12cb7cff3d3472c07ad1

PR #2654, opened 2015-11-01
test(jasmine): add failing test for custom matchers on a not-present element

tepez:
Custom matchers fail if used on an element that is not present (not in the DOM), but only if used more than 2 times.

This specs show this. Will open an issue after the travis build fails.


NickTomlin:
Thanks for bringing this up. I feel like this is better handled in the issue you opened unless your PR is updated with a fix; you could still keep your fork as an example. I'll defer to an actual maintainer on that though :)

tepez:
I made the PR so that the maintainers could easily see how the specs are failing on the Travis build. It's not intended to be merged.

juliemr:
Thanks for the actual example, this is helpful. This smells like a webdriver control flow issue - let me try it with the latest master.

juliemr:
With the latest webdriver, this does not give the element not found errors. Instead, it occasionally errors because it doesn't grab the correct text from the element. The problem isn't deterministic. 

However, the failures only occur when doing the animation toggle. So I think that the issues are with waiting for the animation, not with the custom matcher.

Can you reproduce on master with something other than the animation toggle?

tepez:
I ran my test suit on the master branch of protractor and it didn't have this issue anymore.
I'll close the issue.

Are you interested in merging in any of these specs (once a proper wait for the animation has been added)?

Myself, I feel like protractor test suit could benefit from specs on custom matchers, in file dedicated for that, since currently there is only one spec for that.

juliemr:
Sure, I'd be happy to take a look at a PR adding more tests. Feel free to open up a new one! I don't think it has to be related to the animation stuff at all. For now, I'm going to close this since it seems to no longer be an issue.

PR #2666, opened 2015-11-04
feat(runner): make source map support optional

### LINKED ISSUE #2603: protractor registers global `source-map-support`
i found this line in the code
https://github.com/angular/protractor/blob/master/lib/runner.js#L162
and it messes up with my project's environment. I agree that some globals like `$` or `protractor` can be an exception in test environment, but things like that should be at least optional.
Is there any way it can be disabled/configured?

sjelin:
Closes #2603

### DIFF
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -311,5 +311,12 @@ exports.config = {
     tags: '@dev',
     // How to format features (default: progress)
     format: 'summary'
-  }
+  },
+
+  // See docs/plugins.md
+  plugins: [],
+
+  // Turn off source map support.  Stops protractor from registering global
+  // variable `source-map-support`.  Defaults to `false`
+  skipSourceMapSupport: false
 };
--- a/lib/configParser.js
+++ b/lib/configParser.js
@@ -41,7 +41,8 @@ var ConfigParser = function() {
     },
     chromeDriver: null,
     configDir: './',
-    plugins: []
+    plugins: [],
+    skipSourceMapSupport: false
   };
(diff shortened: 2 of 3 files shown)


docs/referenceConf.js
```diff
@@ -311,5 +311,12 @@ exports.config = {
     tags: '@dev',
     // How to format features (default: progress)
     format: 'summary'
-  }
+  },
+
+  // See docs/plugins.md
+  plugins: [],
+
+  // Turn of source map support.  Stops protractor from registering global
```
hankduan: off

hankduan:
Optimist doesn't handle string parsing. So `protractor conf.js --sourceMap false` will be read as 'sourceMap="false"', which will evaluate to truthy. 

Users then have to do 'protractor conf.js --sourceMap 0' or 'protractor conf.js --no-sourceMap', which they will probably not know and feel it's a bug. 

Let's call it --skipSourceMapSupport, and flip the boolean.

sjelin:
Comments addressed

docs/referenceConf.js
```diff
@@ -311,5 +311,12 @@ exports.config = {
     tags: '@dev',
     // How to format features (default: progress)
     format: 'summary'
-  }
+  },
+
+  // See docs/plugins.md
+  plugins: [],
+
+  // Turn off source map support.  Stops protractor from registering global
```
hankduan: nit: Turns off

sjelin:
Merged in with 4924028

PR #2670, opened 2015-11-04
feat(ElementArrayFinder#get): Implemented ability to pass promise

### LINKED ISSUE #2646: Allow promise to be passed to ElementArrayFinder#get
For example I have tabs and I want access tab content by tab name. In this case I need to: 
```js
index = element.all(by.css).getText().then((names) =>{
  return names.indexOf(tabName);
});

tabContent = element.all(by.css).get(index)
```
Also this one is inconsistent with `sendKeys` which accepts promise as a parameter

stalniy:
It fixes #2646


sjelin:
Merged in with a40a4ba

Issue #1435, opened 2014-10-13
Feature Request: to have an option to restart new browser session between each test case/scenario


jlin412:
It is kind of related to https://github.com/angular/protractor/issues/1431.
I searched on stackoverflow.com and there was no good ways to start a new browser session in each test case whether it is cucumber or Jesmine framework.
The only way to ensure each test case is independent from each other is to have a new session in each test case.  When I am used to test with Capybara in Ruby, that is how it works.
I think that protractor should do the same or at very least give an option from commandline to do so.

vipper:
Is there a way to restart the browser for a specific test?

monkpit:
@hankduan I was confused by your link to 1638, so I'm adding a link to the actual solution here:

https://github.com/angular/protractor/commit/167038499aacfd5def03472f9f548529b273e1e0

PR #2685, opened 2015-11-07
make restart() available on browser instance

### LINKED ISSUE #1696: Adding a new method to restart browser: browser.restart()
Thought we implement restart config for restart between test case (it/scenario), I think that it can be even more useful if we can have browser.restart() method so we can put into an step/scenario/feature/suite/it/describe level depending on test framework and users' needs.

### LINKED ISSUE #1785: restart browser with each describe
now we have the restart browser with each test , it costs so much times.
```
// If true, protractor will restart the browser between each test. 
  // CAUTION: This will cause your tests to slow down drastically.
  restartBrowserBetweenTests: false,
```

can we restart browser with each describe or with each file

#1435

avdd:
A shame to have such a useful function buried in a private scope.  Closes #1696
Allows usage in beforeAll() to close  #1785 and a more general solution to #1435

### DIFF
--- a/lib/protractor.js
+++ b/lib/protractor.js
@@ -267,6 +267,12 @@ Protractor.prototype.getProcessedConfig = null;
  */
 Protractor.prototype.forkNewDriverInstance = null;
 
+/**
+ * Restart the browser instance.
+ *
+ * Set by the runner.
+ */
+Protractor.prototype.restart = null;
 
 /**
  * Instead of using a single root element, search through all angular apps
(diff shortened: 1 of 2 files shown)


juliemr:
We'll need to add a dummy function to lib/protractor, like we do for `forkNewBrowser` here: https://github.com/angular/protractor/blob/master/lib/protractor.js#L269

Otherwise, looks reasonable to me. @hankduan can you think of a reason this would cause trouble?

avdd:
CLA signed.

avdd:
@juliemr done!

hankduan:
Thanks for the PR. Merged here: https://github.com/angular/protractor/commit/97e6703eb0e7a5dffc1017d0a44f0dfeeb91f327

PR #2687, opened 2015-11-09
chore(npm/github): update ignore files

sjelin:
Might be a little aggressive.  Definitely need someone to double check.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,17 +1,20 @@
 debugging/
+docs/
+example/
+scripts/
 spec/
 stress/
 selenium/
-debugging/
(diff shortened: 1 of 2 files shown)


.npmignore
```diff
@@ -1,17 +1,20 @@
 debugging/
+docs/
+example/
```
juliemr: Let's keep the example so people have a tiny thing to run. It's not concerning in terms of file size.

juliemr:
Looks good after the one comment.

sjelin:
Merged in with f0e2195

Issue #2159, opened 2015-05-19
Make cucumber support a separate project


juliemr:
Cucumber support has been sketchy and community-driven for a while. See the disclaimer at http://angular.github.io/protractor/#/frameworks#using-cucumber

Now that a custom framework can be used, we should just make it a separate project to clarify this separation, and unblock protractor releases when weird things are going on with cucumber.

mattfritz:
I'd like to take a crack at this. When are you planning on creating another repo for the cucumber custom framework integration?

sjelin:
Yeah, I actually just started on this!  We need to merge https://github.com/angular/protractor/pull/2691 first, and then we can split cucumber out.  Probably going to publish to an npm module like `protractor-cucumber-framework` or something, at which point people would be able to use [`frameworkPath: 'protractor-cucumber-framework'`](https://github.com/angular/protractor/tree/master/lib/frameworks#custom-frameworks) to access the framework.

We're actually looking for an owner for the project though!  We (the core team) have never really used or understood much about cucumber, and so support was always kind of spotty.  If you're interested in taking control that'd be awesome!  (If not that's okay too :stuck_out_tongue:)

mattfritz:
Sure, if you're looking for an owner I can maintain the framework module. I just signed a CLA and subscribed to https://github.com/angular/protractor/pull/2691 for updates. Thanks!

sjelin:
Ok, #2691 isn't actually necessary (my bad).  So we're good to go with separation!  Here's an outline of how I'd do it (feel free to ignore):

* Clone https://github.com/angular/protractor-timeline-plugin.  Just delete `index.js` and everything in the spec folder except `environment.js`.  This'll give you a tiny version of the test app which cucumber currently runs against.

* Move `lib/frameworks/cucumber.js` to the new projects' `index.js`

* Move [`spec/cucumberConf.js`](https://github.com/angular/protractor/blob/master/spec/cucumberConf.js) and [`spec/cucumber/`](https://github.com/angular/protractor/tree/master/spec/cucumber) into the new repo's `spec/` folder.

* Update `cucumberConf.js` to use a [custom framework](https://github.com/angular/protractor/tree/master/lib/frameworks#custom-frameworks) pointing to `index.js`

* Possibly fix some relative paths

* In `test.js`, update `passingTests` to `['node node_modules/protractor/lib/cli.js spec/cucumberConf.js']`

* Fix the `package.json`

[...]

PR #2691, opened 2015-11-09
chore(frameworks): Remove cucumber dependency on ConfigParser

sjelin:
First step towards splitting cucumber out (#2159)

### DIFF
--- a/lib/frameworks/cucumber.js
+++ b/lib/frameworks/cucumber.js
@@ -1,14 +1,15 @@
-var ConfigParser = require('../configParser'),
-  q = require('q');
+var q = require('q');
 
 /**
  * Execute the Runner's test cases through Cucumber.
  *
  * @param {Runner} runner The current Protractor Runner.
  * @param {Array} specs Array of Directory Path Strings.
+ * @param {function(string): string} resolvePath resolves a path relative to the
+ *     config file
  * @return {q.Promise} Promise resolved with the test results
  */
-exports.run = function(runner, specs) {
+exports.run = function(runner, specs, resolvePath) {
   // TODO - add the event interface for cucumber.
   var Cucumber = require('cucumber'),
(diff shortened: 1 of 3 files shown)


lib/frameworks/cucumber.js
```diff
@@ -1,14 +1,15 @@
...
  * Execute the Runner's test cases through Cucumber.
  *
  * @param {Runner} runner The current Protractor Runner.
  * @param {Array} specs Array of Directory Path Strings.
+ * @param {function(string): string} resolvePath resolves a path relative to the
+ *     config file
  * @return {q.Promise} Promise resolved with the test results
  */
-exports.run = function(runner, specs) {
+exports.run = function(runner, specs, resolvePath) {
```
juliemr: This seems like a messy way of passing this information.

The problem here is that some plugins or frameworks may want to add their own things to the configuration file which will need to be interpreted as file paths, which need to be resolved relative to something. This is fairly simple logic which the frameworks could probably do themselves, but we have an internal helper. Is it even worth exporting this internal helper? Most of the reason that `ConfigParser.resoveFilePatterns` is so long is ONLY to support cucumber stuff. 

I would say let's not export this function anywhere, simplify the internal implementation, and make cucumber do its own complicated logic with line numbers. What do you think?

sjelin: Seams reasonable but we will have to export the config file's directory so cucumber can do its logic.

juliemr: why? `config.configDir` isn't enough?

sjelin: Ah, good point

sjelin:
Canceling this commit, as it turns out we can remove cucumber without it.  Will simplify `ConfigParser` in the same commit as when we remove cucumber.

PR #2692, opened 2015-11-09
Add example command to run more than one suite of tests

phil-lgr:
added example command to run more than one suite of tests on the command line


phil-lgr:
I signed it!

sjelin:
Merged in with 2b025f6

PR #2696, opened 2015-11-10
chore(docs): include `webdriver.Key` and `webdriver.promise` in API page

### LINKED ISSUE #402: docs: Should provide API info on webdriver.FOO interfaces like promises and KEY
Some links to these should be added to https://github.com/angular/protractor/blob/master/docs/api.md

sjelin:
Closes #402

### DIFF
--- a/website/docgen/processors/filter-promise.js
+++ b/website/docgen/processors/filter-promise.js
@@ -1,0 +1,24 @@
+var _ = require('lodash');
+
+/**
+ * Remove docs of private variables from selenium's promise.js file.
+ *
+ * @param {Array.<Object>} docs The jsdoc list.
+ */
+var filter = function(docs) {
+  return _.reject(docs, function(doc) {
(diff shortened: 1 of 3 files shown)


juliemr:
There's a bunch of gross stuff at the end of the promise API that users don't care about. Can we do anything about these?
screen shot 2015-11-10 at 8 00 24 am

sjelin:
Made the change we discussed offline.  Let me know what you think.  :hammer: :wrench: :scissors:

website/docgen/processors/filter-promise.js
```diff
@@ -0,0 +1,24 @@
+var _ = require('lodash');
+
+/**
+ * Remove docs of private variables from selenium's promise.js file
```
juliemr: nit: add '.' at end of sentence.

juliemr:
LGTM.

PR #2706, opened 2015-11-15
fix(cucumber): allow the use of compilers for cucumber

azangru:
As [mentioned in Cucumber's Readme](https://github.com/cucumber/cucumber-js#transpilers), Cucumber currently has an option to use transpilers so that step definitions and support files can be written in languages that transpile to JavaScript. This can be done using the `--compiler` CLI option  and specifying the transpiler for a particular file type:

Currently, Protractor [has no way](https://github.com/angular/protractor/blob/master/lib/frameworks/cucumber.js) of passing this option to Cucumber. In this PR, I updated the `lib/frameworks/cucumber.js` file to make it use the option `compilers` of the `cucumberOpts` object. Thus, step definition and support files can be written in ES6/ES7, Typescript or other JavaScript dialect.

Example:

```javascript
// protractor.conf.js

    cucumberOpts: {
        compilers: ['js:babel-core/register', 'ts:ts-node/register']
    },
```

### DIFF
--- a/lib/frameworks/cucumber.js
+++ b/lib/frameworks/cucumber.js
@@ -56,6 +56,17 @@ exports.run = function(runner, specs) {
       execOptions.push(runner.getConfig().cucumberOpts.format);
     }
 
+    // Process Cucumber Compiler param
+    if (Array.isArray(runner.getConfig().cucumberOpts.compilers)) {
+      runner.getConfig().cucumberOpts.compilers.forEach(function (compiler) {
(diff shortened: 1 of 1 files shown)


azangru:
I signed it.

sjelin:
Hey, we're moving cucumber support into a separate repo/npm module.  Please send your pull request to https://github.com/mattfritz/protractor-cucumber-framework

PR #2727, opened 2015-11-19
fix(ci): Install g++ 4.8 on Travis to prevent compilation errors

mgol:
Node.js 4 and newer require g++ newer than the one included in Travis by
default. Using the default g++ causes `npm install` to fail on compiled
dependencies - currently every such dependency of Protractor is optional
so the whole build doesn't fail but it still causes errors & excessive
logging.

If the number of compiled dependencies grows, install logs might become so
big that Travis stops - Karma has been hit by this problem recently.

Refs karma-runner/karma#1672


juliemr:
Thanks! Sorry for the slow response - merged as 80a64ffab81ba16f15052421f948361c217ffcae

PR #2736, opened 2015-11-24
docs(readme): compatibility with node

jakub-g:


README.md
```diff
@@ -3,6 +3,12 @@ Protractor [![Build Status](https://travis-ci.org/angular/protractor.png?branch=
 
 [Protractor](http://angular.github.io/protractor) is an end-to-end test framework for [AngularJS](http://angularjs.org/) applications. Protractor is a [Node.js](http://nodejs.org/) program built on top of [WebDriverJS](https://github.com/SeleniumHQ/selenium/wiki/WebDriverJs). Protractor runs tests against your application running in a real browser, interacting with it as a user would. 
 
+Compatibility
+-------------
+
+Protractor 3 is compatible with nodejs v4 and newer.
+
+When using nodejs v0.12, use protractor 2 (`npm install -g protractor@2`).
```
sjelin: Probably `2.5.1` would be better right?

jakub-g: when you do `npm install -g protractor@2` it will try to install the newest possible, so 2.5.1 for now. But if a new patch is ever released, if we put `npm install -g protractor@2.5.1` it will just install 2.5.1 and not let's say 2.5.2. So I think it's fine to leave it like this. But I can change if you want.

sjelin: Nope, `@2` makes sense to me then.  I'm just not a power user of `npm` so I didn't know how that worked.

NickTomlin: Yeah `@2` makes sense to me as well :+1:

sjelin:
Merged in with 54dd3db

PR #2751, opened 2015-11-30
chore(configParser): allow non-glob file pattern

darrinholst:
re: the conversation at https://github.com/angular/protractor/pull/2445#issuecomment-159019360

Cucumber allows line numbers to be passed in the filename in the
form of `features/some.feature:42`. Glob expanding that results in
an empty array and nothing being passed to the framework runner.
This change checks for glob magic characters and only tries expanding
it if found. Otherwise it just passes the filename verbatim.
This was previously handled in #2445 by stripping the line number
first, but this is a more generic (non-cucumber) way to do it.

### DIFF
--- a/lib/configParser.js
+++ b/lib/configParser.js
@@ -95,7 +95,7 @@ ConfigParser.resolveFilePatterns =
   if (patterns) {
     for (var i = 0; i < patterns.length; ++i) {
       var fileName = patterns[i];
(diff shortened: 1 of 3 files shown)


darrinholst:
_sighs_

I'll try again when it works with npm 3

jlin412:
You don't have to close it.

darrinholst:
wasn't sure if/when I could fix it

PR #2777, opened 2015-12-11
driverProviders/direct.js: fix driver path generation for *nix platforms

runk:
I was trying to run protractor tests locally on mac os, but was having error messages that `selenium/chromedriver.exe` not found. Very confusing. Hopefully this change would make it more clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Protractor [![Build Status](https://travis-ci.org/angular/protractor.png?branch=master)](https://travis-ci.org/angular/protractor)
 ==========
 
-[Protractor](http://angular.github.io/protractor) is an end-to-end test framework for [AngularJS](http://angularjs.org/) applications. Protractor is a [Node.js](http://nodejs.org/) program built on top of [WebDriverJS](https://github.com/SeleniumHQ/selenium/wiki/WebDriverJs). Protractor runs tests against your application running in a real browser, interacting with it as a user would. 
+[Protractor](http://angular.github.io/protractor) is an end-to-end test framework for [AngularJS](http://angularjs.org/) applications. Protractor is a [Node.js](http://nodejs.org/) program built on top of [WebDriverJS](https://github.com/SeleniumHQ/selenium/wiki/WebDriverJs). Protractor runs tests against your application running in a real browser, interacting with it as a user would.
 
 
 Getting Started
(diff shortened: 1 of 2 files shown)


lib/driverProviders/direct.js
```diff
@@ -51,18 +51,16 @@ DirectDriverProvider.prototype.getNewDriver = function() {
...
   switch (this.config_.capabilities.browserName) {
     case 'chrome':
-      var chromeDriverFile = this.config_.chromeDriver ||
-          path.resolve(__dirname, '../../selenium/chromedriver');
+      var defaultChromeDriverPath = path.resolve(__dirname, '../../selenium/chromedriver');
 
-      // Check if file exists, if not try .exe or fail accordingly
-      if (!fs.existsSync(chromeDriverFile)) {
-        chromeDriverFile += '.exe';
+      if (process.platform.indexOf('win') === 0)
```
NickTomlin: Style nitpick, we use brackets in if statements in the rest of the project (it looks like we don't have that in the jshintrc, i'll add a PR to fix that)

README.md
```diff
@@ -48,6 +48,9 @@ Clone the github repository:
 
 Start up a selenium server. By default, the tests expect the selenium server to be running at `http://localhost:4444/wd/hub`.
 
+Download [ChromeDriver](https://sites.google.com/a/chromium.org/chromedriver/home) and place it under `./selenium`
```
NickTomlin: I think this is helpful to note, but it is probably easier to use the `webdriver-manager` executable instead of manually downloading chromedriver. `./bin/webdriver-manager update`

NickTomlin:
I think this more explicit approach makes sense to me. We should double check that this works on a windows box (I'll try to boot up a VM tomorrow).

README.md
runk: :+1:

lib/driverProviders/direct.js
runk: Sure. Not a big deal.

runk:
@NickTomlin your comments were addressed :+1:

runk:
@NickTomlin did you have chance to try it?

NickTomlin:
I ran the tests and the example under a windows 7 VM with a single failure (I think pretty safe to say that it is unrelated to this change), so i'm :+1: on merging this in. Thoughts @sjelin ?

Single Failed spec below:

screen shot 2015-12-14 at 10 00 25 am

runk:
Ping @sjelin :)

sjelin:
Sorry!  I'll merge a bunch of stuff in today

sjelin:
Merged in with f533341

PR #2778, opened 2015-12-11
chore(style): Add "curly" rule to jshintrc

NickTomlin:
This adds the curly option which I noticed was not being properly caught in CI. 

As a side note, how would we feel about switching to eslint and using the [google eslint config](https://github.com/google/eslint-config-google) module for style definitions? I'd be happy to do that work.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -17,5 +17,6 @@
   "sub": true,
   "undef": true,
   "quotmark": "single",
-  "evil": true
+  "evil": true,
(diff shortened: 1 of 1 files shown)


sjelin:
I would be fine with eslint but you should make an issue and tag julie in it

juliemr:
This change LGTM. @sjelin can you merge?

sjelin:
Merged in with dbde3c6

PR #2783, opened 2015-12-14
Possibility to disable WebDriver's environment variables overrides functionality

aalbul:
I spent few days fighting with capabilities issues (especially, with platform part of it) and this issues forced me to investigate Protractors and later WebDriver's source code.

Unfortunately, i am one of that stupid (or unlucky) person, who named one of his environment variables "SELENIUM_BROWSER". As a result, from the logs of selenium hub i found that Capabilities object didn't contain a "platform" field.

My investigation led me to this part of WebDriver's code base: http://selenium.googlecode.com/git/docs/api/javascript/source/builder.js.src.html#l368

Looks like, no matter what value you specify in your "capabilities" object, it will not be used and overridden by parsed values from "SELENIUM_BROWSER" environment variable.

One of the possible solutions is, of course, to rename the variable, but in my case this caused a lot of buthurt trying to understand the exact reason.

Looks like, some of Selenium users already had that issue that's why this method is introduced in WebDriver: http://selenium.googlecode.com/git/docs/api/javascript/source/builder.js.src.html#l124

*My pull request is quiet simple. I propose to give Protractor user's a chance to enable / disable this "environment variable overrides" feature*

### DIFF
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -336,5 +336,9 @@ exports.config = {
 
   // Turns off source map support.  Stops protractor from registering global
   // variable `source-map-support`.  Defaults to `false`
-  skipSourceMapSupport: false
+  skipSourceMapSupport: false,
+
+  // Turns off WebDriver's environment variables overrides to ignore any
+  // environment variable and to only use the configuration in this file.
+  environmentOverrides: false
 };
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -340,6 +340,6 @@ exports.config = {
 
   // Turns off WebDriver's environment variables overrides to ignore any
   // environment variable and to only use the configuration in this file.
-  // Defaults to `true`
-  environmentOverrides: true
+  // Defaults to `false`
+  disableEnvironmentOverrides: false
 };
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -340,5 +340,6 @@ exports.config = {
 
   // Turns off WebDriver's environment variables overrides to ignore any
   // environment variable and to only use the configuration in this file.
-  environmentOverrides: false
+  // Defaults to `true`
+  environmentOverrides: true
 };
--- a/lib/driverProviders/driverProvider.js
+++ b/lib/driverProviders/driverProvider.js
@@ -36,7 +36,8 @@ DriverProvider.prototype.getNewDriver = function() {
   var builder = new webdriver.Builder().
     usingServer(this.config_.seleniumAddress).
     withCapabilities(this.config_.capabilities);
-  if (!_.isUndefined(this.config_.environmentOverrides) && this.config_.environmentOverrides === false) {
+  if (!_.isUndefined(this.config_.environmentOverrides)
+      && this.config_.environmentOverrides === false) {
     builder.disableEnvironmentOverrides();
   }
   var newDriver = builder.build();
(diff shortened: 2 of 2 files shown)


lib/driverProviders/driverProvider.js
```diff
@@ -32,10 +33,14 @@ DriverProvider.prototype.getExistingDrivers = function() {
...
 DriverProvider.prototype.getNewDriver = function() {
-  var newDriver = new webdriver.Builder().
-      usingServer(this.config_.seleniumAddress).
-      withCapabilities(this.config_.capabilities).
-      build();
+  var builder = new webdriver.Builder().
+    usingServer(this.config_.seleniumAddress).
+    withCapabilities(this.config_.capabilities);
+  if (!_.isUndefined(this.config_.environmentOverrides)
+      && this.config_.environmentOverrides === false) {
```
sjelin: Shouldn't this be `=== true`?

docs/referenceConf.js
```diff
@@ -336,5 +336,9 @@ exports.config = {
 
   // Turns off source map support.  Stops protractor from registering global
   // variable `source-map-support`.  Defaults to `false`
-  skipSourceMapSupport: false
+  skipSourceMapSupport: false,
+
+  // Turns off WebDriver's environment variables overrides to ignore any
+  // environment variable and to only use the configuration in this file.
```
sjelin: Mention the default

lib/driverProviders/driverProvider.js
aalbul: Not exactly. The idea is to be able to turn it off. By default it's turned on.

docs/referenceConf.js
aalbul: Done

lib/driverProviders/driverProvider.js
sjelin: Oh I see.  Isn't the `!_.isUndefined(this.config_.environmentOverrides)` check redundant then?  If `this.config_.environmentOverrides === false` then it certainly won't be `undefined`.

aalbul: Hm. Good point :)

aalbul:
Code cleaned up

lib/driverProviders/driverProvider.js
```diff
@@ -5,6 +5,7 @@
  */
 
 var webdriver = require('selenium-webdriver'),
+    _ = require('lodash'),
```
sjelin: Looks like this isn't used anywhere anymore.

jpaulin: Hey, 

there's only a little thing to do anymore, to make the Pull request succeed. Let's do it! :) 

The above line thus should only have:
var webdriver = require('selenium-webdriver') 

and nothing else. The part mentioning _ (loading lodash module) should be removed.

aalbul: Sorry guys for the slow reaction. Now it's fixed

aalbul:
Lodash dependency removed

docs/referenceConf.js
```diff
@@ -336,5 +336,10 @@ exports.config = {
...
+  // Defaults to `true`
+  environmentOverrides: true
```
juliemr: Adding it here doesn't make it default to true, you need to change the default in https://github.com/angular/protractor/blob/master/lib/configParser.js.

But instead - let's name the variable `disableEnvironmentOverrides` and make it default to false. Having bools default to false is so much easier :)

juliemr:
I'm sorry for the slow response on merging here. Looks good except for my comment above. Let me know if you just want me to go ahead and make that change when I merge.

aalbul:
@juliemr done. Please verify the change

juliemr:
Oops, looks like this was actually already merged as fb10c5caffb895e909ad91d629e2192c74c8e064. I'm gong to submit a clean to get it to the state I suggested, then close this pr.

PR #2822, opened 2016-01-06
 Add a slice method to ElementArrayFinder

akarmarkar:
Currently, we have no way of taking a slice of all elements of an ElementArrayFinder. The slicing syntax is modelled after the slice syntax in python. ie : start defaults to 0, stop defaults to one past the last element in the array and step defaults to 1.

### DIFF
--- a/lib/element.js
+++ b/lib/element.js
@@ -161,6 +161,100 @@ ElementArrayFinder.prototype.all = function(locator) {
 };
 
 /**
+ * Get a slice of elements within the ElementArrayFinder.
+ * Negative indices are wrapped (i.e. -i means ith element from last)
+ * This does not actually retrieve the underlying element.
+ *
+ * Slice can be negative in which case we go backwards.
+ *
+ * Note that if step is positive and start > stop, an error is thrown
+ * Note that if step is negative and start < stop, an error is thrown
+ * Note that slice(null,null,-1) will NOT produce a reversed array - use reverse() on the result instead.
+ *
+ * @view
+ * <ul class="items">
+ *   <li>Zeroth</li>
+ *   <li>First</li>
+ *   <li>Second</li>
+ *   <li>Third</li>
+ *   <li>Fourth</li>
+ *   <li>Fifth</li>
+ *   <li>Sixth</li>
+ *   <li>Seventh</li>
+ * </ul>
+ *
+ * @example
+ * var list = element.all(by.css('.items li'));
+ *
+ * expect(element.all(locator).slice(1,4).getText()).toEqual(['First','Second','Third']);
+ * expect(element.all(locator).slice(6).getText()).toEqual(['Sixth','Seventh']);
+ * expect(element.all(locator).slice(-1,-5,-2).getText()).toEqual(['Seventh','Fifth']);
+ *
+ * @param {sliceStart} start of the slice. Defaults to 0.
+ * @param {sliceStop} end of the slice(not included). Defaults to last element.
+ * @param {sliceStep} The jump. Defaults to 1.
+ * @return {ElementArrayFinder} finder representing sliced elements
+ */
+ElementArrayFinder.prototype.slice = function (sliceStart,sliceStop,sliceStep) {
+  var self = this;
+
+  var getWebElements = function () {
+    return self.getWebElements().then(function (parentWebElements) {
+      var start = (sliceStart === undefined || sliceStart === null) ? 0 : sliceStart;
+      var stop = (sliceStop === undefined || sliceStop === null) ? parentWebElements.length : sliceStop;
+      var step = (sliceStep === undefined || sliceStep === null || sliceStep === 0) ? 1 : sliceStep;
+
+      if (start < 0) {
+        // wrap negative indices
+        start = parentWebElements.length + start;
+      }
+      if (stop < 0) {
+        // wrap negative indices
+        stop = parentWebElements.length + stop;
+      }
+      if (start > parentWebElements.length) {
+        throw new Error('Start ' + start + ' > result length ' + parentWebElements.length);
+      }
+      if (stop > parentWebElements.length) {
+        throw new Error('Stop ' + stop + ' > result length ' + parentWebElements.length);
+      }
+
+      var list = [];
+      var lt = function (x, y) {
+        return x < y;
+      };
+      var gt = function (x, y) {
+        return x > y;
+      };
+      var compare = lt;
+
+      if (step < 0) {
+        compare = gt;
+        // If step is negative and start is less than stop, error out.
+        if (start < stop) {
+          throw new Error('Start ' + start + ' < Stop ' + stop + ' for negative step value : ' + step);
+        }
+      } else {
+        // If step is positive and start is greater than stop, error out.
+        if (start > stop) {
+          throw new Error('Start ' + start + ' > Stop ' + stop + ' for positive step value: ' + step);
+        }
+      }
+
+      var current = start;
+      while (compare(current, stop)) {
+        list.push(parentWebElements[current]);
+        current += step;
+      }
+
+      return list;
+    });
+  };
+
+  return new ElementArrayFinder(this.ptor_, getWebElements, this.locator_);
+};
+
+/**
  * Apply a filter function to each element within the ElementArrayFinder. Returns
  * a new ElementArrayFinder with all elements that pass the filter function. The
  * filter function receives the ElementFinder as the first argument


lib/element.js
```diff
@@ -161,6 +161,100 @@ ElementArrayFinder.prototype.all = function(locator) {
...
 /**
+ * Get a slice of elements within the ElementArrayFinder.
+ * Negative indices are wrapped (i.e. -i means ith element from last)
+ * This does not actually retrieve the underlying element.
+ *
+ * Slice can be negative in which case we go backwards.
+ *
+ * Note that if step is positive and start > stop, an error is thrown
+ * Note that if step is negative and start < stop, an error is thrown
+ * Note that slice(null,null,-1) will NOT produce a reversed array - use reverse() on the result instead.
```
sjelin: This line is too long (80 max)

lib/element.js
```diff
@@ -161,6 +161,100 @@ ElementArrayFinder.prototype.all = function(locator) {
...
+ * @param {sliceStep} The jump. Defaults to 1.
+ * @return {ElementArrayFinder} finder representing sliced elements
+ */
+ElementArrayFinder.prototype.slice = function (sliceStart,sliceStop,sliceStep) {
+  var self = this;
+
+  var getWebElements = function () {
+    return self.getWebElements().then(function (parentWebElements) {
+      var start = (sliceStart === undefined || sliceStart === null) ? 0 : sliceStart;
+      var stop = (sliceStop === undefined || sliceStop === null) ? parentWebElements.length : sliceStop;
```
sjelin: This line is too long (80 max)

lib/element.js
```diff
@@ -161,6 +161,100 @@ ElementArrayFinder.prototype.all = function(locator) {
...
+      var gt = function (x, y) {
+        return x > y;
+      };
+      var compare = lt;
+
+      if (step < 0) {
+        compare = gt;
+        // If step is negative and start is less than stop, error out.
+        if (start < stop) {
+          throw new Error('Start ' + start + ' < Stop ' + stop + ' for negative step value : ' + step);
```
sjelin: This line is too long (80 max)

lib/element.js
```diff
@@ -161,6 +161,100 @@ ElementArrayFinder.prototype.all = function(locator) {
...
+        }
+      } else {
+        // If step is positive and start is greater than stop, error out.
+        if (start > stop) {
+          throw new Error('Start ' + start + ' > Stop ' + stop + ' for positive step value: ' + step);
```
sjelin: This line is too long (80 max)

lib/element.js
```diff
@@ -161,6 +161,100 @@ ElementArrayFinder.prototype.all = function(locator) {
...
+ * </ul>
+ *
+ * @example
+ * var list = element.all(by.css('.items li'));
+ *
+ * expect(element.all(locator).slice(1,4).getText()).toEqual(['First','Second','Third']);
+ * expect(element.all(locator).slice(6).getText()).toEqual(['Sixth','Seventh']);
+ * expect(element.all(locator).slice(-1,-5,-2).getText()).toEqual(['Seventh','Fifth']);
+ *
+ * @param {sliceStart} start of the slice. Defaults to 0.
```
sjelin: should be `@param {number} sliceStart`

lib/element.js
```diff
@@ -161,6 +161,100 @@ ElementArrayFinder.prototype.all = function(locator) {
...
+ * @param {sliceStop} end of the slice(not included). Defaults to last element.
```
sjelin: Should be `@param {number=} sliceStop`

lib/element.js
```diff
@@ -161,6 +161,100 @@ ElementArrayFinder.prototype.all = function(locator) {
...
+ * @param {sliceStep} The jump. Defaults to 1.
```
sjelin: Should be `@param {number=} sliceStep`

lib/element.js
```diff
@@ -161,6 +161,100 @@ ElementArrayFinder.prototype.all = function(locator) {
 };
 
 /**
+ * Get a slice of elements within the ElementArrayFinder.
+ * Negative indices are wrapped (i.e. -i means ith element from last)
+ * This does not actually retrieve the underlying element.
+ *
+ * Slice can be negative in which case we go backwards.
+ *
+ * Note that if step is positive and start > stop, an error is thrown
```
sjelin: This isn't quite right.  For instance, `elems.slice(2,-1)` should be valid.    Also, to be consistent with [`Array.prototype.slice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice), I think we should return an empty array in this case.

lib/element.js
```diff
@@ -161,6 +161,100 @@ ElementArrayFinder.prototype.all = function(locator) {
 };
 
 /**
+ * Get a slice of elements within the ElementArrayFinder.
+ * Negative indices are wrapped (i.e. -i means ith element from last)
+ * This does not actually retrieve the underlying element.
+ *
+ * Slice can be negative in which case we go backwards.
```
sjelin: Should be "Step can be negative".  Also, why does this feature exist?  I don't see it in other slice implementations.

lib/element.js
```diff
@@ -161,6 +161,100 @@ ElementArrayFinder.prototype.all = function(locator) {
...
+ *
+ * @param {sliceStart} start of the slice. Defaults to 0.
+ * @param {sliceStop} end of the slice(not included). Defaults to last element.
+ * @param {sliceStep} The jump. Defaults to 1.
+ * @return {ElementArrayFinder} finder representing sliced elements
+ */
+ElementArrayFinder.prototype.slice = function (sliceStart,sliceStop,sliceStep) {
+  var self = this;
+
+  var getWebElements = function () {
```
sjelin: It might be easier to implement this on top of [ElementArrayFinder.prototype.filter](http://www.protractortest.org/#/api?view=ElementArrayFinder.prototype.filter)

sjelin:
You're gonna need at least 1 test.  Like the feature though!

lib/element.js
akarmarkar: The slicing syntax is modeled after the slice syntax in python. It differs from Array.prototype.slice in the sense that it accepts an additional step parameter. elems.slice(2,-1) is valid and should produce the result as shown below : 

 * @view
 * <ul class="items">
 *   <li>Zeroth</li>
 *   <li>First</li>
 *   <li>Second</li>
 *   <li>Third</li>
 *   <li>Fourth</li>
 *   <li>Fifth</li>
 *   <li>Sixth</li>
 *   <li>Seventh</li>
 * </ul>
 *
expect(element.all(locator).slice(2,-1).getText()).toEqual( ['Second','Third','Fourth','Fifth','Sixth']);

I am wondering if I should model this differently or stick to my current implementation and perhaps rename the method so as to not confuse it with Array.prototype.slice ?

lib/element.js
akarmarkar: Thank you for your review comments. One use case for the step is as follows : 
Let us say I have a plain text grid with 5 columns. 
Name  Age  DOB        Gender Status 
ABC    17     1/1/2011  M          Active
ABC1   10     1/6/2001  M          Inactive
ABC2   83     7/1/1931  M          Active
To extract out all elements for a particular column, I should be able to use slice with a start offset of the column number and step size of 5. So, to get all ages, I can call slice(1,undefined,5) assuming I can locate the parent element of the grid.

I hope this example is clear. I'm using protractor to automate the tests for my angular app and this is very handy to extract out column data from angular grids. There are other use cases as well.

lib/element.js
akarmarkar: fixed.

lib/element.js
akarmarkar: fixed.

lib/element.js
akarmarkar: fixed.

lib/element.js
akarmarkar: fixed.

lib/element.js
akarmarkar: fixed.

lib/element.js
akarmarkar: fixed.

lib/element.js
akarmarkar: fixed.

lib/element.js
NickTomlin: I think it makes more sense to keep things in line with how `Array.prototype.slice` works in JavaScript or give this a different name.

sjelin: It should match `Array.prototype.slice`, since Protractor is a javascript tool and not a python tool.

Also, the comment implies that `elems.slice(2,-1)` would be invalid, since `2 > -1`.  Just some rewording would be good.

lib/element.js
sjelin: I see.  Not sure how I feel about this extra feature.  @juliemr, thoughts?

lib/element.js
sjelin: Great!  Please push the updated version to your github for review

lib/element.js
```diff
@@ -93,71 +93,302 @@ ElementArrayFinder.prototype.clone = function() {
...
-      });
-    }
+      var current = start;
+      while (compare(current, stop)) {
+        list.push(parentWebElements[current]);
+        current += step;
+      }
+
+      return list;
+    });
```
sjelin: It looks like this got duplicated?

lib/element.js
```diff
@@ -93,71 +93,302 @@ ElementArrayFinder.prototype.clone = function() {
...
+ * </ul>
  *
  * @example
- * var foo = element.all(by.css('.parent')).all(by.css('.foo'))
- * expect(foo.getText()).toEqual(['1a', '2a'])
- * var baz = element.all(by.css('.parent')).all(by.css('.baz'))
- * expect(baz.getText()).toEqual(['1b'])
- * var nonexistent = element.all(by.css('.parent')).all(by.css('.NONEXISTENT'))
- * expect(nonexistent.getText()).toEqual([''])
+ * var list = element.all(by.css('.items li'));
```
NickTomlin: This example variable doesn't seem to be referenced again?

PR #2834, opened 2016-01-13
chore(testapp): section off angular 1 testapp

sjelin:
Doing this in a separate commit/pull request because it was very non-trivial


juliemr:
Can't we change the config `baseUrl` instead of changing this everywhere? Say change `spec/environment.js` so that it exposes:

```js
ng1BaseUrl:   baseUrl:
    'http://' + (process.env.HTTP_HOST || 'localhost') +
          ':' + (process.env.HTTP_PORT || webServerDefaultPort) + '/ng1/'
```
?

sjelin:
Ok, updated

juliemr:
Changes look good. Can you verify that it passes `npm test` as well as passing Travis? Then go ahead and merge.

sjelin:
Did verify - that's how I found the other error! :stuck_out_tongue:

PR #2838, opened 2016-01-14
fix(config parser): lodash removal had some errors

sjelin:


juliemr:
looks sane, thanks!

juliemr:
Why didn't these show up in a test failure? Can we add something to our CI suite?

sjelin:
We don't run all the tests in CI.  I guess we don't hit the suite feature at all?

PR #2848, opened 2016-01-18
docs(style-guide): Add protractor style guide

andresdominguez:
docs(style-guide): Add protractor style guide

Added the protractor style guide markdown file

### DIFF
--- a/docs/style-guide.md
+++ b/docs/style-guide.md
@@ -287,6 +287,15 @@ var personName = element(by.model('person.name'));
 * Both are very performant and readable locators
 * Access elements easier
 
+### Avoid text locators for text that changes frequently
+
+* Try to avoid text-based locators such as `by.linkText`, `by.buttonText`,
+  `by.cssContainingText`.
+
+**Why?**
+* Text for buttons, links, and labels tends to change over time
+* Your tests should not break when you make minor text changes
+
 # Page objects
 
 Page Objects help you write cleaner tests by encapsulating information about
@@ -560,12 +569,3 @@ describe('protractor website', function() {
 **Why?**
 * When you have a large team and multiple e2e tests people tend to write
   their own custom locators for the same directives.
-
-### Avoid text locators for text that changes frequently
-
-* Try to avoid text-based locators such as `by.linkText`, `by.buttonText`,
-  `by.cssContainingText`.
-
-**Why?**
-* Text for buttons, links, and labels tends to change over time
-* Your tests should not break when you make minor text changes
--- a/docs/style-guide.md
+++ b/docs/style-guide.md
@@ -66,13 +66,6 @@ close to the 'real deal' as possible.
 * It is supported by Protractor out of the box
 * You can use `beforeAll` and `afterAll`
 
-### Make your tests independent at file level
-
-**Why?**
-* You can run tests in parallel with sharding
-* The execution order is not guaranteed
-* You can run suites in isolation
-
 ### Make your tests independent from each other
 
 This rule holds true unless the operations performed to initialize the state of 
@@ -150,7 +143,9 @@ describe('when the user Teddy B is created', function(){
 ```
 
 **Why?**
-* You can run each single test in isolation
+* You can run tests in parallel with sharding
+* The execution order is not guaranteed
+* You can run suites in isolation
 * You can debug your tests (ddescribe/fdescribe/xdescribe/iit/fit/xit)
 
 ### Navigate to the page under test before each test
--- a/docs/style-guide.md
+++ b/docs/style-guide.md
@@ -1,5 +1,5 @@
-Style Guide
-===========
+Protractor style guide
+======================
 
 This style guide was originally created by
 [Carmen Popoviciu](https://github.com/CarmenPopoviciu) and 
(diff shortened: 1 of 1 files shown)


docs/style-guide.md
```diff
@@ -0,0 +1,612 @@
...
+ exclude: [],
+ multiCapabilities: [{
+   browserName: 'chrome',
+   shardTestFiles: true,
+   maxInstances: 3
+ }],
+ allScriptsTimeout: 11000,
+ getPageTimeout: 10000,
+ jasmineNodeOpts: {
+   isVerbose: false,
```
juliemr: Most of the JasmineNodeOpts don't matter with Jasmine2, so let's cut out `isVerbose` and `includeStackTrace`.

docs/style-guide.md
```diff
@@ -0,0 +1,612 @@
...
+
+```javascript
+ /* Recommended */
+
+ /* protractor.conf.js */
+ exclude: [],
+ multiCapabilities: [{
+   browserName: 'chrome',
+   shardTestFiles: true,
+   maxInstances: 3
```
juliemr: I'm wary of suggesting sharding before tests get too big and require it, since it can add some confusion, so I would suggest not including it in here without some explanation.

docs/style-guide.md
```diff
@@ -0,0 +1,612 @@
...
+### NEVER use xpath
+
+**Why?**
+* It's the slowest and most brittle locator strategy of all
+* Markup is very easily subject to change and therefore xpath locators require
+  a lot of maintenance
+* xpath expressions are unreadable and very hard to debug
+
+```javascript
+var elem = element(by.xpath('/*/p[2]/b[2]/following-sibling::node()' +
```
juliemr: Let's add an `/* avoid */` comment here just to be super clear.

docs/style-guide.md
```diff
@@ -0,0 +1,612 @@
...
+        .toBe('http://www.protractortest.org/#/api');
+  });
+});
+```
+
+**Why?**
+* When you have a large team and multiple e2e tests people tend to write
+  their own custom locators for the same directives.
+
+### Avoid text locators for text that changes frequently
```
juliemr: This one is a duplicate of above, let's remove it from here.

docs/style-guide.md
```diff
@@ -0,0 +1,612 @@
...
+### Use Jasmine2
+
+**Why?**
+* Jasmine is well documented
+* It is supported by Protractor out of the box
+* You can use `beforeAll` and `afterAll`
+
+### Make your tests independent at file level
+
+**Why?**
```
juliemr: This why section should go with the duplicate heading above.

docs/style-guide.md
```diff
@@ -0,0 +1,612 @@
...
+
+# Generic rules
+
+### Don't e2e test what’s been unit tested
+
+**Why?**
+* Unit tests are much faster than e2e tests
+* Avoid duplicate tests
+
+### Use one configuration file
```
juliemr: I'm not 100% convinced for this one. I think many setups work fine with multiple config files and sharing code between them via node exports... maybe this could instead just say "share common code between configuration files using node imports"?

docs/style-guide.md
```diff
@@ -0,0 +1,612 @@
...
+### Avoid text locators for text that changes frequently
+
+* Try to avoid text-based locators such as `by.linkText`, `by.buttonText`,
+  `by.cssContainingText`.
+
+**Why?**
+* Text for buttons, links, and labels tends to change over time. Minor text
+  changes in your application should not break your tests.
+
+# Test suites
```
juliemr: I think this section has some of the most important advice to avoid the most common bad tests. Maybe we can move the whole section up to emphasize its importance? What do you think?

juliemr:
Thanks for getting this together! I'd be happy to make the edits I note above or let you handle it, just let me know.

docs/style-guide.md
andresdominguez: I'm dropping it

docs/style-guide.md
andresdominguez: removed this rule

docs/style-guide.md
andresdominguez: removed this rule

docs/style-guide.md
andresdominguez: done

docs/style-guide.md
andresdominguez: done

docs/style-guide.md
andresdominguez: moving it to the top

docs/style-guide.md
andresdominguez: I didn't understand this comment. I will ask you tomorrow

juliemr: There's "Make your tests independent at least at the file level" above, so we should consolidate these two rules.

docs/style-guide.md
juliemr: Should this go up in the 'Locator Strategies' section?

docs/style-guide.md
andresdominguez: done

docs/style-guide.md
andresdominguez: done

juliemr:
Thanks! Merged as dd4ef900867ab72a5e58b2f34fdeb3fda94788cc

PR #2857, opened 2016-01-21
Fix ElementArrayFinder#filter example docs

chrismikehogan:
#filter returns an ElementArrayFinder, not a promise. The original example was inconsistent with this, and made it look like #filter return a promise.


chrismikehogan:
I signed it!

juliemr:
Thanks! Merged as 05fd60d48d128b04f86856b10c9f37bd1842f10f

Issue #2605, opened 2015-10-13
When using seleniumServerJar option to run server, server runs on explicit ip address instead of localhost


cdanielsen:
(Using protractor version 1.6.1)

When I launch a standalone selenium server on my local machine with webdriver-manager, it does so at the address `http://localhost:4444/wd/hub` and my tests run fine. If I elect to have protractor launch and teardown the server by specifying the seleniumServerJar property in my config file, my tests timeout:
```
Error: Timed out waiting for the WebDriver server at http://<explicit ip address>:4443/wd/hub
    at Error (native)
    at onResponse (C:\Code\e2e-test-framework\node_modules\protractor\node_modules\selenium-webdriver\http\util.js:87:66)
    at C:\Code\e2e-test-framework\node_modules\protractor\node_modules\selenium-webdriver\http\util.js:42:21
    at C:\Code\e2e-test-framework\node_modules\protractor\node_modules\selenium-webdriver\lib\webdriver\http\http.js:96:5
    at ClientRequest.<anonymous> (C:\Code\e2e-test-framework\node_modules\protractor\node_modules\selenium-webdriver\http\index.js:145:7)
    at ClientRequest.emit (events.js:107:17)
    at Socket.socketErrorListener (_http_client.js:271:9)
    at Socket.emit (events.js:107:17)
    at net.js:459:14
    at process._tickCallback (node.js:355:11)
```
I believe this timeout is due to firewall restrictions in my local setup, which I could investigate, but I'm wondering why protractor doesn't also use localhost as the selenium server address? Are there good reasons for the discrepancy?

sjelin:
We use [an API](https://github.com/angular/protractor/blob/6ebc4c3f8b557a56e53e0a1622d1b44b59f5bc04/lib/driverProviders/local.js#L80) made my the [selenium team](https://github.com/SeleniumHQ/selenium), so you'd have to ask them.

cdanielsen:
Fair enough; I guess I'm just confused as to why having protractor run the server versus running it manually with webdriver-manager results in different host addresses. Do they not do the same thing/use the same code?

sjelin:
[...]

PR #2860, opened 2016-01-22
Allow binding to loopback address

timroes:
Implements what was asked for in #2605 

With an additional (optional) option `seleniumUseLoopback` selenium will bind to `127.0.0.1` instead of your current network ip (e.g. `192.168.0.10`).

The default for that option is false to not introduce any changes (even though they most likely wouldn't break anything).

This is useful especially in situations where you might use some VPN like let's say Ciscos AnyConnect, which forbids connections to your computer via your current network ip, but would allow connecting to localhost/127.0.0.1. Without that we currently unfortunately cannot run End2End tests via protractor when working remote.

Unfortunately I wasn't able to run the tests (`npm start` and `npm test`) without nearly all of them failing. Perhaps if this flag should have a test, someone would be so kind and write one, or give me some additional help, why my tests might not be working (and where the best place for that test would be).

### DIFF
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -38,6 +38,10 @@ exports.config = {
   // seleniumArgs: ['-browserTimeout=60']
   // Ignored if seleniumServerJar is null.
   seleniumArgs: [],
+  // If you set this option to true, the started server will listen on 127.0.0.1.
+  // If it's set to false or not set at all it will listen to the ip address in your
+  // local network (e.g. 192.168.0.10).
+  seleniumUseLoopback: false,
   // ChromeDriver location is used to help find the chromedriver binary.
   // This will be passed to the Selenium jar as the system property
   // webdriver.chrome.driver. If null, Selenium will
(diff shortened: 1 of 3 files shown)


timroes:
I signed it!

juliemr:
After taking another look at `remote.SeleniumServer`, what if instead we add a `localSeleniumStandaloneOpts` object to the config, and just pass that directly to the `SeleniumServer` constructor? We can set `config.localSeleniumStandaloneOpts.port = config.seleniumPort` and `config.localSeleniumStandaloneOpts.args = config.seleniumArgs` for backwards compatibility. Then you can use `config.localSeleniumStandaloneOpts.loopback` to do this.

timroes:
@juliemr Sounds like a good idea. I just added changes to enable this.

juliemr:
Thanks! Merged as b6932560d66730203e0e7b0c65c80a44ad4747de

I added a couple more docs with 261b49e1129e0288215d865d5953872c4839877a

Issue #2345, opened 2015-07-14
Protractor and https proxy


fgather:
Hi,

we have to use protractor with browserstack behind a https proxy and until now a solution like this worked:
```js
process.env.http_proxy = 'https://' + config.proxy.host + ':' + config.proxy.port;
var globalTunnel = require ('global-tunnel');
globalTunnel.initialize ();
```
Unfortunately, when doing this, the communication with the proxy itself is via http, while the communication from the proxy to browserstack is via https. The proxy was recently configured to prohibit incoming http requests. We then tried to add 
```js
proxy = {
            proxyType: 'manual',
            sslProxy: config.proxy.host + ':' + config.proxy.port
        }
```
to our browser capabilities, but the proxy configuration is not forwarded towards webdriver. Furthermore, adding a new proxy config variable and adding
```js
usingWebDriverProxy(this.config_.proxy)
```
to the getNewDriver in driverProvider.js also doesn't work, and it still uses http requests. Is this a limitation of the webdriver? Or could it be fixed from within protractor?

Many Regards,
Florian

fgather:
Hi, 

think I solved the problem. It is an issue with globaltunnel.js. 

However, the change in driverProvider.js, described above, still has to be done to support webdriver's ability for proxies. Are there any plans to integrate this feature?

sjelin:
@hankduan weren't you doing something with proxies?

fgather:
#2872 is merged, so this should be closed.

PR #2872, opened 2016-01-26
feat(webdriver): add support for selenium webdriver proxy

vliloia:
Allows the configuration of a proxy by using selenium webdriver `Builder#usingWebDriverProxy()` (was added with [v2.46.0](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/CHANGES.md#v2460)).

This is useful when you sit behind a corporate proxy and you need to connect to a remote webdriver server (e.g. BrowserStack).

There is also an issue related to it #2345.

### DIFF
--- a/lib/driverProviders/driverProvider.js
+++ b/lib/driverProviders/driverProvider.js
@@ -34,6 +34,7 @@ DriverProvider.prototype.getExistingDrivers = function() {
 DriverProvider.prototype.getNewDriver = function() {
   var builder = new webdriver.Builder().
     usingServer(this.config_.seleniumAddress).
+    usingWebDriverProxy(this.config_.seleniumProxy).
     withCapabilities(this.config_.capabilities);
   if (this.config_.disableEnvironmentOverrides === true) {
     builder.disableEnvironmentOverrides();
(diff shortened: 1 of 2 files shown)


vliloia:
`I signed it!`

lib/driverProviders/driverProvider.js
```diff
@@ -34,6 +34,7 @@ DriverProvider.prototype.getExistingDrivers = function() {
 DriverProvider.prototype.getNewDriver = function() {
   var builder = new webdriver.Builder().
     usingServer(this.config_.seleniumAddress).
+    usingWebDriverProxy(this.config_.seleniumProxy).
```
juliemr: Can you add a note to https://github.com/angular/protractor/blob/master/docs/referenceConf.js explaining what this config option does?

juliemr: Could we name it `webDriverProxy`? Unless there's a reason to have a discrepancy between the function name and the config name.

vliloia:
Thanks for reviewing. Added changes

juliemr:
Thanks! Merged as aa5ceb576f0283b6591faaa95e342ef3c603c717

PR #2874, opened 2016-01-27
feat(attachSession): attach protractor to existing webdriver session

cnishina:
Attaching an existing selenium browser session to protractor rather than always creating new one. The session can be passed into the config file as a string via the `sessionId`.

### DIFF
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -55,6 +55,10 @@ exports.config = {
   // connect to an already running instance of Selenium. This usually looks like
   // seleniumAddress: 'http://localhost:4444/wd/hub'
   seleniumAddress: null,
+  // The selenium session id allows Protractor to attach to an existing selenium
+  // browser session. The selenium session is maintained after the test has
+  // completed. Ignored if seleniumAddress is null.
+  seleniumSessionId: null,
 
   // ---- 3. To use remote browsers via Sauce Labs -----------------------------
   // If sauceUser and sauceKey are specified, seleniumServerJar will be ignored.
--- a/lib/configParser.js
+++ b/lib/configParser.js
@@ -32,6 +32,8 @@ var ConfigParser = function() {
       stackFilter: helper.filterStackTrace,
       defaultTimeoutInterval: (30 * 1000)
     },
+    seleniumArgs: [],
+    seleniumSessionId: null,
     mochaOpts: {
       ui: 'bdd',
       reporter: 'list'
--- a/lib/driverProviders/attachSession.js
+++ b/lib/driverProviders/attachSession.js
@@ -1,0 +1,58 @@
+/*
+ *  This is an implementation of the Attach Session Driver Provider.
+ *  It is responsible for setting up the account object, tearing
+ *  it down, and setting up the driver correctly.
+ */
+
+var util = require('util'),
+    q = require('q'),
+    DriverProvider = require('./driverProvider'),
+    log = require('../logger'),
+    webdriver = require('selenium-webdriver'),
+    executors = require('selenium-webdriver/executors');
+
+var AttachedSessionDriverProvider = function(config) {
+  DriverProvider.call(this, config);
+};
+util.inherits(AttachedSessionDriverProvider, DriverProvider);
+
+/**
+ * Configure and launch (if applicable) the object's environment.
+ * @public
+ * @return {q.promise} A promise which will resolve when the environment is
+ *     ready to test.
+ */
+AttachedSessionDriverProvider.prototype.setupEnv = function() {
+  log.puts('Using the selenium server at ' + this.config_.seleniumAddress);
+  log.puts('Using session id - ' + this.config_.seleniumSessionId);
+  return q(undefined);
+};
+
+
+/**
+ * Getting a new driver by attaching an existing session.
+ *
+ * @public
+ * @return {webdriver.WebDriver} webdriver instance
+ */
+AttachedSessionDriverProvider.prototype.getNewDriver = function() {
+  var executor = executors.createExecutor(this.config_.seleniumAddress);
+  var newDriver;
+  newDriver = new webdriver.WebDriver
+      .attachToSession(executor, this.config_.seleniumSessionId);
+  this.drivers_.push(newDriver);
+  return newDriver;
+};
+
+/**
+ * Maintains the existing session and does not quit the driver.
+ *
+ * @public
+ */
+AttachedSessionDriverProvider.prototype.quitDriver = function() {
+};
+
+// new instance w/ each include
+module.exports = function(config) {
+  return new AttachedSessionDriverProvider(config);
+};
--- a/scripts/attachSession.js
+++ b/scripts/attachSession.js
@@ -1,0 +1,105 @@
+#!/usr/bin/env node
+
+'use strict';
+
+var http = require('http'),
+    spawn = require('child_process').spawnSync;
+
+var sessionId = '';
+
+// 1. Create a new selenium session.
+var postData = JSON.stringify(
+  {'desiredCapabilities': {'browserName': 'firefox'}});
+var createOptions = {
+  hostname: 'localhost',
+  port: 4444,
+  path: '/wd/hub/session',
+  method: 'POST',
+  headers: {
+    'Content-Type': 'application/x-www-form-urlencoded',
+    'Content-Length': Buffer.byteLength(postData)
+  }
+};
+var req = http.request(createOptions, function(res) {
+  res.on('data', setBody);
+  res.on('end', checkSession);
+});
+req.write(postData);
+req.end();
+
+// 2. After making the request to create a selenium session, read the selenium
+// session id.
+var setBody = function(chunk) {
+  var body = chunk.toString();
+  sessionId = JSON.parse(body).sessionId;
+};
+
+// 3. After getting the session id, verify that the selenium session exists.
+// If the session exists, run the protractor test.
+var checkSession = function() {
+  var checkOptions = {
+    hostname: 'localhost',
+    port: 4444,
+    path: '/wd/hub/session/' + sessionId,
+    method: 'GET'
+  };
+  var state = '';
+  var req = http.request(checkOptions, function(res) {
+    res.on('data', function(chunk) {
+      state = JSON.parse(chunk.toString()).state;
+    });
+    res.on('end', function() {
+      if (state === 'success') {
+        var runProtractor = spawn('bin/protractor',
+            ['spec/attachSession.js', '--seleniumSessionId=' + sessionId]);
+        console.log(runProtractor.stdout.toString());
+        if (runProtractor.status !== 0) {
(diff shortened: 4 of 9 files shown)


lib/configParser.js
```diff
@@ -33,6 +33,7 @@ var ConfigParser = function() {
       defaultTimeoutInterval: (30 * 1000)
     },
     seleniumArgs: [],
+    sessionId: null,
```
juliemr: Can you also update our master list of configuration docs at https://github.com/angular/protractor/blob/master/docs/referenceConf.js?

juliemr: Let's rename to `seleniumSessionId` - to err on the side of being too verbose but obvious what this belongs to.

lib/driverProviders/attachSession.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+AttachedSessionDriverProvider.prototype.setupEnv = function() {
+  log.puts('Using the selenium server at ' + this.config_.seleniumAddress);
+  log.puts('Using session id - ' + this.config_.sessionId);
+  return q.fcall(function() {});
+};
+
+
+/**
+ * Getting a new driver by attaching an existing session
```
juliemr: Nit: always end comments with a '.'

lib/driverProviders/attachSession.js
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @public
+ * @return webdriver instance
+ */
+AttachedSessionDriverProvider.prototype.getNewDriver = function() {
+  var executor = executors.createExecutor(this.config_.seleniumAddress, null);
```
juliemr: Nit: Looks like we can just do `executors.createExecutor(this.config_.seleniumAddress)`, no need for the last null parameter.

lib/driverProviders/attachSession.js
```diff
@@ -0,0 +1,53 @@
...
+  var newDriver = new webdriver.WebDriver
+      .attachToSession(executor, this.config_.sessionId);
```
juliemr: What's the error message look like if the session id is invalid? Is it reasonably actionable for a user, or do we need to wrap it?

scripts/attachSessionHelper.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+};
+var TEST = {};
+TEST.body = '';
+
+var setBody = function(chunk) {
+  TEST.body += chunk.toString();
+  var sessionId = JSON.parse(TEST.body).sessionId;
+  console.log('created new session with id: ' + sessionId);
+  fs.writeFile('spec/attachSessionTemp.js',
```
juliemr: I think it would be cleaner to use Node's [child_process](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback) to start up protractor and add a command line flag, e.g. `bin/protractor spec/attachSessionConf.js --sessionId=***`

Then we don't have to mess with the file system.

juliemr: And, if we use child process, we can intelligently wait for protractor to be done instead of using an arbitrary timeout.

lib/driverProviders/attachSession.js
```diff
@@ -0,0 +1,53 @@
...
+  this.drivers_.push(newDriver);
+  return newDriver;
+
+};
+
+
```
juliemr: I think we'll need to override `quitDriver` as well to avoid killing the session when we're done with it.

scripts/test.js
```diff
@@ -26,16 +26,17 @@ var passingTests = [
   'node lib/cli.js spec/plugins/browserGetUnsyncedConf.js',
   'node lib/cli.js spec/plugins/waitForAngularConf.js',
   'node lib/cli.js spec/interactionConf.js',
-  'node lib/cli.js spec/directConnectConf.js',
+  //'node lib/cli.js spec/directConnectConf.js',
```
juliemr: These are working on my machine - lets leave them in, and we can figure out why your machine doesn't like them (or maybe we're missing some setup instructions that I've forgotten about).

scripts/attachSessionHelper.js
```diff
@@ -0,0 +1,42 @@
...
+    'var env = require(\'./environment.js\'); \n' +
+    '\n' +
+    'exports.config = { \n' +
+    '  seleniumAddress: env.seleniumAddress, \n' +
+    '  sessionId: \'' + sessionId + '\', \n' +
+    '  framework: \'jasmine\', \n' +
+    '  specs: [\'attachSessionProvider/*_spec.js\'], \n' +
+    '  baseUrl: \'http://localhost:8081\' \n' +
+    '};\n');
+  setTimeout(function(){}, 3000);
```
juliemr: We should send a command to quit the session after protractor finishes, as well.

spec/attachSessionProvider/attachSession_spec.js
```diff
@@ -0,0 +1,11 @@
+describe('attach session connect', function() {
+  var URL = '/ng2/#/async';
+
+  beforeEach(function() {
+    browser.get(URL);
+  });
+  it('create a new session', function() {
```
juliemr: Nit: `creates` - specs read like a sentence; "attach session connect creates a new session" :)

juliemr:
Added some review comments!

scripts/test.js
cnishina: This didn't run for me because it is using chromedriver directly. My error is: Could not find chromedriver at protractor/selenium/chromedriver

lib/driverProviders/attachSession.js
```diff
@@ -0,0 +1,53 @@
...
+var AttachedSessionDriverProvider = function(config) {
+  DriverProvider.call(this, config);
+};
+util.inherits(AttachedSessionDriverProvider, DriverProvider);
+
+/**
+ * Configure and launch (if applicable) the object's environment.
+ * @public
+ * @return {q.promise} A promise which will resolve when the environment is
```
sjelin: nit: q.promise<function>

lib/driverProviders/attachSession.js
```diff
@@ -0,0 +1,53 @@
...
+ *     ready to test.
+ */
+AttachedSessionDriverProvider.prototype.setupEnv = function() {
+  log.puts('Using the selenium server at ' + this.config_.seleniumAddress);
+  log.puts('Using session id - ' + this.config_.sessionId);
+  return q.fcall(function() {});
```
sjelin: Why not `q(undefined)`?

lib/driverProviders/attachSession.js
```diff
@@ -0,0 +1,53 @@
...
+};
+
+
+/**
+ * Getting a new driver by attaching an existing session
+ *
+ * @public
+ * @return webdriver instance
```
sjelin: Needs type `{webdriver.WebDriver}`

scripts/attachSession.sh
```diff
@@ -0,0 +1,2 @@
+node scripts/attachSessionHelper.js
```
sjelin: I think we should put this & attachSessionHelper in a subfolder

lib/driverProviders/attachSession.js
cnishina: The error message when the session id is invalid returns an `UnkonwnError`. The error is being caused in the selenium/webdriver.js and I cannot catch the error.

lib/driverProviders/attachSession.js
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Configure and launch (if applicable) the object's environment.
+ * @public
+ * @return {q.promise} A promise which will resolve when the environment is
+ *     ready to test.
+ */
+AttachedSessionDriverProvider.prototype.setupEnv = function() {
+  log.puts('Using the selenium server at ' + this.config_.seleniumAddress);
+  log.puts('Using session id - ' + this.config_.seleniumSessionId);
+  return q.fcall(function() {});
```
sjelin: Why not `q(undefined)`?

cnishina: I could change that to `q(undefined)`. I was basing the `setupEnv` off of hosted.js.

scripts/attachSession.sh
cnishina: Sure. What folder should I add them to?

cnishina:
See comments above. Thanks!

lib/driverProviders/attachSession.js
sjelin: No I mean why not `q(undefined)` instead of `q.fcall(function() {})`?

spec/.gitignore
```diff
@@ -0,0 +1 @@
+attachSessionTemp.js
```
sjelin: Is this change still necessary?

docs/referenceConf.js
```diff
@@ -49,6 +49,10 @@ exports.config = {
   // connect to an already running instance of Selenium. This usually looks like
   // seleniumAddress: 'http://localhost:4444/wd/hub'
   seleniumAddress: null,
+  // The selenium session id allows Protractor to attach to an existing selenium
+  // browser session. The selenium session is maintained after the test has
+  // completed. Ignored if seleniumSessionId is null.
```
juliemr: I think this should be "ignored if seleniumAddress is null"

scripts/attachSession.js
```diff
@@ -0,0 +1,101 @@
...
+  };
+  var state = '';
+  var req = http.request(checkOptions, function(res) {
+    res.on('data', function(chunk) {
+      state = JSON.parse(chunk.toString()).state;
+    });
+    res.on('end', function() {
+      if (state === 'success') {
+        var runProtractor = spawn('bin/protractor',
+            ['spec/attachSession.js', '--seleniumSessionId=' + sessionId]);
```
juliemr: Can we check on the `ChildProcess#close` event and check the exit value of the Protractor process? Then we can also ensure that protractor finishes before stopping the session.

scripts/test.js
```diff
@@ -32,10 +32,11 @@ var passingTests = [
   'node lib/cli.js spec/controlLockConf.js',
   'node lib/cli.js spec/customFramework.js',
   'node lib/cli.js spec/angular2Conf.js',
-  'node scripts/interactive_tests/interactive_test.js',
-  'node scripts/interactive_tests/with_base_url.js',
+  'node scripts/attachSession.js',
+  //'node scripts/interactive_tests/interactive_test.js',
```
juliemr: Make sure to revert commenting-out in this file before submitting.

docs/referenceConf.js
```diff
@@ -49,6 +49,10 @@ exports.config = {
...
```
juliemr: Github ate my comment :(

This should be "Ignored if seleniumAddress is null", right?

juliemr:
LGTM after the two cleanup comments

docs/referenceConf.js
cnishina: Yes, changing it.

cnishina:
Closed with 3f3805f.

PR #2889, opened 2016-01-30
feat(element): Add $x and $$x shortcuts for xpath selectors

### LINKED ISSUE #2855: Add $x shortcut for element(by.xpath())
Similarly to the `$` which is used for `element(by.css())` and `$$` for `element.all(by.css())`, please consider adding shortcuts for the `by.xpath()` locators (similar to what we have on the Chrome Dev Tools console):

    $x -> element(by.xpath())
    $$x -> element.all(by.xpath())

Thank you.

NickTomlin:
(Closes #2855)

I personally don't use a lot of xpath selectors in the tests, and it advised against in the [styleguide](https://github.com/angular/protractor/blob/master/docs/style-guide.md#never-use-xpath), but I can see this coming in handy for those that do. The change itself is very simplistic (it mimics the behavior of `$` and `$$` exactly) so I don't see this as harmful.

Please feel free to reject this if it is going to add too much bloat or encouraging non ideal usage of xpath; we can then close the corresponding issue.

@alexce i'd appreciate another set of eyes on this as well since it was your idea :smiley_cat:

### DIFF
--- a/lib/element.js
+++ b/lib/element.js
@@ -308,6 +308,15 @@ ElementArrayFinder.prototype.$$ = function(selector) {
 };
 
 /**
+ * Shorthand function for finding arrays of elements by XPath.
+ *
+ * @type {function(string): ElementArrayFinder}
+ */
+ElementArrayFinder.prototype.$$x = function(selector) {
+  return this.all(webdriver.By.xpath(selector));
+};
+
+/**
  * Returns an ElementFinder representation of ElementArrayFinder. It ensures
  * that the ElementArrayFinder resolves to one and only one underlying element.
  *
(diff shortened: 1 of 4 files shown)


lib/element.js
```diff
@@ -1012,3 +1073,62 @@ var build$$ = function(element, by) {
...
+ *   <span class="two">Second</span>
+ * </div>
+ *
+ * @example
+ * var item = $x('//span[@id="two"]');
+ * expect(item.getText()).toBe('Second');
+ *
+ * @param {string} selector An XPath selector
+ * @return {ElementFinder} which identifies the located
+ *     {@link webdriver.WebElement}
```
sjelin: nit: period at end of comment

lib/element.js
```diff
@@ -1012,3 +1073,62 @@ var build$$ = function(element, by) {
   };
 };
 exports.build$$ = build$$;
+
+/**
+ * Shortcut for querying the document directly with XPath.
+ * `element(by.xpath('.abc'))` is equivalent to `$('.abc')`
```
sjelin: nit: period at end of comment

NickTomlin: Done thanks!

juliemr:
I'm conflicted about this one. It's an easy change to make and people have asked for it. But, there's two downsides I'm concerned about. First, it adds to our globals - we already assume that we're in a test environment and it's OK for Protractor to add globals, but I'd still like to set a high bar for what gets done. (That said, $$x is unlikely to conflict with anything). Second, we shouldn't be encouraging xpath, as [Andres and Carmen point out here](https://github.com/angular/protractor/blob/master/docs/style-guide.md#never-use-xpath).

I could probably be convinced either way. Thoughts?

sjelin:
I'm less concerned about polluting the global namespace, since these are weird names.  That said I opened up https://github.com/angular/protractor/issues/2893 which is a feature request for an option that avoids putting so much under the global namespace.

Not sure what do say about not wanting to encourage people to use xpath other than that people are using it anyway.

NickTomlin:
@sjelin good idea on a no globals option!

@juliemr I agree with Sammy in that this is probably not going to lure people into XPath; the global footprint is pretty safe. Users who _do_ use XPath get a nice bonus to convenience at relatively little cost to the project.

That said, I don't feel strongly about this feature (since I don't use XPath much myself) so perhaps it is best to go ahead and close this if there isn't a stronger champion for the feature.

kendall-powerchord:
Please reconsider this feature.  It sounded like it would be easy to add.  I don't want to get into a "tabs versus spaces" conversation, but the fact of the matter is that people do use xpath because you can do css and id searching inside of it as well as all the capabilities and functions that xpath provides.  I'll argue that you should lay personal preference aside, and make this change if it is relatively easy to do so for the benefit of everyone out there using xpath, and there are a lot of us.

rakeshnambiar:
@NickTomlin to double check after goig through all the comments, the `$x` feature is not implemented yet right? I cannot see it on the version `5.4.2`.

PR #2890, opened 2016-01-30
feat(config): allow onComplete to return a promise

### LINKED ISSUE #1944: onComplete does not allow for a blocking callback
I'm trying to implement an asynchronous action (_Nodemailer_) during the `onComplete` of my test config. I found that while the action starts, onComplete would finish and kill the runner before the action completed. 

In order to get `onComplete` working for me using a deferred promise, I needed to modify the `lib/framework/jasmine.js` as such:
```
jasmineNodeOpts.onComplete = function(jasmineRunner, log) {
    try {
        if (originalOnComplete) {
            originalOnComplete(jasmineRunner, log).then(resolveOnComplete);
        }else{
            resolveOnComplete();
        }
		
        function resolveOnComplete(){
            resolve({
                failedCount: jasmineRunner.results().failedCount,
                specResults: testResult
            });
        }
			
    } catch (err) {
        reject(err);
    }
};
```

sjelin:
Closes #1944

### DIFF
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -248,6 +248,8 @@ exports.config = {
   },
 
   // A callback function called once tests are finished.
+  // onComplete can optionally return a promise, which Protractor will wait for
+  // before shutting down webdriver.
   onComplete: function() {
     // At this point, tests will be done but global objects will still be
     // available.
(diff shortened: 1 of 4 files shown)


docs/referenceConf.js
```diff
@@ -248,6 +248,7 @@ exports.config = {
   },
 
   // A callback function called once tests are finished.
+  // onComplete can optionally return a promise, which Protractor will wait for.
```
juliemr: Maybe add "which Protractor will wait for before shutting down webdriver"?

juliemr:
I'm thinking that our runners (jasmine and mocha) are complex enough that we should have a 'frameworks' folder under our specs, which exercises them with some simple files.

But that can be work for a separate PR - this looks good to me.

Issue #1203, opened 2014-08-18
Typescript support


aleks-sidorenko:
Is there a way to write tests in typescript without compiling it manually before testing? Some kind of typescript preprocessor?
I know it is possible to write tests in coffee, is it possible to do same with typescript?

milton-rodriguez-uruit:
Hi all ! I don't know if this is the right place to write down this question, but I'm having some problems to run tests writen using Typescript in different classes.
I want to abstract some common parts of all my test to a base test class, and then extend it from my spec files. Everything compiles correctly, but it fails when executes js files, where the prototype is not defined.

Any help on this ?

elgalu:
Should be "Typescript support" or "Typescript rewrite"?
Given [angular 2.0 will be in TS](https://jaxenter.com/angular-typescript-dart-115426.html) maybe Protractor should be fully rewritten in TS ;)

AbraaoAlves:
@elgalu maybe, but for now, we can create definitions files from jsdoc. With this https://www.npmjs.com/package/typescript-closure-tools

In my typescript projects I use [tsd](definitelytyped.org/tsd/) to manage my definitions files, includes [angular-protractor](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/angular-protractor/angular-protractor.d.ts). The big problem in definition files is that they are part of project independent, although they were always synchronized, even if manually!

A tool to generate here a "protractor.d.ts" will be great help for everyone.

juliemr:
Update here - I think a typescript rewrite is probably a far-future goal, that wouldn't give us a lot of immediate benefit.

However - we should absolutely include typescript definition files in our package. That's a high priority.

ocombe:
And what about a typescript preprocessor to be able to write definitions in typescript instead of ES5 js ?
Or at least a babel preprocessor for some ES6?

mccataldo:
It looks like we'll need the following definitions to write protractor tests in typescript:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/angular-protractor/angular-protractor.d.ts
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/jasmine/jasmine.d.ts
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/node/node.d.ts
[...]

PR #2902, opened 2016-02-04
feat(typescript): adding typescript to protractor

cnishina:
Converting a 3 files over to typescript.

On the `prepublish` step, it will download the typings, transpile the files with tsc, and move the transpiled javascript files from the built directory to the lib.

Also adding scripts to package.json for `npm run tsc` and `npm run tsc:w` for transpiling.


cnishina:
The linter was checking the generated javascript files from typescript. Removing the check to lib as a "fix".

bin/protractor
```diff
@@ -2,4 +2,11 @@
 
 process.env.NODE_ENV = process.env.NODE_ENV || 'test';
 
+// Transpile code to javascript prior to running javascript
+var spawn = require('child_process').spawnSync;
+var transpile = spawn('node_modules/typescript/bin/tsc');
```
alexeagle: the strategy of compiling files in the client is very different from what we do elsewhere. why not ship compiled .js files to the client?

lib/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.d.ts
```
alexeagle: why are .d.ts files ending up here? they should not get mixed in with sources - your compiler output ought to go to a .gitignore'd folder like built/

tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+  "compilerOptions": {
+    "target": "es5",
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "sourceMap": true,
+    "declaration": true,
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+    "removeComments": false,
+    "noImplicitAny": false
```
alexeagle: add outDir: built
or similar

tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "sourceMap": true,
+    "declaration": true,
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
```
alexeagle: these two lines - do you use decorators or was this copy/paste from an angular2 app?

lib/configParser.js
```diff
@@ -1,50 +1,170 @@
...
-    chromeDriver: null,
-    configDir: './',
-    plugins: [],
-    skipSourceMapSupport: false
-  };
-};
-
+catch (e) {
+}
+var ConfigParser = (function () {
```
alexeagle: are these clang-format changes? if so it would be better in a separate commit to make this review more focused

lib/configParser.ts
```diff
@@ -0,0 +1,230 @@
+import * as path from 'path';
+import * as glob from 'glob';
```
alexeagle: nit: `import {symbol}` is preferred over `import *`

lib/configParser.ts
```diff
@@ -0,0 +1,230 @@
...
+  require('LiveScript');
+} catch (e) {
+  // Intentionally blank - ignore if LiveScript is not available.
+}
+
+export class ConfigParser {
+  public constructor() {}
+
+  // Default configuration.
+  private config_ = {
```
alexeagle: can there be an interface that defines the typeshape of valid configuration objects?

lib/configParser.ts
```diff
@@ -0,0 +1,230 @@
...
+
+// LiveScript is required here to enable config files written in LiveScript.
+try {
+  require('LiveScript');
+} catch (e) {
+  // Intentionally blank - ignore if LiveScript is not available.
+}
+
+export class ConfigParser {
+  public constructor() {}
```
alexeagle: no need to explicitly write the no-arg constructor

lib/configParser.ts
```diff
@@ -0,0 +1,230 @@
...
+  /**
+   * Resolve a list of file patterns into a list of individual file paths.
+   *
+   * @param {Array.<string> | string} patterns
+   * @param {=boolean} opt_omitWarnings Whether to omit did not match warnings
+   * @param {=string} opt_relativeTo Path to resolve patterns against
+   *
+   * @return {Array} The resolved file paths.
+   */
+  public static resolveFilePatterns(patterns, opt_omitWarnings, opt_relativeTo) {
```
alexeagle: if you add --noImplicitAny to your compiler options, you'll get typechecker errors to add types here. But you could do it later since this is an incremental migration

lib/configParser.ts
```diff
@@ -0,0 +1,230 @@
+import * as path from 'path';
```
alexeagle: it's really confusing that you'll have `configParser.ts` and `configParser.js` in the same directory, seems like the output from the compiler will collide?

lib/configParser.ts
```diff
@@ -0,0 +1,230 @@
...
+  public static resolveFilePatterns(patterns, opt_omitWarnings, opt_relativeTo) {
+    var resolvedFiles = [];
+    var cwd = opt_relativeTo || process.cwd();
+
+    patterns = (typeof patterns === 'string') ?
+        [patterns] : patterns;
+
+    if (patterns) {
+      for (var i = 0; i < patterns.length; ++i) {
+        var fileName = patterns[i];
```
alexeagle: nit: prefer `let` over `var`

lib/configParser.ts
```diff
@@ -0,0 +1,230 @@
...
+   */
+  public static resolveFilePatterns(patterns, opt_omitWarnings, opt_relativeTo) {
+    var resolvedFiles = [];
+    var cwd = opt_relativeTo || process.cwd();
+
+    patterns = (typeof patterns === 'string') ?
+        [patterns] : patterns;
+
+    if (patterns) {
+      for (var i = 0; i < patterns.length; ++i) {
```
alexeagle: `for..of` loop?

lib/configParser.ts
```diff
@@ -0,0 +1,230 @@
...
+        union(specs, makeArray(suiteList));
+      });
+      return specs;
+    }
+
+    if (config.specs.length > 0) {
+      return config.specs;
+    }
+
+    Array.prototype.forEach.call(config.suites, function(suite) {
```
alexeagle: why not `config.suites.forEach()` ?

alexeagle: many places for fat arrows, eg
```
(suite) => { union(); }

File: lib/logger.js [hunk 14]
```diff
@@ -1,3 +1,5 @@
+/// <reference path="../typings/node/node.d.ts" />
```
alexeagle: prefer to give node.d.ts to the compiler as an input, don't reference in your code

alexeagle: also this is a .js file, so the comment does nothing? unless you're using TS 1.8 support
https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#including-js-files-with---allowjs

alexeagle: oh! you are checking in generated code!! don't do that :)

tsd.json
```diff
@@ -0,0 +1,21 @@
+{
+  "version": "v4",
+  "repo": "borisyankov/DefinitelyTyped",
```
alexeagle: looks old, should be DefinitelyTyped/DefinitelyTyped

typings/glob/glob.d.ts
```diff
@@ -0,0 +1,112 @@
+// Type definitions for Glob 5.0.10
```
alexeagle: checking in the typings can be a valid thing to do, because the github connection can be flaky...
but in our other projects, we run `typings install ...` in an npm postinstall script and don't check in the files

typings/tsd.d.ts
```diff
@@ -0,0 +1,6 @@
+
```
alexeagle: opt: .gitignore the typings folder

cnishina: Done.

tsd.json
cnishina: Removing tsd.json for typings.json.

typings/glob/glob.d.ts
cnishina: Done. Added  `typings install` to post install script in package.json.

lib/configParser.ts
cnishina: * Replaced `Array.prototype.forEach` with `config.suites.forEach`.
* Went over typescript, cleaned up all with fat arrows.

lib/configParser.ts
cnishina: Done.

lib/configParser.ts
cnishina: Cleaned up all files for `let` over use of `var`

lib/configParser.ts
cnishina: Fixed all implicit any and added types.

lib/configParser.ts
cnishina: Removed.

lib/configParser.ts
cnishina: Done, created a Config interface.

tsconfig.json
cnishina: Removed.

lib/.gitignore
cnishina: Moved *.d.ts to built/ directory and added built/ to the .gitignore

tsconfig.json
cnishina: Added outDir: built

lib/.gitignore
```diff
@@ -0,0 +1,4 @@
+# TODO: remove this file when all files have been replaced with typescript
+configParser.js
```
alexeagle: do you still need this? seems like the file is now removed by this change

lib/configParser.ts
```diff
@@ -0,0 +1,258 @@
+import { resolve, dirname } from 'path';
```
alexeagle: consider running all your .ts files through clang-format so I don't need to nitpick about whitespace :)

You can see in other repos (eg. angular/ts2dart) how we've hooked up clang-format so it's convenient to run and enforced by the build.

lib/configParser.ts
```diff
@@ -0,0 +1,258 @@
...
+  // Intentionally blank - ignore if LiveScript is not available.
+}
+
+export interface Config {
+  specs: Array<string>;
+  multiCapabilities: Array<any>;
+  rootElement: string;
+  allScriptsTimeout: number;
+  getPageTimeout: number;
+  params: any;
```
alexeagle: I am surprised that none of these are marked optional. Doesn't it fail to typecheck if you create a Config object that is missing one of the properties?

lib/configParser.ts
```diff
@@ -0,0 +1,258 @@
...
+      seleniumSessionId: null,
+      mochaOpts: {
+        ui: 'bdd',
+        reporter: 'list'
+      },
+      chromeDriver: null,
+      configDir: './',
+      plugins: [],
+      skipSourceMapSupport: false,
+      suites: null,
```
alexeagle: seems like this property wants to be optional, for example

lib/configParser.ts
```diff
@@ -0,0 +1,258 @@
...
+      suite: null
+    };
+
+  }
+
+  /**
+   * Resolve a list of file patterns into a list of individual file paths.
+   *
+   * @param {Array.<string> | string} patterns
```
alexeagle: what's your long-term plan with the JSDoc type annotations? will these be kept in sync with the typescript typings? Is anything type-checking them? Does the docgen use them?

Seems to me these should just be removed in this change so later maintainers don't have to deal with them.

lib/configParser.ts
```diff
@@ -0,0 +1,258 @@
...
+   * @param {=boolean} opt_omitWarnings Whether to omit did not match warnings
+   * @param {=string} opt_relativeTo Path to resolve patterns against
+   *
+   * @return {Array} The resolved file paths.
+   */
+  public static resolveFilePatterns(patterns: Array<string>|string,
+      opt_omitWarnings: boolean, opt_relativeTo: string) {
```
alexeagle: missing the optionality (should have a `?`)

lib/configParser.ts
```diff
@@ -0,0 +1,258 @@
...
+    let resolvedFiles: Array<string> = [];
+    let cwd = opt_relativeTo || process.cwd();
+
+    patterns = (typeof patterns === 'string') ?
+        [patterns] : patterns;
+
+    if (patterns) {
+      for (let fileName of patterns) {
+        let matches = sync(fileName, {cwd: cwd});
```
alexeagle: I believe ES6 allows this to be just `{cwd}` as a shorthand for property and value being the same identifier
http://people.mozilla.org/~jorendorff/es6-draft.html#sec-object-initializer

try that out?

lib/configParser.ts
```diff
@@ -0,0 +1,258 @@
...
+  /**
+   * Public function specialized towards merging in a file's config
+   *
+   * @public
+   * @param {String} filename
+   */
+  public addFileConfig(filename: string) {
+    try {
+      if (!filename) {
+        return this;
```
alexeagle: should this method have a declared return type?

lib/driverProviders/sauce.js
```diff
@@ -5,7 +5,7 @@
  */
 
 var util = require('util'),
-    log = require('../logger.js'),
+    log = require('../logger.js').Logger,
```
alexeagle: If you like, you can avoid this change by continuing to use default export for the Logger symbol.
https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#11.3.4.2

lib/logger.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Utility functions for command line output logging from Protractor.
+ * May be used in different processes, since the launcher spawns
+ * child processes for each runner.
+ *
+ * All logging directly from Protractor, its driver providers, or runners,
+ * should go through this file so that it can be customized.
+ */
+export module Logger {
```
alexeagle: this looks like old TypeScript syntax for a "internal module" which was replaced by namespace in https://github.com/Microsoft/TypeScript/issues/2159

https://basarat.gitbooks.io/typescript/content/docs/project/namespaces.html

package.json
```diff
@@ -45,7 +47,10 @@
   },
   "main": "lib/protractor.js",
   "scripts": {
-    "pretest": "./scripts/pretest.sh",
+    "prepublish": "./node_modules/.bin/typings install && ./node_modules/typescript/bin/tsc && cp built/*.js lib/",
```
alexeagle: I think I was wrong about this. You have typings in devDependencies, so the typings install command belongs in postinstall script (otherwise your teammates won't be able to compile until they attempt to publish!)

tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+    "removeComments": false,
+    "noImplicitAny": true,
+    "outDir": "built/"
+  },
+  "exclude": [
+    "typings",
+    "node_modules",
+    "testapp"
+  ],
+  "files": [
```
alexeagle: don't include both exclude and files.

https://github.com/Microsoft/TypeScript/wiki/tsconfig.json#details says
The "files" property cannot be used in conjunction with the "exclude" property. If both are specified then the "files" property takes precedence.

files will probably get you into trouble when you have to list every single test file, so just use exclude and let the compiler find all `.ts` files under this directory.

lib/configParser.ts
juliemr: We should indeed remove them.

package.json
juliemr: Would this override what's currently in lib? I think it would be cleaner to just move everything we need into `built` and publish from there.

lib/.gitignore
cnishina: This is still needed. On the prepublish step, the transpiled javascript files are being copied into the lib folder.

alexeagle: I see, because the published app references lib/file.js in package.json
maybe it should reference built/file.js instead, like Julie suggests

lib/driverProviders/sauce.js
cnishina: In logger.js, looks like it would require changing the module to a class to use `export default`. Trying that out, it looks like `default` would make calls like `log.set(` change to `log.default.set(` for javascript files like this one.

lib/logger.ts
cnishina: When changing to `export namespace Logger {`, it appears that when doing a `var log = require('./logger')` would change the calls like `log.set` to  `log.Logger.set(` for javascript files that use it. Please advise.

tsconfig.json
cnishina: Fixed.

lib/configParser.ts
cnishina: Done. Added variables as optional to the Config object.

lib/configParser.ts
cnishina: Done. Removed null from config_ since they are now optional.

lib/driverProviders/sauce.js
alexeagle: No, there is another syntax to add an export default at the end of the file, choosing which symbol from the scope should be the default export.
So TypeScript should be able to express the logger.ts file so that the usage sites are unchanged.

lib/logger.ts
alexeagle: look around for another way to export - if you are stuck please ping me on IM and I'll patch your change locally and confirm if I'm right :)

lib/configParser.ts
cnishina: Done. Added clang-format to part of the prepublish / pretest step.

lib/configParser.ts
cnishina: Per a discussion we had today, we will remove them in the future. Issue track for this is: #2930.

cnishina:
Added jslint check back for `lib/` since transpiled files are now in `built/`

.gitignore
```diff
@@ -12,3 +12,6 @@ xmloutput*
 npm-debug.log
 
 *.swp
+
+typings/
+built/
```
juliemr: built/ is already at the top from the other PR, remove this line

gulpfile.js
```diff
@@ -1,39 +1,59 @@
...
 var gulp = require('gulp');
+var clangFormat = require('clang-format');
+var gulpFormat = require('gulp-clang-format');
 var runSequence = require('run-sequence');
-var spawnSync = require('child_process').spawnSync;
+var spawn = require('child_process').spawn;
+
+var runSpawn = function(done, task, opt_arg) {
+  var child = spawn(task, opt_arg);
+  child.stdout.on('data', function(data) {
```
juliemr: Instead of manually using console.log, can we use child_process stdio options? (https://nodejs.org/dist/latest-v4.x/docs/api/child_process.html#child_process_options_stdio) I think 'inherit' does what you want here.

testapp/ng2/app/async/async.component.js
```diff
@@ -29,7 +29,7 @@ System.register(['angular2/core', 'angular2/common', 'angular2/src/facade/async'
                     this.val3 = 0;
```
juliemr: Changes here seem unrelated, can this be a separate PR?

tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    "declaration": true,
+    "removeComments": false,
+    "noImplicitAny": false,
+    "outDir": "built/"
+  },
+  "exclude": [
+    "built",
+    "node_modules",
+    "testapp",
+    "typings/browser",
```
juliemr: Is there a reason not to just exclude all of 'typings'?

juliemr:
Just a couple cleanup comments, otherwise looks good!

tsconfig.json
cnishina: There needs to be a reference to the main.d.ts file.

testapp/ng2/app/async/async.component.js
cnishina: This should not be part of the PR. Will remove this.

gulpfile.js
cnishina: Done.

.gitignore
cnishina: Done.

cnishina:
Reference feature request: #1203

PR #2919, opened 2016-02-09
chore(built): adding gulp prepublish and gulp pretest tasks

cnishina:


package.json
```diff
@@ -18,9 +18,11 @@
     "jasmine": "2.4.1",
     "saucelabs": "~1.0.1",
     "glob": "~6.0",
+    "gulp": "^3.9.1",
```
juliemr: This should be in `devDependencies`

package.json
```diff
@@ -18,9 +18,11 @@
...
     "adm-zip": "0.4.7",
     "optimist": "~0.6.0",
     "q": "1.4.1",
+    "run-sequence": "^1.1.5",
```
juliemr: Ditto

.gitignore
```diff
@@ -1,3 +1,4 @@
+built/
 node_modules/
 selenium/
 testapp/inbrowsertest/
```
juliemr: I think we should we add lib/ to `.npmignore` as well, right? Or will TypeScript users need that for typing info?

gulpfile.js
```diff
@@ -0,0 +1,38 @@
...
+    console.log(child.stdout.toString());
+  }
+  if (child.status !== 0) {
+    throw new Error('jslint: child error');
+  }
+});
+
+gulp.task('pretest', function() {
+  runSequence(
+    ['webdriver:update', 'jslint']
```
juliemr: Should we also run `built:copy` here since we'll need it for testing now?

.gitignore
cnishina: Done. Added `lib/` to the `.npmignore`

juliemr:
Looks good! Go ahead and merge.

PR #2926, opened 2016-02-10
fix(configParser): use all the suites if no other spec sources are provided

fpipita:
Hi, this PR restores the behavior where if both `config.specs` and `config.suite` are empty, but `config.suites` is provided, all the specs from `config.suites` are scheduled to be executed.


fpipita:
I signed it!

juliemr:
Thanks for the fix! Merged as edfb52fadccf10c34d885c37e990dea0efbb0081

PR #2929, opened 2016-02-10
fix(config): Do not sort spec keys

### LINKED ISSUE #2928: @3.1.1 - Suites are loaded in alphabetical order instead of as described
suites: {
        b                 : [],
        c    : [],
        aa                : [],
        a           : [],
        d                    : []
}

These suites are ran in the following order: a > aa > b > c > d
I would like to run these suites as in protractor 3.0.0 (the order in list): b > c > aa > a > d

NickTomlin:
I think it makes sense to preserve the original non-sorted behavior here.

This fixes #2928

### DIFF
--- a/lib/configParser.js
+++ b/lib/configParser.js
@@ -151,7 +151,7 @@ ConfigParser.getSpecs = function(config) {
     return config.specs;
   }
 
-  Object.keys(config.suites || {}).sort().forEach(function(suite) {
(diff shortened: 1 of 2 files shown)


juliemr:
Seems reasonable. Merged as a2c7a4bf1fb2a3a509040ae8ec7737cc002b764e.

PR #2939, opened 2016-02-12
Fix docs for using protractor-cucumber-framework

lennym:
The code at https://github.com/angular/protractor/blob/master/lib/configParser.js#L167-L181 will convert a frameworkPath of `protractor-cucumber-framework` to a relative path relative to the configuration file. Instead the path must be supplied as an absolute path, and require.resolve is the best method to achieve this.

It also makes documentation consistent with https://github.com/mattfritz/protractor-cucumber-framework#implementation

### DIFF
--- a/docs/frameworks.md
+++ b/docs/frameworks.md
@@ -78,7 +78,7 @@ exports.config = {
   framework: 'custom',
 
   // path relative to the current config file
-  frameworkPath: 'protractor-cucumber-framework'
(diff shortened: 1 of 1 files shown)


lennym:
I signed it!

juliemr:
Thanks! Merged as a7734f85cf33dc1c8cb8c3db795974785400bd47

PR #2948, opened 2016-02-16
feat(typescript): converting task files to typescript

cnishina:


lib/taskLogger.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Log output such that metadata are appended.
+   * Calling log(data) will not flush to console until you call flush()
+   *
+   * @constructor
+   * @param {object} task Task that is being reported.
+   * @param {number} pid PID of process running the task.
+   */
+  constructor(task: any, pid: number) {
```
alexeagle: you can declare the property here as well, with `private task: any` - now you don't need to declare it above, or set the value in the body

lib/taskLogger.ts
```diff
@@ -0,0 +1,92 @@
...
+    this.task = task;
+    this.pid = pid;
+    this.buffer = '';
```
alexeagle: assign the default value at the declaration site - it is called immediately following the constructor

lib/taskRunner.ts
```diff
@@ -0,0 +1,131 @@
...
+   * './runner.js') or from a new process (via './runnerCli.js').
+   *
+   * @constructor
+   * @param {string} configFile Path of test configuration.
+   * @param {object} additionalConfig Additional configuration.
+   * @param {object} task Task to run.
+   * @param {boolean} runInFork Whether to run test in a forked process.
+   * @constructor
+   */
+  constructor(configFile, additionalConfig, task, runInFork) {
```
alexeagle: same as above, can be shorter

lib/taskRunner.ts
```diff
@@ -0,0 +1,131 @@
...
+
+      // stdout pipe
+      childProcess.stdout.on('data', (data) => { taskLogger.log(data); });
+
+      // stderr pipe
+      childProcess.stderr.on('data', (data) => { taskLogger.log(data); });
+
+      childProcess
+          .on('message',
+              (m) => {
```
alexeagle: hmm, any particular reasoning that you left noImplicitAny as false?

alexeagle:
LG for typescript

lib/taskLogger.ts
cnishina: Done.

lib/taskLogger.ts
cnishina: Done.

lib/taskRunner.ts
cnishina: Done.

lib/taskRunner.ts
cnishina: Changed it back to true.

juliemr:
Let's change the commit name to `chore(typescript)` - with the reasoning that it's not a feature that someone reading the changelog would care about.

lib/taskScheduler.ts
```diff
@@ -0,0 +1,159 @@
+import ConfigParser, {Config} from './configParser';
```
juliemr: Should `ConfigParser` be in the`{}`?

juliemr: Or is this the default thing working as intended?

tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "sourceMap": false,
     "declaration": true,
     "removeComments": false,
-    "noImplicitAny": false,
```
juliemr: Why this change? I feel like we might want to be more lenient about this unless we have to be strict.

juliemr:
Looks good other than a couple questions.

lib/taskScheduler.ts
cnishina: The class ConfigParser is labeled as the default and does not require `{}`.

tsconfig.json
cnishina: `noImplicitAny` is helpful to ensure types in the method signature. If we need to be lenient, we could use `any` as the type.

juliemr: Sorry, I misread the boolean. LGTM!

cnishina:
Reference feature request: #1203

cnishina:
Closed with commit 608104b

PR #2965, opened 2016-02-19
chore(typescript): fix `gulp tsc` to log to console

cnishina:
child_process.spawn with `{stdio: inherit}` was not printing to console.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -7,6 +7,7 @@ var runSequence = require('run-sequence');
 var spawn = require('child_process').spawn;
 
 var runSpawn = function(done, task, opt_arg) {
+  opt_arg = typeof opt_arg !== 'undefined' ? opt_arg : [];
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -39,7 +39,13 @@ gulp.task('typings', function(done) {
 });
 
 gulp.task('tsc', function(done) {
-  runSpawn(done, 'node_modules/typescript/bin/tsc');
+  var tsc = spawn('node_modules/typescript/bin/tsc');
```
juliemr: Hmmm, there must be a way to do this with stdio options. Gulp must be doing something funny to make `inherit` fail. Can we try `{stdio: [process.stdin, process.stdout, process.stderr]}`?

cnishina: Fixed. `spawn(task, opt_arg, {stdio: 'inherit'});` required that `opt_arg` to be defined as an array.

juliemr:
LGTM!

PR #2979, opened 2016-02-26
chore(issue template): ask people to tag their issues

sjelin:


NickTomlin:
:+1: makes sense to me

scripts/unit_test.json
```diff
@@ -1,7 +1,6 @@
 {
   "spec_dir": "",
   "spec_files": [
-    "spec/unit/*.js",
-    "website/docgen/spec/*.js"
+    "spec/unit/*.js"
```
NickTomlin: Is this just cleanup that we are including in the PR?

sjelin: This was a mistake :smile:

juliemr:
I actually find this clutters stuff - we can tag issues ourselves in a better, searchable way, and issue title is precious space to describe the actual problem. So I'm luke-warm on this change.

NickTomlin:
I think that's fair. Although it might help put some context around an issue that might help triage a little better (e.g. I recently closed something I thought was a question when it was really a feature request).

I also like the fact that it highlights that thinking about the type of issue is important, hopefully that further might help people opt-out on debugging/usage questions. 

Those are both pretty soft arguments though, so i'm fine either way :smile:

sjelin:
I think we still have a problem where its not clear if someone is asking a support question, a bug report, or a feature request.  We can also just remove the title tags when we add the github labels.

juliemr:
I'm going to go ahead and executively close this one - I think that the current template is meeting our needs decently.

Issue #2893, opened 2016-02-01
feat(config): Don't pollute global namespace


sjelin:
We should have an option in the config like `noGlobals` which, instead of putting `$`/`$$`/`by`/`element`/`browser` directly into the global namespace, instead makes them elements of `protractor`.

cnishina:
Closed with #2988 (commit 1cbbe4f).

PR #2988, opened 2016-03-02
feat(config): no globals option

cnishina:


cnishina:
Reference feature request: #2893

lib/runner.js
```diff
@@ -156,13 +156,26 @@ Runner.prototype.controlFlow = function() {
...
-  // Export protractor to the global namespace to be used in tests.
+  if (this.config_.noGlobals) {
+    // Keep $, $$, element, and by/By under the global protractor namespace
+    protractor.browser = browser_;
+    protractor.$ = browser_.$;
+    protractor.$$ = browser_.$$;
+    protractor.element = browser_.element;
+    protractor.by = protractor.By = protractor.By;
+  }
+  else {
```
NickTomlin: Should this be bumped to the previous line?

lib/runner.js
```diff
@@ -156,13 +156,26 @@ Runner.prototype.controlFlow = function() {
  * @private
  */
 Runner.prototype.setupGlobals_ = function(browser_) {
-  // Export protractor to the global namespace to be used in tests.
+  if (this.config_.noGlobals) {
+    // Keep $, $$, element, and by/By under the global protractor namespace
+    protractor.browser = browser_;
```
NickTomlin: What about always assigning these and just not setting globals if `noGlobals` is true? It would make moving between the two easier (although I guess one could argue that opening the door to mixing and matching is a _bad_ thing)

lib/runner.js
```diff
@@ -156,13 +156,26 @@ Runner.prototype.controlFlow = function() {
...
+    protractor.$ = browser_.$;
+    protractor.$$ = browser_.$$;
+    protractor.element = browser_.element;
+    protractor.by = protractor.By = protractor.By;
```
cnishina: Remove `protractor.By = protractor.By`

lib/runner.js
cnishina: Yes. Making the change.

lib/runner.js
cnishina: Great catch @cnishina. Done.

lib/runner.js
cnishina: I'm indifferent on this one. Deferring to issue author @sjelin.

sjelin: I'm with Nick on this one.  That's how I've seen other libraries do it

package.json
```diff
@@ -33,12 +33,12 @@
     "gulp": "^3.9.1",
     "gulp-clang-format": "^1.0.23",
     "jshint": "2.9.1",
-    "lodash": "^2.4.1",
+    "lodash": "^4.5.1",
```
sjelin: why this?

cnishina: On node v5, this is causing an error. Upgrading to 4.5.1 removes the error.

```
[chrome #01] PID: 15508
[chrome #01] Specs: /Users/.../protractor/spec/directConnect/directconnect_spec.js
[chrome #01]
[chrome #01] module.js:341
[chrome #01]     throw err;
[chrome #01]     ^
[chrome #01]
[chrome #01] Error: Cannot find module 'lodash/assign'
[chrome #01]     at Function.Module._resolveFilename (module.js:339:15)
```

lib/runner.js
cnishina: Fixing.

Issue #987, opened 2014-06-26
Add argument to webdriver-manager to allow custom jar file name


sajidmohammed:
Just like argument "out_dir", allow "out_jarname" to allow custom name for the selenium server standalone jar. For example, currently the file name is selenium-server-standalone-2.42.0.jar which contains version number in the name. I would like to name it generic: selenium-server-standalone.jar so that I do not have to hard code in the config file.

fat-troll:
Isn't it "seleniumServerJar" parameter?

cnishina:
Instead of allowing for a generic name for selenium server jar or having to fix the `config.json`,  `webdriver-manager ` could download / start different selenium version numbers from the command line. See #2998 for more details.

PR #2998, opened 2016-03-04
feat(webdriver): add support for custom versions for selenium chrome, and ie driver

cnishina:
The `config.json` has the default versions that `webdriver-manager` uses to download the selenium server jar, internet explorer driver, and chrome driver. This can now be overridden by the `versions.standalone`, `versions.chrome`, and `versions.ie` options. Below describes the new features:

### Download version specific drivers
```
webdriver-manager update --versions.chrome=2.19 --versions.standalone=2.51.0 --versions.ie=2.51.0 --ie32
```
Note that either `--ie32`  or `--ie` option is still required to download the internet explorer driver. Also, the `--ie32` overrides the `--ie` option and will only download the 32-bit version.

### Starting Selenium with version specific drivers
```
webdriver-manager start --versions.chrome=2.19 --versions.standalone=2.51.0

seleniumProcess.pid: 40388
18:40:11.657 INFO - Launching a standalone Selenium Server
Setting system property webdriver.chrome.driver to /src/protractor/selenium/chromedriver_2.19
18:40:11.695 INFO - Java: Oracle Corporation 25.72-b15
18:40:11.696 INFO - OS: Mac OS X 10.11.3 x86_64
18:40:11.707 INFO - v2.51.0, with Core v2.51.0. Built from revision 1af067d
18:40:11.779 INFO - Driver provider org.openqa.selenium.ie.InternetExplorerDriver registration is skipped:
registration capabilities Capabilities [{ensureCleanSession=true, browserName=internet explorer, version=, platform=WINDOWS}] does not match the current platform MAC
```

### Check current versions downloaded
```
webdriver-manager status

selenium standalone versions available: 2.51.0, 2.52.0 [default]
chromedriver versions available: 2.19, 2.21 [default]
IEDriver versions available: Win32_2.52.0, x64_2.52.0
```
Note that the IEDriver versions have which version that is currently downloaded (32-bit and 64-bit).

### Remove the files stored in the `out_dir`
```
webdriver-manager clean
```

### Notes on files and version:
The unzipped driver previously did not have a version number. With this new feature, the file was renamed to include the version number. For example chrome driver version 2.21, the file has been renamed from `chromedriver` to `chromedriver_2.21`. This change also applies to the internet explorer driver. For example, version 2.52.0 of `IEDriverServer.exe` has been renamed to `IEDriverServer_Win32_2.52.0.exe` for the 32-bit version and `IEDriverServer_x64_2.52.0.exe` for the 64-bit version.

### DIFF
--- a/bin/webdriver-manager
+++ b/bin/webdriver-manager
@@ -25,71 +25,124 @@ var shortVersion = function(version) {
   return version.slice(0, version.lastIndexOf('.'));
 }
 
+var seleniumPrefix = 'selenium-server-standalone-',
+    seleniumVersion = versions.selenium,
+    chromePrefix = 'chromedriver_',
+    chromeVersion = versions.chromedriver,
+    iePrefix = 'IEDriverServer_',
+    ie32 = 'Win32_',
+    ie64 = 'x64_',
+    ieVersion = versions.iedriver;
+
 var binaries = {
   standalone: {
     name: 'selenium standalone',
     isDefault: true,
-    prefix: 'selenium-server-standalone',
-    filename: 'selenium-server-standalone-' + versions.selenium + '.jar',
+    defaultVersion: versions.selenium,
+    osType: 'ANY',
+    prefix: function() {
+      return seleniumPrefix;
+    },
+    version: function() {
+      return seleniumVersion;
+    },
+    suffix: function() {
+      return '.jar';
+    },
+    filename: function() {
+      return this.prefix() + this.version() + this.suffix();
+    },
     cdn: 'https://selenium-release.storage.googleapis.com/',
     url: function() {
-      return this.cdn +
-          shortVersion(versions.selenium) + '/' +
-          'selenium-server-standalone-' + versions.selenium + '.jar';
+      return this.cdn + shortVersion(this.version()) + '/' +
+          this.prefix() + this.version() + this.suffix();
     }
   },
   chrome: {
     name: 'chromedriver',
     isDefault: true,
-    prefix: 'chromedriver_',
-    filename: 'chromedriver_' + versions.chromedriver + '.zip',
-    cdn: 'https://chromedriver.storage.googleapis.com/',
-    url: function() {
-      var urlPrefix = this.cdn +
-          versions.chromedriver + '/chromedriver_';
+    osType: 'ANY',
+    defaultVersion: versions.chromedriver,
+    prefix: function() {
+      return chromePrefix;
+    },
+    version: function() {
+      return chromeVersion;
+    },
+    suffix: function() {
       if (os.type() == 'Darwin') {
-        return urlPrefix + 'mac32.zip';
+        return 'mac32.zip';
       } else if (os.type() == 'Linux') {
         if (os.arch() == 'x64') {
-          return urlPrefix + 'linux64.zip';
+          return 'linux64.zip';
         } else {
-          return urlPrefix + 'linux32.zip';
+          return 'linux32.zip';
         }
       } else if (os.type() == 'Windows_NT') {
-        return urlPrefix + 'win32.zip';
+        return 'win32.zip';
       }
+    },
+    filename: function() {
+      return this.prefix() + this.version() + this.suffix();
+    },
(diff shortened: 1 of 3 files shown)


cnishina:
Reference issue: #987.

bin/webdriver-manager
```diff
@@ -260,12 +261,19 @@ for (name in binaries) {
 
 switch (argv._[0]) {
   case 'start':
-    if (!binaries.standalone.exists) {
+    if (!binaries.standalone.exists && !fs.existsSync(argv.seleniumServerJar)) {
```
sjelin: Shouldn't this be `!binaries.standalone.exists && (!argv.seleniumServerJar || !fs.existsSync(argv.seleniumServerJar))`?

cnishina: Fixed.

juliemr:
Ok I know I was the one who tagged that issue as a feature request with PRS PLZ. But, now that I see the changes - I'm worried about feature creep and command line option creep. What about making a custom chromedriver location? Custom IE Driver?

`webdriver-manager` is intended as one simple way of downloading/starting up selenium, but it's doing nothing advanced. If someone wants a more personalized configuration, in my opinion, they should be just calling the standalone directly.

Am I missing a use case which makes this a particularly useful option to be able to change?

cnishina:
As discussed, we will support the ability to override the config.json to download a different selenium jar file. Both for `update` and `start`, will have the option to specify the `--seleniumVersion`

cnishina:
Looking at the `status` command, the showing that selenium is up to date is not helpful if there are multiple versions of selenium downloaded. This is the proposed output:

```
❯ ./bin/webdriver-manager status
selenium version(s) available: 2.51.0 [default], 2.52.0,
chromedriver is up to date
IEDriver is not present
IEDriver-32bit is not present
```

where 2.51.0 is shown as default per the config.json

bin/webdriver-manager
```diff
@@ -103,6 +115,7 @@ var cli = optimist.
     describe('out_dir', 'Location to output/expect ').
     default('out_dir', SELENIUM_DIR).
     describe('seleniumPort', 'Optional port for the selenium standalone server').
+    describe('seleniumVersion', 'Optional selenium standalone server version').
```
juliemr: Can we have the `versions` object be available as a command line option, and use that for controlling all versions? So the invocation could look like:

```
> webdriver-manager update --versions.standalone=1.2.3 --versions.chromedriver=3.0
```

cnishina:
work in progress note: when files are unzipped, they need to be renamed to keep track of the current version.

cnishina:
after unzipping the file such as `chromedriver` and `IEDriverServer.exe`, they are renamed as follows:
```
./bin/webdriver-manager update --versions.chromedriver=2.20
```
downloads the file: `chromedriver_2.20linux64.zip`
unzips the file: `chromedriver`
then renames it to: `chromedriver_2.20`

bin/webdriver-manager
```diff
@@ -247,12 +311,26 @@ var executableName = function(file) {
 // Setup before any command.
 var existingFiles = fs.readdirSync(argv.out_dir);
 
+// update versions
+if (argv.versions) {
+  if (argv.versions.standalone) {
+    seleniumVersion = argv.versions.standalone;
```
juliemr: What about not having separate `*Version` variables and just doing:

```js
if (argv.versions) {
  for (binary in argv.version) {
    binaries[binary].version = argv.versions[binary]
  }
}
```

juliemr:
LGTM other than one cleanup suggestion.

evilaliv3:
so looking forward for this! thanks @cnishina 

@juliemr / @sjelin is there already a schedule for the release?

juliemr:
LGTM! Let's note in the commit message that this adds a version string to unzipped executables (in case someone is doing something custom and relying on that in a weird way, we'll have it documented. See https://xkcd.com/1172/)

cnishina:
Added the comment about versions to the commit message.

cnishina:
Somehow my pull request had an old fix. #3022 has the correct changes.

PR #3018, opened 2016-03-10
fix(bootstrap): enable debug info before setting defer label

### LINKED ISSUE #3009: Protractor 3.1.1 doesn't run test for application having .debugInfoEnalbled(false)
Pretty new here,

Recently upgraded to Protractor 3.1.1 and testing the debug feature [(148f020)](https://github.com/angular/protractor/commit/148f020bf4bbd71e17326581a2f7ed6f4ff55832)

I understand that the debug flag added in the protractor.js helps to run the tests for applications with `.debugInfoEnabled(false);`

But my test doesn't without the following hack 
`browser.executeScript('window.name = "NG_ENABLE_DEBUG_INFO!"');`

But wasn't this the whole point of upgrading to Protractor 3.1.1 so you don't have to do this?!

Any suggestions?

juliemr:
Note that in most cases, this should not have surfaced as an issue because
the base test mock modules will also try to turn on debug info.

Closes #3009

### DIFF
--- a/lib/protractor.js
+++ b/lib/protractor.js
@@ -637,7 +637,8 @@ Protractor.prototype.get = function(destination, opt_timeout) {
 
   this.driver.get(this.resetUrl).then(null, deferred.reject);
   this.executeScript_(
-      'window.name = "' + DEFER_LABEL + ENABLE_DEBUG_INFO_LABEL + '" + window.name;' +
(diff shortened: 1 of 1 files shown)


juliemr:
@sjelin sanity check please!

cnishina:
Closed with d3bd170.

Issue #607, opened 2014-03-13
Extract "webdriver-manager" as an independent Node module


JamesMGreene:
The latest project I'm working on is not Angular-based, so using Protractor doesn't make much sense.  However, specific parts of Protractor would still be incredibly useful for other WebDriver purposes.

I'd love to see `webdriver-manager` extracted from Protractor into its own independent Node module. As far as the `webdriver-manager update` process goes, then, you'd probably either (a) want to install it into the `webdriver-manager` module folder and then provide a class that Protractor can `require` to get the path, (b) allow a path to be passed in as an arg, or (c) install into the CWD.

Thoughts?  At a cursory glance, I don't see any strong reason why it can't be easily extracted. Adding a dependency on `protractor` just to get `webdriver-manager` seems a bit extreme.

JamesMGreene:
@juliemr: And the survey says: ______________? :grinning:

dsebastien:
+1

cnishina:
Released [webdriver-manager](https://github.com/angular/webdriver-manager) into a separate node module [v10.0.1](https://github.com/angular/webdriver-manager/tree/10.0.1).

cnishina:
Since webdriver-manager is its own node module, should we remove it from protractor?

Installing webdriver-manager globally then installing protractor globally, the web-manager is overwritten.

Thoughts? @juliemr, @sjelin, @mgiambalvo?

JamesMGreene:
@cnishina: Awesome, thanks!  Out of curiosity, why is it starting at version `10` instead of `1`? 😛

juliemr:
@cnishina yes, absolutely

cnishina:
@JamesMGreene 
Version 9 and lower https://github.com/pose/webdriver-manager
Version 10+ https://github.com/angular/webdriver-manager

JamesMGreene:
Ah, right, now I remember that someone else had that module name before.  Thanks again!

igniteram:
@JamesMGreene @juliemr @cnishina  Thanks a ton!, i was looking for something like this

PR #3022, opened 2016-03-14
chore(webdriver): fix version support

cnishina:


mgiambalvo:
It's a little confusing to me to have prefix be a property now but suffix() is still an accessor method. I'd recommend making it getSuffix(). Other than that, looks good.

cnishina:
The accessor methods will change with #607.

PR #3023, opened 2016-03-14
feat(plugins) Calculate skipAngularStability dynamically.

mgiambalvo:
This allows plugins to turn Protractor's default synchronization on and
off as needed.

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -22,7 +22,6 @@ var Plugins = function(config) {
   this.pluginObjs = [];
   this.assertions = {};
   this.resultsReported = false;
-  this.skipAngularStability = false;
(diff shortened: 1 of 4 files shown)


lib/plugins.js
```diff
@@ -129,6 +126,18 @@ function printPluginResults(specResults) {
 }
 
 /**
+ * Returns true if any loaded plugin has skipAngularStability enabled.
```
sjelin: This should go at the end of the file probably

sjelin:
Small style change needed, some kind of test would be good too probably

mgiambalvo:
Added tests and fixed that style issue.

PR #3041, opened 2016-03-18
fix(protractor): Shows locator used when a timeout occur.

### LINKED ISSUE #2996: More descriptive errors for element.all(locator) when waitForAngular fails.
Feature Request:
**tl;dr:** Add the locator to error messages in waitForAngular.

If the waitForAngular call times out when calling element.all(locator) we get an error message like the following:

`Timed out waiting for Protractor to synchronize with the page after 9 seconds.
Please see https://github.com/angular/protractor/blob/master/docs/faq.md. 
The following tasks were pending:
    - $timeout: function o(){var e=n&&n.$$destroyed,t=e?[]:r.queue,o=e?null:r.digest;r.queue=[],r.timeout=null,r.digest=!1,t.forEach(function(e){e()}),o&&d.$digest()}
`
Unfortunately, we don't have any information about the specific locator which initiated the waitForAngular call. Should be easy enough to pass the locator as the description here:
https://github.com/angular/protractor/blob/master/lib/element.js#L133

I would love to have an error message like this:

`Timed out waiting for Protractor to synchronize with the page after 9 seconds. 
Please see https://github.com/angular/protractor/blob/master/docs/faq.md.
Locator: by.css('.parent'). 
The following tasks were pending:
    - $timeout: function o(){var e=n&&n.$$destroyed,t=e?[]:r.queue,o=e?null:r.digest;r.queue=[],r.timeout=null,r.digest=!1,t.forEach(function(e){e()}),o&&d.$digest()}
`
Improve error clarity and ease of debugging.
 
No. Should be really easy to implement too!

luishvz:
Fixes #2996

### DIFF
--- a/lib/element.js
+++ b/lib/element.js
@@ -130,7 +130,7 @@ ElementArrayFinder.prototype.all = function(locator) {
   var getWebElements = function() {
     if (self.getWebElements === null) {
       // This is the first time we are looking for an element
-      return ptor.waitForAngular().then(function() {
+      return ptor.waitForAngular('Locator: ' + locator).then(function() {
         if (locator.findElementsOverride) {
           return locator.findElementsOverride(ptor.driver, null, ptor.rootEl);
         } else {
--- a/lib/protractor.js
+++ b/lib/protractor.js
@@ -401,8 +401,11 @@ Protractor.prototype.waitForAngular = function(opt_description) {
       }
       if (timeout) {
         var errMsg = 'Timed out waiting for Protractor to synchronize with ' +
-            'the page after ' + timeout + '. Please see ' +
-            'https://github.com/angular/protractor/blob/master/docs/faq.md';
+        'the page after ' + timeout + '. Please see ' +
+        'https://github.com/angular/protractor/blob/master/docs/faq.md';
+        if(description.startsWith(' - Locator: ')){
+            errMsg += '\n' + description
+        };
         var pendingTimeoutsPromise;
         if (self.trackOutstandingTimeouts_) {
           pendingTimeoutsPromise = self.executeScript_(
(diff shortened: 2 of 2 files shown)


luishvz:
Signed it!

lib/protractor.js
```diff
@@ -401,8 +401,8 @@ Protractor.prototype.waitForAngular = function(opt_description) {
       }
       if (timeout) {
         var errMsg = 'Timed out waiting for Protractor to synchronize with ' +
-            'the page after ' + timeout + '. Please see ' +
-            'https://github.com/angular/protractor/blob/master/docs/faq.md';
+        'the page after ' + timeout + '. Locator: ' + description + '. Please see ' +
```
sjelin: Maybe some newlines here?  Also, the description won't always be a locator

lib/element.js
```diff
@@ -130,7 +130,7 @@ ElementArrayFinder.prototype.all = function(locator) {
   var getWebElements = function() {
     if (self.getWebElements === null) {
       // This is the first time we are looking for an element
-      return ptor.waitForAngular().then(function() {
+      return ptor.waitForAngular(locator).then(function() {
```
sjelin: `opt_description` is supposed to be a string.  Maybe instead it should be `waitForAngular('Locator: ' + locator)`?

luishvz:
@sjelin modified the parameter sent to waitForAngular() and now the `description` is only added to error message when is a locator.

`Failed: Timed out waiting for Protractor to synchronize with the page after 11 seconds. Please see https://github.com/angular/protractor/blob/master/docs/faq.md
     - Locator: by.binding("slowAngularTimeoutPromiseStatus"). 
    The following tasks were pending:
     - $timeout: function () {
          // intentionally empty
        }
`

cnishina:
closed with b6f1a63da77ec88a3f487e5a099df8febe3742aa

juliemr:
For the future - changes like this are not a 'fix', since it's not fixing a bug. I would call it a 'feat' or 'chore'.

luishvz:
got it @juliemr , thanks.

PR #3056, opened 2016-03-23
feat(protractor): isPresent() now is friendly to out of bounds errors

### LINKED ISSUE #2917: first is not friendly with isPresent
ElementArrayFinder.first() (and probably last too) is not currently friendly with the ElementFinder.isPresent. I.e. assume the following test: 

it('First should be friendly with isPresent', function () {
    expect(element.all(by.css('nonexistingselector')).first().isPresent()).toBe(false);
});

That is quite natural assumption but the test is not passing. It fails currently with the error:

Failed: Index out of bound. Trying to access element at index: 0, but there are only 0 elements that match locator By.cssSelector("nonexistingselector")

Can this be improoved somehow?

PS.
Actually I faced it in a more complex case where I was trying to do something like this:

function byExactText(text) {
    // Cannot use by.cssContainingText here because
    // I need item with name "Element1" whereas
    // there also is another item with name "Element1 Extended".
    return element.all(by.css('.item')).filter(function(item) {
        return item.element(by.css('.name')).getText().then(function(text) {
            return text === name;
        });
    }).first();
}
expect(byExactText('Element1').isPresent()).toBe('false');
expect(byExactText('AbsentElmnt').isPresent()).toBe('false');

And here I cannot easily replace to something like:

byExactText('AbsentElmnt').then(function(items){
    return items.length > 0 ? items[0] : undefined; // well, what to replace the indefined with?
}).isPresent();

Because the yet-unresolved-promise returned from "then" is not an instance of ElementFinder and has no appropriate methods...

PPS. This issue has something common with https://github.com/angular/protractor/issues/2314

luishvz:
fixes #2917 
@juliemr worked around your comments in that issue,could you check these changes?. Modified ElementArrayFinder.prototype.get function, now it does't crash the flow when an out of bounds err is thrown, instead of that a warning is prompted in console.

### DIFF
--- a/lib/element.js
+++ b/lib/element.js
@@ -242,10 +242,11 @@ ElementArrayFinder.prototype.get = function(index) {
         i = parentWebElements.length + i;
       }
       if (i < 0 || i >= parentWebElements.length) {
-        throw new Error('Index out of bound. Trying to access element at ' +
+            console.warn('Index out of bound. Trying to access element at ' +
             'index: ' + index + ', but there are only ' +
             parentWebElements.length + ' elements that match locator ' +
             self.locator_.toString());
+            return [];
       }
       return [parentWebElements[i]];
     });
--- a/spec/basic/elements_spec.js
+++ b/spec/basic/elements_spec.js
@@ -12,6 +12,7 @@ describe('ElementFinder', function() {
         browser.findElement(by.binding('username')).getText());
   });
 
+
   it('should wait to grab the WebElement until a method is called', function() {
     // These should throw no error before a page is loaded.
     var usernameInput = element(by.model('username'));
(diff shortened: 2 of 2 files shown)


lib/element.js
```diff
@@ -242,7 +242,7 @@ ElementArrayFinder.prototype.get = function(index) {
         i = parentWebElements.length + i;
       }
       if (i < 0 || i >= parentWebElements.length) {
-        throw new Error('Index out of bound. Trying to access element at ' +
+            console.warn('Index out of bound. Trying to access element at ' +
```
sjelin: I still think `.get` should fail for out of bounds indexes, instead of returning `undefined`

luishvz:
@sjelin I understand your point, but if .get fails there is no friendly response with another actions like isPresent() even if the test passes, now it will crash when operations like click() in undefined elements but there will be the warning in the console explaining the out of bounds error.

sjelin:
I think people are less likely to see a warning logged to the console than an error thrown at the time they use `.get`.  We could add an optional variable to `.get` called `opt_warnOnOutOfBounds` or something for users who want to do `elems.get(i).isPresent()`.

@juliemr @cnishina @mgiambalvo

luishvz:
I was thinking on an approach like that with an optional variable but I tought it was going to be the same result at the end. The flow of the test will crash and show the out of bounds error.
But yeah that would be a nice approach.

juliemr:
Can't we have `isPresent` check for the rejected promise, intercept it, and do the right thing? No need for any boolean params.

sjelin:
Good idea Julie.  See #3108

PR #3064, opened 2016-03-25
feat(driverProvider/sauce) Add build Id as configurable option

### LINKED ISSUE #2240: Attach build identifier to tests on SauceLabs
SauceLabs has just been updated so you can pass your CI build identifier and it'll group all the tests for that build together.

Neat!

Is there anyway we currently set this up within our protractor config (or better yet, on the command line) ?

They want it in this format:

    "build": "build-1234"

eduncan26:
* Add build Id as a configurable option for sauce users
* Update sanity tests to run on the built dir
* Update sanity tests to call getNewDriver() function

closes #2240

Tested by running sanity tests locally. Attached are screen shots of the execution, and the build shown at sauce labs.

test_execution

sauce_labs

Note: I've regenerated my sauce key so no worries about it showing in the screenshot

### DIFF
--- a/docs/referenceConf.js
+++ b/docs/referenceConf.js
@@ -72,6 +72,8 @@ exports.config = {
   // Use sauceAgent if you need customize agent for https connection to
   // saucelabs.com (i.e. your computer behind corporate proxy)
   sauceAgent: null,
+  // Use sauceBuild if you want to group test capabilites by a build ID
+  sauceBuild: null,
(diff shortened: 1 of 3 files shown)


eduncan26:
Looks like the test failed because of flash.

failed_test_screencast

cnishina:
closed with 582411b7ad6c0f9176b231dc51dc328b98affbdf

PR #3068, opened 2016-03-28
feat(webdriver): extract webdriver-manager into a separate node module

### LINKED ISSUE #607: Extract "webdriver-manager" as an independent Node module

### LINKED ISSUE #2402: Add ability to create chromedriver.log
I have seen that chromedriver can output a logfile (https://sites.google.com/a/chromium.org/chromedriver/logging)

This page shows how to set this up when executing the exe directly: `chromedriver.exe --verbose --log-path=chromedriver.log`

There is no apparent way to set these arguments from protractor however.

See this StackOverflow question for more info:

http://stackoverflow.com/questions/31662828/how-to-access-chromedriver-logs-for-protractor-test/31662935

cnishina:
closes #607, closes #2402


NickTomlin:
So much :clap: :confetti_ball: 

This is very OT, but have we considered asking the `webdriver-manager` folks to transfer ownership to the angular org and publishing this there?

sjelin:
Nit: commit should say "extract webdriver-tool into..."

re: @NickTomlin 
Yeah, Craig, if you want I'm sure you could get a repo under `angular`.

Also, we are the `webdriver-manager` folks :smile:

juliemr:
We're discussing with the owner of `webdriver-manager` on npm - conclusion TBD.

NickTomlin:
@sjelin ah yeah I should have been clearer, I mean the community npm [webdriver-manager](https://www.npmjs.com/package/webdriver-manager) fork

Glad to hear we are talking with the NPM owner. Hopefully we can work that out!

cnishina:
Name change from webdriver-tool to [webdriver-manager](https://github.com/angular/webdriver-manager). Updating this pr to reflect this.

cnishina:
browserStack tests are failing which needs to be addressed.

README.md
```diff
@@ -45,9 +45,10 @@ Clone the github repository:
     npm install
     cd ..
 
-Start up a selenium server. By default, the tests expect the selenium server to be running at `http://localhost:4444/wd/hub`. A selenium server can be started with `webdriver-manager`.
+Start up a selenium server. By default, the tests expect the selenium server to be running at `http://localhost:4444/wd/hub`. A selenium server can be started with [webdriver-manager](https://github.com/angular/webdriver-manager).
```
juliemr: Maybe add something like: "which is included in bin/webdriver-manager"?

config.json
```diff
@@ -3,5 +3,11 @@
     "selenium": "2.53.0",
     "chromedriver": "2.21",
     "iedriver": "2.53.1"
+  },
+  "cdnUrls": {
```
juliemr: Are these not the default?

juliemr:
We should mention in the commit description that this will change the filenames by giving everything versions.

cnishina:
* As discussed, versioning of binaries was previously added in ~3.2
* Fixes for webdriver-manager 10.0.4 for configuration file and path issues.
* Removed the config.json. This will be managed now by webdriver-manager
* wedriver-manager downloads the file to the node_modules/webdriver-manager/selenium folder. This will no longer appear in the protractor directory.

PR #3080, opened 2016-04-01
fix(NoSuchElementError): add 'new' keyword for creating class

davinkevin:
Prevent the following error if an element isn't found in the dom : `Class constructors cannot be invoked without 'new'`


davinkevin:
CLA signed :smile:

Issue #3079, opened 2016-04-01
Counting non-existing elements fails in Protractor 3.2.x


szimek:
Bug report
---

- Node Version: `5.9.1`
- Protractor Version: `3.2.2`
- Browser(s): `Chrome 50`
- Operating System and Version `OS X 10.11.4`
- Your protractor configuration file: https://github.com/angular/protractor/blob/master/spec/basicConf.js
- A relevant example test:
```javascript
it('should return count for not existing elements', function() {
  browser.get('index.html#/form');
  var parent = element(by.css('not-existing-element'));
  var listItems = parent.all(by.css('li'));

  expect(listItems.count()).toBe(0);
});
```
- Output from running the test
```
1) ElementFinder should return count for not existing elements
  Message:
    Failed: Class constructors cannot be invoked without 'new'
  Stack:
    TypeError: Class constructors cannot be invoked without 'new'
        at Object.NoSuchElementError (/Users/szimek/Projects/src/protractor/node_modules/selenium-webdriver/error.js:229:14)
        at /Users/szimek/Projects/src/protractor/built/element.js:343:37
        at Promise.invokeCallback_ (/Users/szimek/Projects/src/protractor/node_modules/selenium-webdriver/lib/promise.js:1329:14)
        at TaskQueue.execute_ (/Users/szimek/Projects/src/protractor/node_modules/selenium-webdriver/lib/promise.js:2790:14)
        at TaskQueue.executeNext_ (/Users/szimek/Projects/src/protractor/node_modules/selenium-webdriver/lib/promise.js:2773:21)
        at /Users/szimek/Projects/src/protractor/node_modules/selenium-webdriver/lib/promise.js:2652:27
        at /Users/szimek/Projects/src/protractor/node_modules/selenium-webdriver/lib/promise.js:639:7
        at process._tickCallback (node.js:412:9)
    From: Task: Run it("should return count for not existing elements") in control flow
        at Object.<anonymous> (/Users/szimek/Projects/src/protractor/node_modules/jasminewd2/index.js:81:14)
        at /Users/szimek/Projects/src/protractor/node_modules/jasminewd2/index.js:18:5
        at Promise.invokeCallback_ (/Users/szimek/Projects/src/protractor/node_modules/selenium-webdriver/lib/promise.js:1329:14)
        at TaskQueue.execute_ (/Users/szimek/Projects/src/protractor/node_modules/selenium-webdriver/lib/promise.js:2790:14)
[...]

PR #3081, opened 2016-04-01
fix(NoSuchElementError): add 'new' keyword to instantiate class

davinkevin:
The class NoSuchElementError is called without the new keyword in the `ElementArrayFinder.prototype.count` causing a `Class constructors cannot be invoked without 'new'`


cnishina:
This looks like it will close #3079. 

https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/error.js#L239

cnishina:
Closed with 88dd568c5295234a5211a23e12666e199606e437

PR #3094, opened 2016-04-06
feat(element): equals

sjelin:
Easier to use version of webdriver.WebElement.equals

Did this instead of documenting webdriver version


lib/element.js
```diff
@@ -953,6 +953,20 @@ ElementFinder.prototype.allowAnimations = function(value) {
   return this.elementArrayFinder_.allowAnimations(value).toElementFinder_();
 };
 
+
+/**
+ * Compares an element to this one for equality
```
juliemr: Nit: descriptions should be full sentences, end with `.`

PR #3095, opened 2016-04-06
chore(docs): protractor style docs for `findElement`, `sendKeys`, `ge…

sjelin:
…tTagName`, and `getCssValue`

### DIFF
--- a/lib/selenium-webdriver/webdriver.js
+++ b/lib/selenium-webdriver/webdriver.js
@@ -71,16 +71,6 @@ webdriver.WebElement = function(driver, id) {};
 goog.inherits(webdriver.WebElement, webdriver.Serializable);
 
 /**
- * Compares to WebElements for equality.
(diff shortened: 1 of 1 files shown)


lib/selenium-webdriver/webdriver.js
```diff
@@ -108,44 +98,13 @@ webdriver.WebElement.prototype.getRawId = function() {};
...
- *         return visibleLinks[0];
- *       });
- *     }
- *
- * @param {!(webdriver.Locator|webdriver.By.Hash|Function)} locator The
- *     locator strategy to use when searching for the element.
- * @return {!webdriver.WebElement} A WebElement that can be used to issue
- *     commands against the located element. If the element is not found, the
- *     element will be invalidated and all scheduled commands aborted.
+ * @param {webdriver.Locator} subLocator
```
cnishina: We should keep the return type: `@return {!webdriver.WebElement}`

PR #3098, opened 2016-04-06
docs(webelement): Adding examples to WebElement docs.

mgiambalvo:


lib/selenium-webdriver/webdriver.js
```diff
@@ -361,6 +440,14 @@ webdriver.WebElement.prototype.clear = function() {};
 
 /**
  * Schedules a command to test whether this element is currently displayed.
+ *
+ * @view
+ * <div id="foo" style="visibility:hidden">
+ *
+ * @example
+ * var foo = element(by.id('foo'));
+ * expect(foo.isDisplayed()).toBe(false);
```
sjelin: You sure about this?  In CSS a hidden element might still be displayed

mgiambalvo: Yup, I tried it out. That's why I made that the example, because I wasn't sure what it would do.

PR #3101, opened 2016-04-07
docs(webdriver): protractor style docs for `getDriver`, `getId`, and `click`

cnishina:


lib/selenium-webdriver/webdriver.js
```diff
@@ -82,11 +82,33 @@ webdriver.WebElement.equals = function(a, b) {};
...
+ * @view
+ * <ul class="pet">
+ *   <li class="dog">Dog</li>
+ *   <li class="cat">Cat</li>
+ * </ul>
+ *
+ * @example
+ * // returns the dog web element
+ * var dog = element(by.css('.dog')).getWebElement();
+ * expect(dog.getId()).not.toBe(undefined);
```
sjelin: *not* toBe undefined?

cnishina: getId returns a string that is a number. So, I'm getting the id to the web element and it exists.

sjelin: Ok, I understand now.  You should add a line up top explaining what this method actually does, because I assumed it got the `id` attribute of an element

PR #3105, opened 2016-04-07
docs(webelement): Remove takeScreenshot from docs.

mgiambalvo:
We don't seem to wrap this for WebElements.

### DIFF
--- a/lib/selenium-webdriver/webdriver.js
+++ b/lib/selenium-webdriver/webdriver.js
@@ -455,19 +455,6 @@ webdriver.WebElement.prototype.isDisplayed = function() {};
 
 
 /**
- * Take a screenshot of the visible region encompassed by this element's
- * bounding rectangle.
- *
- * @param {boolean=} opt_scroll Optional argument that indicates whether the
- *     element should be scrolled into view before taking a screenshot. Defaults
- *     to false.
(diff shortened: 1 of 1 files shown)


mgiambalvo:
Note that this is the takeScreenshot() method on WebElement, not the one on browser.

juliemr:
Should we instead wrap this? It seems reasonable to add it to the `WEB_ELEMENT_FUNCTIONS` and document it.

mgiambalvo:
I assumed it wasn't wrapped for a reason. But if you can't think of any reason not to wrap it, then we might as well go that way.

juliemr:
It's new, so we just haven't gotten around to it. (https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/CHANGES.md#v2482)

mgiambalvo:
Ah, I see. As near as I can tell, this isn't supported yet in Chromedriver (even in 2.21)

juliemr:
Hmm, let's wrap it anyway - if the error message is reasonable. (The command is `/session/:sessionId/element/:id/screenshot`: https://github.com/SeleniumHQ/selenium/blob/c10e8a955883f004452cdde18096d70738397788/javascript/webdriver/http/http.js#L275)

sjelin:
Closing this since instead we're deciding to wrap it

PR #3108, opened 2016-04-08
fix(protractor): isPresent() should work with out of bounds errors

sjelin:
Follow up from https://github.com/angular/protractor/pull/3056


juliemr:
LGTM once Travis is happy (sauce connect failed to start, I've restarted the job)

PR #3111, opened 2016-04-11
docs: Document that by.xpath will ignore parent if chaining

jozzhart:
docs(protractor) Mention by.xpath ignores parent context when using element chaining


juliemr:
This is now documented with the webdriver By locator here: https://github.com/angular/protractor/blob/master/lib/selenium-webdriver/locators.js#L203

PR #3114, opened 2016-04-11
chore(typescript): clean up file names and do not use export default

cnishina:


cnishina:
Plan to clean up "export default" in a different pull request.

PR #3024, opened 2016-03-14
chore(tests): test for ngUpgrade support

sjelin:
`upgrade.js` taken from https://github.com/angular/angular/pull/7603

Issue #3000, opened 2016-03-04
Test spec/ng2/async_spec.js occasionally fails


cnishina:
The test: "should wait for a series of periodic increments" has been failing occasionally on Travis and should be fixed. The test clicks the action button, waits, then clicks cancel.

The text for the value should be the same after 3 seconds. However, this is resulting in a timeout... or mismatch in values. Below is the failure for timing out.

```
[launcher] Running 1 instances of WebDriver
Started
....F
Failures:
1) async angular2 application long async spec should wait for a series of periodic increments
  Message:
    Failed: Wait timed out after 4388ms
  Stack:
    Error: Wait timed out after 4388ms
        at /home/travis/build/angular/protractor/node_modules/selenium-webdriver/lib/promise.js:2241:22
        at Promise.invokeCallback_ (/home/travis/build/angular/protractor/node_modules/selenium-webdriver/lib/promise.js:1329:14)
        at TaskQueue.execute_ (/home/travis/build/angular/protractor/node_modules/selenium-webdriver/lib/promise.js:2790:14)
        at TaskQueue.executeNext_ (/home/travis/build/angular/protractor/node_modules/selenium-webdriver/lib/promise.js:2773:21)
        at /home/travis/build/angular/protractor/node_modules/selenium-webdriver/lib/promise.js:2652:27
        at /home/travis/build/angular/protractor/node_modules/selenium-webdriver/lib/promise.js:639:7
        at process._tickCallback (node.js:364:9)
    From: Task: <anonymous wait>
        at ControlFlow.wait (/home/travis/build/angular/protractor/node_modules/selenium-webdriver/lib/promise.js:2229:17)
        at WebDriver.wait (/home/travis/build/angular/protractor/node_modules/selenium-webdriver/lib/webdriver.js:688:29)
        at [object Object].to.(anonymous function) [as wait] (/home/travis/build/angular/protractor/built/protractor.js:68:25)
```

cnishina:
https://github.com/angular/protractor/blob/9144494a28dac5a0409de4c5384e933f2d2f8156/spec/ng2/async_spec.js#L58

cnishina:
```
describe('long async spec', function() {
    var originalTimeout;
    beforeEach(function() {
      originalTimeout = jasmine.DEFAULT_TIMEOUT_INTERVAL;
      jasmine.DEFAULT_TIMEOUT_INTERVAL = 30000;
    });

    it('should wait for a series of periodic increments', function() {
[...]

PR #3121, opened 2016-04-12
chore(testapp): Use newer angular beta

sjelin:
Need this for #3024


cnishina:
This looks fine except this definitely makes the flaky test #3000 fail 100% of the time. This could might need some attention.

cnishina:
Should probably get #3127 merged in first to pass this test. Other than that, LGTM

sjelin:
Blocked on https://github.com/angular/angular/issues/8089.  Assigning Julie since that bug is much more her domain.

juliemr:
See https://github.com/angular/angular/issues/8219

We decided this is not blocked, since we should be running `setTimeout` outside of the Angular zone.

sjelin:
`setInterval` is now being run outside of NgZone

sjelin:
Craig assigning back to you since you're on call

PR #3131, opened 2016-04-14
feat(logger): improve logging methods

cnishina:
- add log levels info, debug, warn, error
- add logging to file, console, both, or none
- add color and some log formatting
- fix error logging

### DIFF
--- a/lib/logger2.ts
+++ b/lib/logger2.ts
@@ -1,0 +1,269 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import {Config} from './configParser';
+
+// Will use chalk if chalk is available to add color to console logging
+let chalk: any;
+let printRed: Function;
+let printYellow: Function;
+let printGray: Function;
+
+try {
+  chalk = require('chalk');
+  printRed = chalk.red;
+  printYellow = chalk.yellow;
+  printGray = chalk.gray;
+} catch (e) {
+  printRed = printYellow = printGray = (msg: any) => { return msg; };
+}
+
+export enum LogLevel {
+  ERROR,
+  WARN,
+  INFO,
+  DEBUG
+}
+
+export enum WriteTo {
+  CONSOLE,
+  FILE,
+  BOTH,
+  NONE
+}
+
+let logFile = 'protractor.log';  // the default log file name
+
+/**
+ * Logger class adds timestamp output, log levels, and identifiers to help
+ * when debugging. Also could write to console, file, both, or none.
+ */
+export class Logger {
+  static logLevel: LogLevel = LogLevel.INFO;
+  static showTimestamp: boolean = true;
+  static showId: boolean = true;
+  static writeTo: WriteTo = WriteTo.CONSOLE;
+  static fd: any;
+  static firstWrite: boolean = false;
+
+  /**
+   * Set up the logging configuration from the protractor configuration file.
+   * @param config The protractor configuration
+   */
+  static set(config: Config): void {
+    if (config.troubleshoot) {
+      Logger.logLevel = LogLevel.DEBUG;
+    }
+  }
+
+  /**
+   * Set up the write location. If writing to a file, get the file descriptor.
+   * @param writeTo The enum for where to write the logs.
+   * @param opt_logFile An optional parameter to override the log file location.
+   */
+  static setWrite(writeTo: WriteTo, opt_logFile?: string): void {
+    if (opt_logFile) {
+      logFile = opt_logFile;
+    }
+    Logger.writeTo = writeTo;
+    if (Logger.writeTo == WriteTo.FILE || Logger.writeTo == WriteTo.BOTH) {
+      Logger.fd = fs.openSync(path.resolve(logFile), 'a');
+      Logger.firstWrite = false;
+    }
+  }
+
+  /**
+   * Creates a logger instance with an ID for the logger.
+   * @constructor
+   */
+  constructor(private id: string) {}
+
+  /**
+   * Log INFO
+   * @param ...msgs multiple arguments to be logged.
+   */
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -13,9 +13,10 @@
   "author": "Julie Ralph <ju.ralph@gmail.com>",
   "dependencies": {
     "adm-zip": "0.4.7",
+    "chalk": "^1.1.3",
```
juliemr: Chalk has a non-trivial dependency tree:

`
└─┬ chalk@1.1.3
  ├── ansi-styles@2.2.1
  ├── escape-string-regexp@1.0.5
  ├─┬ has-ansi@2.0.0
  │ └── ansi-regex@2.0.0
  ├── strip-ansi@3.0.1
  └── supports-color@2.0.0
`

which is a pain for import. Can we make it optional? Just something like trying to require, catching an error and just returning a plain string if not available.

spec/unit/logger_test.js
```diff
@@ -0,0 +1,133 @@
+var Logger = require('../../built/logger2').Logger;
+var WriteTo = require('../../built/logger2').WriteTo;
+var LogLevel = require('../../built/logger2').LogLevel;
+var path = require('path');
+var fs = require('fs');
+
+describe('the logger', function() {
+  var logFile = 'logger_test.log';
```
juliemr: Put this in `os.tmpdir()`

spec/unit/logger_test.js
```diff
@@ -0,0 +1,133 @@
...
+    try { fs.unlinkSync(path.resolve(logFile)); } catch(e) { }
+  });
+
+  afterEach(function() {
+    try { fs.unlinkSync(path.resolve(logFile)); } catch(e) { }
+  });
+
+  describe('log strings to file', function() {
+    beforeEach(function() {
+      try { fs.unlinkSync(path.resolve(logFile)); } catch(e) { }
```
juliemr: You can remove this line, it's redundant with the higher level beforeEach

spec/unit/logger_test.js
```diff
@@ -0,0 +1,133 @@
...
+      logger.debug('hello debug');
+      logger.info('hello info');
+      logger.warn('hello warn');
+      logger.error('hello error');
+    };
+
+    afterEach(function() {
+      Logger.setWrite(WriteTo.CONSOLE);
+      Logger.logLevel = LogLevel.DEBUG;
+      try { fs.unlinkSync(path.resolve(logFile)); } catch(e) { }
```
juliemr: Ditto.

spec/unit/logger_test.js
```diff
@@ -0,0 +1,133 @@
...
+      var lines = fs.readFileSync(path.resolve(logFile)).toString();
+      var linesSplit = lines.split('\n');
+      expect(linesSplit.length).toBe(3);
+      expect(linesSplit[1]).toContain('E/FooBar');
+    });
+  });
+
+  describe('log json objects/array to file', function() {
+    beforeEach(function() {
+      try { fs.unlinkSync(path.resolve(logFile)); } catch(e) { }
```
juliemr: Ditto, and below.

package.json
cnishina: Added a try catch for chalk. If chalk is not available, it will not use chalk logging features.

spec/unit/logger_test.js
cnishina: Fixed. Using `path.resolve(os.tmpdir(), 'logger_test.log');`

spec/unit/logger_test.js
cnishina: Oh cool... removed.

spec/unit/logger_test.js
cnishina: Fixed.

spec/unit/logger_test.js
cnishina: Ditto x 2.

lib/logger2.ts
```diff
@@ -0,0 +1,199 @@
...
+        if (logLevel <= LogLevel.DEBUG) {
+          this.print_(logLevel, msgs);
+        }
+        break;
+      default:
+        throw new Error('Log level undefined');
+    }
+  }
+
+  print_(logLevel: LogLevel, msgs: any[]): void {
```
juliemr: What's the point of this fn?

lib/logger2.ts
```diff
@@ -0,0 +1,199 @@
...
+    this.printBuffer_(logLevel, msgs);
+  }
+
+  printBuffer_(logLevel: LogLevel, msgs: any[]): void {
+    let consoleLog: string = '';
+    let fileLog: string = '';
+
+    if (Logger.showTimestamp) {
+      consoleLog += Logger.timestamp_(true);
```
juliemr: Hm, I don't like this calling the same thing with true/false booleans. But I can't think of a better way to do it. Formatting is hard :(

lib/logger2.ts
```diff
@@ -0,0 +1,199 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import {Config} from './configParser';
+
+let useChalk = false;
```
juliemr: What about consolidating all the if statements to up here:

var printRed();
var printYellow();
var printGrey(); 
```js
try {
  chalk = require('chalk');
  printRed = chalk.red;
  // etc
} catch (e) {
  printRed = printYellow = printGrey = function(msg) { return msg; };
}
```

juliemr:
LGTM after cleanup.

lib/logger2.ts
cnishina: Wow, that's embarrassing....

cnishina:
Note to self: Needs comments...

cnishina:
Added comments.

PR #3133, opened 2016-04-15
chore(exitCodes): adding exit code for browser connect errors

cnishina:


cnishina:
Unit tests added for BrowserError for direct connect.

cnishina:
depends on #3131

spec/unit/driverProviders/direct_test.js
```diff
@@ -0,0 +1,69 @@
+
+var path = require('path');
+var fs = require('fs');
+var BrowserError = require('../../../built/exitCodes').BrowserError;
+
+var removeSeleniumFolder = function() {
```
juliemr: Instead of doing this, just use a config file that points at a different (and invalid) chromedriver location.

cnishina:
Instead of using the `selenium` folder, creating a "chromedriver" file in the `os.tmpdir()` for my unit test.

spec/unit/driverProviders/direct_test.js
```diff
@@ -0,0 +1,69 @@
...
+        expect(e.code).toBe(BrowserError.CODE);
+      }
+      expect(errorFound).toBe(true);
+    });
+  });
+
+  describe('with selenium drivers', function() {
+    var chromedriver = '';
+    beforeEach(function() {
+      // add files to selenium folder
```
juliemr: Why is it important that the file have the correct version? Aren't we just testing that if the file exists but isn't a proper driver, we get the right error code?

I think this should either just use a bogus file that we create (like os.tpdir() + '/foo/bar') or not specify `chromeDriver` in the config and let the driverProvider use its normal logic.

lib/driverProviders/browserStack.ts
```diff
@@ -3,10 +3,11 @@
  * It is responsible for setting up the account object, tearing
  * it down, and setting up the driver correctly.
  */
-import * as request from 'request';
+import * as https from 'https';
```
juliemr: Can this 'request' removal be split into its own PR?

lib/exitCodes.ts
```diff
@@ -1,28 +1,79 @@
 import {Logger} from './logger2';
 
 const CONFIG_ERROR_CODE = 105;
+const BROWSER_ERROR_CODE = 135;
```
juliemr: I think `BROWSER_ERROR` is too ambiguous - let's make this `BROWSER_CONNECT_ERROR_CODE`. It's long but our users shouldn't need to type it.

lib/exitCodes.ts
```diff
@@ -1,28 +1,79 @@
...
+  static DEFAULT_MSG = 'protractor encountered an error';
 
-export class ProtractorError {
   error: Error;
   description: string;
   code: number;
   stack: string;
   constructor(logger: Logger, description: string, code: number) {
+    super();
     this.error = new Error();
```
juliemr: We have both inheritance and composition - if we're inheriting from error can we just use `this.stack`?

juliemr: Or, not inherit?

lib/exitCodes.ts
```diff
@@ -1,28 +1,79 @@
...
 /**
  * Configuration file error
  */
 export class ConfigError extends ProtractorError {
+  static DEFAULT_MSG = 'configuration error';
   static CODE = CONFIG_ERROR_CODE;
-  constructor(logger: Logger, description: string) {
-    super(logger, description, ConfigError.CODE);
+  constructor(logger: Logger, opt_msg?: string) {
+    super(logger, opt_msg || ConfigError.DEFAULT_MSG, ConfigError.CODE);
```
juliemr: Why have a default configuration msg? 'configuration error' is pretty useless by itself, I'd say keep this required.

lib/exitCodes.ts
```diff
@@ -1,28 +1,79 @@
...
+    process.exit(ConfigError.CODE);
+  }
+}
+
+/**
+ * Browser errors including getting a driver session, direct connect, etc.
+ */
+export class BrowserError extends ProtractorError {
+  static DEFAULT_MSG = 'browser error';
```
juliemr: Ditto for not having a default error, make the error-thrower say something useful.

lib/launcher.ts
```diff
@@ -176,7 +177,18 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
+          if (errorCode) {
+            logger.error(e.stack);
+            process.exit(errorCode);
+          } else {
+            logger.error(e.stack);
+          }
+        });
+
         process.on('exit', (code: number) => {
+          console.log(code);
```
juliemr: Should we be logging this? If someone wants to use it, they can just use the exit code.

juliemr: And it's noted below.

juliemr:
A couple clean-up comments.

lib/driverProviders/browserStack.ts
cnishina: Done.

lib/exitCodes.ts
cnishina: Removing the default message and plan to use the message from the thrown error.

lib/exitCodes.ts
cnishina: This was a weird work around because `stack` is not being part of `node` typings. Using `captureStackTrace` with `any` to get the stack.

lib/exitCodes.ts
cnishina: Done

lib/exitCodes.ts
cnishina: Removed.

lib/launcher.ts
cnishina: Removed.

lib/ptor.ts
```diff
@@ -2,7 +2,11 @@ import {Browser, ElementHelper} from './browser';
 import {ProtractorBy} from './locators';
```
juliemr: This looks like an unrelated change.

cnishina: Yes, I need to stop sneaking these in. Will revert this file back.

juliemr:
LGTM once that's done.

PR #3142, opened 2016-04-20
chore(typescript): driverProviders converted to typescript

cnishina:


juliemr:
Travis errors are legitimate.

lib/globals.d.ts
```diff
@@ -1 +1,11 @@
 declare var browser: any;
+
+declare class WebDriver {
```
juliemr: I'm in favor of everything that's required from WebDriver being put in a namespace.

typings.json
```diff
@@ -1,9 +1,11 @@
 {
   "ambientDependencies": {
     "chalk": "registry:dt/chalk#0.4.0+20160317120654",
+    "form-data": "registry:dt/form-data#0.0.0+20160316155526",
```
juliemr: What are the form-data typings for?

cnishina: When installing `request` it wanted to install `form-data`. It is on my radar to remove `request`...

lib/globals.d.ts
cnishina: Yup, after seeing your pr... I am going to change mine to match.

cnishina:
#1203

juliemr:
Looks good! Merge it.

PR #3149, opened 2016-04-21
chore(typescript): runner and launcher converted to typescript

cnishina:


lib/globals.d.ts
```diff
@@ -1,4 +1,22 @@
+// This will clash with the node 'global' namespace...
```
juliemr: Does this comment need resolving?

juliemr: Note - if it doesn't require action, it should be a bit more verbose, because I'm not sure what it's trying to tell me.

lib/launcher.ts
```diff
@@ -0,0 +1,283 @@
...
+let logger = new Logger('launcher');
+let RUNNERS_FAILED_EXIT_CODE = 100;
+
+/**
+ * Keeps track of a list of task results. Provides method to add a new
+ * result, aggregate the results into a summary, count failures,
+ * and save results into a JSON file.
+ */
+class TaskResults {
+  results_: any[] = [];
```
juliemr: Maybe for a follow-up CL, but it might make sense to introduce a class type for results.

lib/launcher.ts
```diff
@@ -0,0 +1,283 @@
...
+                exitCode = 1;
+              }
+              return cleanUpAndExit(exitCode);
+            })
+            .done();
+      })
+      .done();
+};
+
+exports.init = init;
```
juliemr: can we do `export var init` instead?

lib/runner.ts
```diff
@@ -0,0 +1,380 @@
...
+    }
+  }
+
+  /**
+   * Responsible for cleaning up test run and exiting the process.
+   * @private
+   * @param {int} Standard unix exit code
+   */
+  exit_ = function(exitCode: number):
+      any {
```
juliemr: This indentation seems bad but I guess if that's what clang wants...

lib/runner.ts
```diff
@@ -0,0 +1,380 @@
...
+        this.driverprovider_.quitDriver.bind(this.driverprovider_)));
+  }
+
+  /**
+   * The primary workhorse interface. Kicks off the test running process.
+   *
+   * @return {q.Promise} A promise which resolves to the exit code of the tests.
+   * @public
+   */
+  run() {
```
juliemr: Add return type

lib/globals.d.ts
cnishina: This does not require any immediate action but the tsc command will complain about global namespace being declared twice.

lib/launcher.ts
cnishina: Adding a TODO

lib/launcher.ts
cnishina: Done.

lib/runner.ts
cnishina: I think clang is wrong on this one but this is what clang wants.

lib/runner.ts
cnishina: Done.

juliemr:
LGTM

PR #3151, opened 2016-04-22
chore(typescript): convert cli.js to cli.ts

sjelin:


lib/cli.ts
```diff
@@ -0,0 +1,159 @@
...
+    .alias('name', 'capabilities.name')
+    .alias('platform', 'capabilities.platform')
+    .alias('platform-version', 'capabilities.version')
+    .alias('tags', 'capabilities.tags')
+    .alias('build', 'capabilities.build')
+    .alias('grep', 'jasmineNodeOpts.grep')
+    .alias('invert-grep', 'jasmineNodeOpts.invertGrep')
+    .alias('explorer', 'elementExplorer')
+    .string('capabilities.tunnel-identifier')
+    .check(function(arg: ProtractorArgs) {
```
juliemr: I don't think this type is right - this is the leftover args, which is just an array of strings.

lib/cli.ts
```diff
@@ -0,0 +1,159 @@
...
+      if (arg._.length > 1) {
+        throw 'Error: more than one config file specified';
+      }
+    });
+
+var argv: ProtractorArgs = optimist.parse(args);
```
juliemr: I'm not sure typing helps here, since this is an arbitrary object (since the user could include any flags)

lib/cli.ts
```diff
@@ -0,0 +1,159 @@
...
+if (!configFile && !argv.elementExplorer && args.length < 3) {
+  console.log(
+      '**you must either specify a configuration file ' +
+      'or at least 3 options. See below for the options:\n');
+  optimist.showHelp();
+  process.exit(1);
+}
+
+// Run the launcher
+require('./launcher').init(configFile, argv);
```
juliemr: Change to:
```js
import {init} from './launcher';
init(configFile, argv);
```

lib/cli.ts
sjelin: Would you prefer `any` then?

lib/cli.ts
sjelin: It definitely isn't an array of strings.  I didn't change the `if` statement on the line below.  Maybe it's just `{_: string[]}` though

lib/cli.ts
juliemr: Yup.

lib/cli.ts
juliemr: My bad, I missed that. Yes, you're right. But if we're removing the type below, doesn't really matter.

lib/cli.ts
```diff
@@ -0,0 +1,149 @@
...
+if (!configFile && !argv.elementExplorer && args.length < 3) {
+  console.log(
+      '**you must either specify a configuration file ' +
+      'or at least 3 options. See below for the options:\n');
+  optimist.showHelp();
+  process.exit(1);
+}
+
+// Run the launcher
+import * as launcher from './launcher';
```
juliemr: Maybe we should move the import to the top of the file.

PR #3152, opened 2016-04-22
chore(typescript): migrate protractor.js to TypeScript.

mgiambalvo:
This isn't done yet, but it's a big one so I'd like some early feedback.


mgiambalvo:
Oh and I'm pretty sure I broke a couple things in the process, so I need to fix tests as well.

lib/globals.d.ts
```diff
@@ -1,12 +1,28 @@
...
+    navigate: Function;
+    get: Function;
+    wait: Function;
+    schedule: Function;
+    controlFlow: Function;
+    findElements: Function;
     static attachToSession: Function;
   }
 
+  class Locator {
```
juliemr: does it make sense to get the locator file migrated before this one?

lib/protractor.ts
```diff
@@ -0,0 +1,1208 @@
...
+ *
+ * @private
+ * @param {!(string|Function)} script The script to execute.
+ * @param {string} description A description of the command for debugging.
+ * @param {...*} var_args The arguments to pass to the script.
+ * @return {!webdriver.promise.Promise.<T>} A promise that will resolve to the
+ *    scripts return value.
+ * @template T
+ */
+  // TODO add varargs parameter
```
juliemr: Can remove this todo now, right?

lib/protractor.ts
```diff
@@ -0,0 +1,1208 @@
...
+   * });
+   *
+   * @param {!string} name The name of the module to load or override.
+   * @param {!string|Function} script The JavaScript to load the module.
+   *     Note that this will be executed in the browser context, so it cannot
+   *     access variables from outside its scope.
+   * @param {...*} varArgs Any additional arguments will be provided to
+   *     the script and may be referenced using the `arguments` object.
+   */
+  addMockModule(name: string, script: string|Function, ...moduleArgs: Array<any>) {
```
juliemr: Prefer `any[]` to `Array<any>` here and throughout.

lib/protractor.ts
```diff
@@ -0,0 +1,1208 @@
...
+ * @private
+ * @param {!(string|Function)} script The script to execute.
+ * @param {string} description A description for debugging purposes.
+ * @param {...*} var_args The arguments to pass to the script.
+ * @return {!webdriver.promise.Promise.<T>} A promise that will resolve to the
+ *    scripts return value.
+ * @template T
+ */
+  // TODO add varargs parameter
+  private executeAsyncScript_(script: string|Function, description: string):
```
juliemr: I think we need to add another `...args` here

lib/protractor.ts
```diff
@@ -0,0 +1,1208 @@
...
+      ...args: Array<any>): webdriver.Promise {
+    if (typeof script === 'function') {
+      script = 'return (' + script + ').apply(null, arguments);';
+    }
+
+    // TODO use args instead of arguments
+    return this.driver.schedule(
+        new Command(CommandName.EXECUTE_SCRIPT)
+            .setParameter('script', script)
+            .setParameter('args', Array.prototype.slice.call(arguments, 2)),
```
juliemr: This can be `args` instead of slicing `arguments`

lib/protractor.ts
```diff
@@ -0,0 +1,1208 @@
...
+
+/**
+ * @type {ProtractorBy}
+ */
+export let By = new ProtractorBy();
+
+/**
+ * @type {ElementFinder}
+ */
+export {ElementFinder} from './element';
```
juliemr: I think we can do `export {ElementFinder, ElementArrayFinder} from './element';`

juliemr:
Some early comments - overall strategy looks OK once we get the wrapping working.

lib/globals.d.ts
mgiambalvo: That might be helpful.

lib/protractor.ts
mgiambalvo: What should the JSDoc be like if we did that? Or does it matter?

juliemr: I don't _think_ we use these parsed types in our docs.

mgiambalvo:
Please take a look. This is mostly ready, though I still need to fix the interactive tests.

juliemr:
LGTM once things are passing.

PR #3153, opened 2016-04-22
chore(typescript): convert locators.js to locators.ts

sjelin:


lib/locators.ts
```diff
@@ -0,0 +1,462 @@
...
+let clientSideScripts: any = require('./clientsidescripts');
+
+/**
+ * webdriver's By is an enum of locator functions, so we must set it to
+ * a prototype before inheriting from it.
+ */
+export class WebdriverBy {};
+WebdriverBy.prototype = webdriver.By;
+
+export interface ProtractorLocator extends webdriver.Locator {
```
juliemr: Since the webdriver Locator is namespaced, can we just `export interface Locator`? There should be no ambiguity since it would be imported from protractor.

lib/locators.ts
```diff
@@ -0,0 +1,462 @@
...
+        toString: function toString() {
+          return 'by.' + name + '("' +
+              Array.prototype.join.call(locatorArguments, '", "') + '")';
+        }
+      };
+    };
+  };
+
+  /**
+ * Find an element by text binding. Does a partial match, so any elements bound
```
juliemr: Indent +2 for the rest of the comment here.

lib/locators.ts
```diff
@@ -0,0 +1,462 @@
...
+            rootSelector));
+      },
+      toString: function toString() {
+        return 'by.binding("' + bindingDescriptor + '")';
+      }
+    };
+  };
+
+  /**
+ * Find an element by exact binding.
```
juliemr: And indent +2 here and elsewhere.

lib/locators.ts
sjelin: You can never trust clang

sjelin:
All comments addressed.  Will merge as soon as travis is done

Issue #3154, opened 2016-04-22
Element ids starting with a digit followed by a valid hex-digit are not found with by.id()


arlowhite:
Bug report
---

- Node Version: `5.10.1`
- Protractor Version: `3.2.2`
- Browser(s): `Chrome, Firefox` (probably all browsers)
- OS: `Linux (Ubuntu 15.10)`
- A relevant example test

`element(by.id('1234')).isPresent()` does not find the element with id="1234"

### Fixed in **selenium-webdriver** project

This was just fixed April 1st, this bug will be resolved by upgrading to a Selinium Webdriver release with this commit:
https://github.com/SeleniumHQ/selenium/commit/c554cf3b7ee1302a5ad4e830e6cdc4e418906287

## Current behavior

An id such as "1234" will fail to be located with by.id(). However, Protractor is correctly escaping the first digit with \3, just not adding a space after the first escaped digit. This can be seen in the webdriver log:

> INFO - Executing: [find elements: By.cssSelector: *[id="\31234"]])

(There should be a space between \31 and 234)

This means that Protractor will only find digit leading ids if the 2nd character is not a hex digit, such as "1z234" (Where both \31z234 and \31 z234 select this id)

I'm not sure where the fix should occur, maybe in one of Protractor's dependencies?

## Background

**id** attributes starting with a number are [legal in HTML5](https://www.w3.org/TR/html5/dom.html#the-id-attribute), but not HTML4.
However, in CSS the first digit must be escaped.
https://www.w3.org/TR/CSS21/syndata.html#characters
Also, note this [StackOverflow discussion](http://stackoverflow.com/questions/20306204/using-queryselector-with-ids-that-are-numbers)

arlowhite:
I think this is a bug in selenium-webdriver, so the fix will be to bump the webdriver version once there is a fix upstream. I will work on a bug report for the Selenium project.

arlowhite:
Ok, this was just fixed April 1 in selenium-webdriver
https://github.com/SeleniumHQ/selenium/commit/c554cf3b7ee1302a5ad4e830e6cdc4e418906287

So this bug will be resolved by updating selenium-webdriver

juliemr:
Thank you for your detailed report! Well written issues like this make me happy :)

We'll be updating our dependency after we do a stable release with our TS and website changes!

juliemr:
This should now be fixed!

PR #3155, opened 2016-04-22
Add elements_spec.js test for numeric ids. element(by.id('1234'))

arlowhite:
See #3154


arlowhite:
I signed it!

arlowhite:
elements_spec.js is expected to fail until bug #3154 is fixed.

I added the test to index.html#form because that seemed the most appropriate area without creating a new section.

arlowhite:
This test belongs in selenium-webdriver, not Protractor. And the bug was just fixed April 1st:
https://github.com/SeleniumHQ/selenium/commit/c554cf3b7ee1302a5ad4e830e6cdc4e418906287

PR #3156, opened 2016-04-23
chore(typescript): make plugings work in ts

sjelin:
Make sure to read "Writing Plugins in TypeScript" because I want feedback on that


docs/plugins.md
```diff
@@ -303,6 +304,42 @@ exports.addWarning(message, options);
...
+import {ProtractorPlugin} from 'protractor/built/plugins';
+
+export class MyPlugin extends ProtractorPlugin {
+  onPageLoad() {
+    this.addSuccess({specName: 'Hello, World!'});
+  };
+};
+```
+
+Then, in a separate file, export an instance of that class using an assignment
```
NickTomlin: I'm not knowledgable about typescript, but is there a reason this needs to be split into two files? Can't you just `export = new MyPlugin()` within the file that defines `MyPlugin`?

sjelin: You can't use `export =` if you are also exporting `MyPlugin`, and you can't export `new MyPlugin()` alone because it's class is private (or you'll get errors at least).

Issue #3167, opened 2016-04-25
Deps - upgrade selenium-webdriver + friends to 2.53


juliemr:
selenium-webdriver now has v2.53.1

And selenium standalone + IE driver have v 2.53.0.

cnishina:
selenium standalone = 2.53.0 and iedriver = 2.53.1.

juliemr:
Ah thanks!

cnishina:
selenium webdriver 2.53.2 is released. Time to upgrade the `package.json`.

PR #3172, opened 2016-04-26
deps: Updating selenium-webdriver + selenium + iedriver

pedrro:
Updating versions of the following dependencies:

```
Selenium-webdriver : 2.53.1
selenium: 2.53.0
iedriver: 2.53.1
```
Referenced on #3167, #3154

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,7 +1,7 @@
 {
   "webdriverVersions": {
-    "selenium": "2.52.0",
+    "selenium": "2.53.0",
     "chromedriver": "2.21",
-    "iedriver": "2.52.0"
+    "iedriver": "2.53.1"
   }
 }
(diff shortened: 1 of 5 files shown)


pedrro:
I signed it!

cnishina:
The upgrade to selenium-webdriver 2.53.1 passes all tests except for attaching a session.

#3173

juliemr:
@cnishina we should also change the versions that we test against on SauceLabs in https://github.com/angular/protractor/blob/master/spec/ciFullConf.js and friends.

pedrro:
@juliemr The version also should be updated on ciSmokeConf.js and ciNg2Conf.js? I can do this updates

juliemr:
Yup! That'd be great.

PR #3189, opened 2016-04-29
Chore(conf): Refactor npm tests to run the tests in a unique task.

pedrro:
Hey guys, I made some refactor on how the tests are running with npm/gulp.
Basically, before this PR, to run the tests the user needs to run manually three tasks:

- bin/webdriver-manager start
- npm start
- npm test

So, I change this putting the **npm start** as dependency of **gulp pretest** and also **bin/webdriver-manager start** is a dependency task of **npm start**. With these changes, the user just need to run the task **npm test** to run the tests on your local environment. 

Another important thing is that I put the **webdriver-manager start** to run on background.

Please, give me some feedbacks about this change and how I can improve it. 👍

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
   "scripts": {
     "prepublish": "gulp prepublish",
     "pretest": "gulp pretest",
+    "prestart": "webdriver-manager start > /dev/null 2>&1 &",
     "start": "node testapp/scripts/web-server.js",
     "test": "node scripts/test.js",
     "tsc": "./node_modules/typescript/bin/tsc",
--- a/README.md
+++ b/README.md
@@ -45,14 +45,10 @@ Clone the github repository:
     npm install
     cd ..
 
-Start up a selenium server. By default, the tests expect the selenium server to be running at `http://localhost:4444/wd/hub`. A selenium server can be started with `webdriver-manager`.
-
-    bin/webdriver-manager start
-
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -54,6 +54,7 @@
   "scripts": {
     "prepublish": "gulp prepublish",
     "pretest": "gulp pretest",
+    "prestart": "webdriver-manager start > /dev/null 2>&1 &",
```
NickTomlin: Input redirection and backgrounding isn't going to work on windows unfortunately :(

pedrro: Has some solution created for this?

cnishina: It might be better if we created a child process of webdriver-manager and then stopped the process when the tests are completed.

pedrro: Ok, I'll work on this ;)

NickTomlin: Sorry for the late response! Yep, I think child processes are the way to go. I hate adding more layers to this but it's probably best handled in the Gulpfile.

pedrro: Awesome! So, I think that is a better way if you close this PR, because it's showing my last commit that was already merged in another PR, so, if you close this one I'll open again with the new solution for this, Could u do that? :)

NickTomlin: Sure (you should be able to close it yourself via the UI if I remember correctly, but i'm happy to do it as well 😄 ). Thanks!

Issue #3191, opened 2016-05-02
Configuration option getMultiCapabilities does not resolve when configured as a promise.


mdasberg:
As of version 3.3.0 the option getMultiCapabilities when configured as a promise, does not resolve anymore and shuts down immediately, without running any tests.

When for instance using the following configuration:
```js
getMultiCapabilities: function () {
   return require('q').when([{
      'browserName': 'chrome',
      'shardTestFiles': true,
      'maxInstances': 10
   }]);
},
```

Remco75:
same issue here, PR seems to fix the problem

juliemr:
You're right - we didn't notice because our test for this (https://github.com/angular/protractor/blob/master/spec/getCapabilitiesConf.js) is happily doing nothing and exiting immediately.

PR #3195, opened 2016-05-02
Fix Configuration option getMultiCapabilities when configured as a promise.

mdasberg:
Fix for #3191

Since the rewrite to typescript the getMultiCapabilities promise configuration did not resolve.
This fix will resolve the promise like it used to before the rewrite


juliemr:
Hey @cnishina do you have time to merge?

PR #3206, opened 2016-05-09
docs(dgeni): doc generation with typescript files instead of javascript files

manoj9788:
Due to recent typescript file conversion, there are few number of 404 errors on links from the docs.

I am not 100% sure if this is the right place to change. I cloned this copy and ran on my local host - I could see only half of API but not the full API list say(e.g) [I was not able to see By locators of protractor, element module of protractor etc..] Hence submitting this patch using file edit from Github.

### DIFF
--- a/website/docgen/dgeni-config.js
+++ b/website/docgen/dgeni-config.js
@@ -74,7 +74,7 @@ myPackage.config(function(readFilesProcessor, templateFinder, writeFilesProcesso
     {include: 'built/element.ts'},
     {include: 'built/protractor.ts'},
     {include: 'built/locators.ts'},
-    {include: 'built/expectedConditions.js'},
+    {include: 'built/expectedConditions.ts'},
     {include: 'lib/selenium-webdriver/locators.js'},
     {include: 'lib/selenium-webdriver/webdriver.js'}
   ];
--- a/website/docgen/dgeni-config.js
+++ b/website/docgen/dgeni-config.js
@@ -71,9 +71,9 @@ myPackage.config(function(readFilesProcessor, templateFinder, writeFilesProcesso
   readFilesProcessor.basePath = path.resolve(__dirname, '../..');
 
   readFilesProcessor.sourceFiles = [
-    {include: 'built/element.js'},
-    {include: 'built/protractor.js'},
-    {include: 'built/locators.js'},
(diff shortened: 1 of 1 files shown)


cnishina:
I believe this has already been addressed in a previous commit. Were you able to build the website and resolve the links?
https://github.com/angular/protractor/commit/f149bd1e91e749a77e8ee147fdb3881584ae6851

manoj9788:
I saw the last commit. If you click on the view code button of 'by' it still lands on the 404 page error so i assume this has been missed.

manoj9788:
https://github.com/angular/protractor/blob/3.3.0/lib/locators.js

cnishina:
This has been resolved and should appear in a future website release.

cnishina:
This fix might be better than the previous fix...

NickTomlin:
@manoj9788 @cnishina i'm don't see this locally running master. I can navigate to `by` and click the "view code" button and it takes me to `https://github.com/angular/protractor/blob/3.3.0/lib/locators.ts`. Perhaps I'm missing something though 😄

manoj9788:
Thanks @cnishina 
@NickTomlin  Sorry unusre about that. But its reproducible in the docs site. Yeah might be missing something. But if you generate the docs dgeni-config.js file supplies the necessary url's to the other js files - just like anything config files :)

cnishina:
@NickTomlin's previous commit fixes the linking error. The title of this pull request is misleading.

I cloned @manoj9788's pull request and it appears that dgeni will read in the jsdoc's on a typescript file as shown in the dgeni-config.json. I might have tried it at one point but might have used the transpiled files instead. This is a good improvement and I believe this should be merged in.

@mgiambalvo Could you verify that this works and it doesn't just work on my machine?

NickTomlin:
@cnishina ahh, that makes sense. Interesting! 👍

cnishina:
Yikes, I must have been looking at the wrong branch. This doesn't appear to work. I only get the webdriver docs and none of the typescript documentation. We'll have to investigate this more.

manoj9788:
Yes, that's what  I have mentioned in the description on the this PR even I was able to see only the webDriver docs but not the other and so I was not able to validate my PR, but just raised this PR by my understanding after reading through the code base. This isn't merged ?

PR #3213, opened 2016-05-12
docs(expectedConditions): fix link to selenium's expected conditions

manoj9788:


cnishina:
This is definitely clang doing bad things "fixing" comments.

PR #3214, opened 2016-05-12
chore(typings): rename protractor to browser and add a protractor namespace

cnishina:
rename protractor to browser and add a protractor namespace

* added wrapDriver method from the browser.ts and ExpectedConditions
* added a new instance of ExpectedConditions to the protractor namespace
* imported selenium webdriver ActionSequence, Key, Command, and CommandName to the protractor namespace
* added promise object to the browser class

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -8,6 +8,7 @@ import {ProtractorBy} from './locators';
 import {ElementArrayFinder, ElementFinder, build$, build$$} from './element';
 export {ElementFinder, ElementArrayFinder};
 import {Plugins} from './plugins';
+import {protractor} from './ptor';
 
 let clientSideScripts = require('./clientsidescripts');
 
@@ -16,8 +17,6 @@ let clientSideScripts = require('./clientsidescripts');
 import * as EC from './expectedConditions';
 
 let webdriver = require('selenium-webdriver');
-let Command = require('selenium-webdriver/lib/command').Command;
-let CommandName = require('selenium-webdriver/lib/command').Name;
 
 // jshint browser: true
 /* global angular */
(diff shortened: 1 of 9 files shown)


juliemr:
We spoke briefly, but let's split the change to `Browser` into a separate PR.
I decided I'm fine calling it just `Browser`, it's clear in context and won't namespace conflict (yay typescript!)

lib/browser.ts
```diff
@@ -184,7 +185,7 @@ export class Protractor {
    *
    * @type {Plugins} Object containing plugin funtions from config.
    */
-  private plugins_: Plugins;
+  plugins_: Plugins;
```
mgiambalvo: Why are these no longer private?

cnishina: They were marked as private; however, [lib/browser.ts](https://github.com/angular/protractor/blob/9b48143f90077b2d02f70d0dff19c376bb4c6824/lib/runner.ts#L199) makes a call to a private variable. This is the same issue for all the private variables in browser. They probably should be using a Get method.

mgiambalvo: Maybe double check with Julie, but if they're not actually private then we should just drop the underscore at the end and remove the @private annotations in the jsdoc as well.

juliemr: Hm yeah, these were fake private. We don't really want them as part of our documented API really, but the runner needs to access them. I'm fine making them public for now.

spec/basic/elements_spec.js
```diff
@@ -155,7 +155,7 @@ describe('ElementFinder', function() {
 
     var invalidElement = element(by.binding('INVALID'));
     invalidElement.getText().then(function(value) {
-      successful.fulfill(true);
+      success.fulfill(true);
```
mgiambalvo: Is this right? It looks like the promise is "success" everywhere else.

mgiambalvo: It looks like the promise is "succesful" everywhere else? Is this right?

PR #3226, opened 2016-05-16
docs(changelog): fix 3.3.0 issue text

Awk34:


cnishina:
A better fix would be to change how these are being created. Possibly fix this line in the [release.md](https://github.com/angular/protractor/blob/master/release.md):
```
git log 0.0.J..HEAD --format="- ([%h](https://github.com/angular/protractor/commit/%H)) %n%w(100,2,2)%B" > /tmp/changes.txt
```

juliemr:
Agreed, we don't want to manually edit these every time.

Awk34:
Hmm, so looking into it more, the problem is that https://github.com/angular/protractor/commit/6f22d5ade48f0d97990cbe69d956da122f2f8358 's commit message body had just the right amount of characters before the issue number that the git log wrapper (`%w(100,2,2)`) put it onto a new line. What if we just took out the 100 character wrapper?

cnishina:
@juliemr Just thinking about how these are generated, this change is acceptable since the issue number appears on the second line and markdown assumes it is a header. The issue number does appears in the raw file.

This is either a one-off changelog issue or we could remove the `%w(100,2,2)`

Awk34:
@cnishina if anyone's really worried about it, this `sed` addition will fix it:

```
git log 3.2.2...3.3.0 --format="- ([%h](https://github.com/angular/protractor/commit/%H)) %n%w(100,2,2)%B" | sed 's/^\(\s*\)\#\(.*\)$/\1\\#\2/'
```

juliemr:
Hah, sorry, I didn't see the original issue. Yeah let's just merge the first part of this.

cnishina:
Using the first part of the commit. Closed with 8ca84507e7a7deb4497a7c64f2e222cec6b80a28.

PR #3237, opened 2016-05-22
[feature] Adding urlIs and urlContains to expectedConditions class

manoj9788:
Added a new feature UrlIs and UrlContains. Tests have passed and all check is done from my side.
happy to change if anything is needed.

### DIFF
--- a/spec/basic/expected_conditions_spec.js
+++ b/spec/basic/expected_conditions_spec.js
@@ -1,4 +1,5 @@
 var EC = protractor.ExpectedConditions;
+var env = require('../environment');
 
 describe('expected conditions', function() {
   beforeEach(function() {
@@ -74,7 +75,7 @@ describe('expected conditions', function() {
 
   it('should have urlIs', function() {
       var baseUrlFromSpec = browser.baseUrl;
-      expect(EC.urlIs('http://localhost:8081').call()).toBe(false);
+      expect(EC.urlIs(env.baseUrl).call()).toBe(false);
       expect(EC.urlIs(baseUrlFromSpec+'index.html#/form').call()).toBe(true);
   });
   
--- a/spec/basic/expected_conditions_spec.js
+++ b/spec/basic/expected_conditions_spec.js
@@ -66,6 +66,18 @@ describe('expected conditions', function() {
     expect(EC.titleIs('My AngularJS App').call()).toBe(true);
   });
 
+  it('should have urlContains', function() {
+      var baseUrlFromSpec = browser.baseUrl;
+      expect(EC.urlContains('/form').call()).toBe(true);
+      expect(EC.urlContains(baseUrlFromSpec+ 'index.html#/form').call()).toBe(true);
+  });
+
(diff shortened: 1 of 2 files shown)


manoj9788:
@juliemr any update on this PR ?

juliemr:
This looks good to me. We should re-think how Expected Conditions and webdriver's `until` work together and unify them at somepoint, but that's out of scope here.

@cnishina do you mind merging?

manoj9788:
@juliemr  I do some contributions to until class of selenium too along with jleyba help. 
you can involve me if required for any discussion on that.

spec/basic/expected_conditions_spec.js
```diff
@@ -66,6 +66,18 @@ describe('expected conditions', function() {
...
 
+  it('should have urlContains', function() {
+      var baseUrlFromSpec = browser.baseUrl;
+      expect(EC.urlContains('/form').call()).toBe(true);
+      expect(EC.urlContains(baseUrlFromSpec+ 'index.html#/form').call()).toBe(true);
+  });
+
+  it('should have urlIs', function() {
+      var baseUrlFromSpec = browser.baseUrl;
+      expect(EC.urlIs('http://localhost:8081').call()).toBe(false);
```
cnishina: Nit: Instead of using this url, we should use the one provided from the environment variables:
```
var env = require('../environment');
expect(EC.urlIs(env.baseUrl).call()).toBe(false);
```

manoj9788: Are we waiting for @juliemr to comment or Should I push in the commit ?

juliemr: @cnishina is right, good catch.

manoj9788: @juliemr  @cnishina 
`baseUrl: env.baseUrl + '/ng1/',` is what I am seeing in the conf file. 
So its the same I believe.
Refer: https://github.com/angular/protractor/blob/master/spec/basicConf.js#L21

cnishina: Using the `baseUrlFromSpec` is fine too. It just should not have the url as localhost. @manoj9788 please fix.

spec/basic/expected_conditions_spec.js
```diff
@@ -66,6 +67,18 @@ describe('expected conditions', function() {
     expect(EC.titleIs('My AngularJS App').call()).toBe(true);
   });
 
+  it('should have urlContains', function() {
+      var baseUrlFromSpec = browser.baseUrl;
+      expect(EC.urlContains('/form').call()).toBe(true);
```
manoj9788: @cnishina Just realized what you meant. Fixed the hardcoded url with env.baseUrl

spec/basic/expected_conditions_spec.js
manoj9788: @cnishina Just realized what you meant. Fixed the hardcoded url with env.baseUrl

PR #3247, opened 2016-05-26
chore: url of BrowserStack Session at end of test

nidhimj22:
Log url of BrowserStack Automate Session at end of tests.


cnishina:
It appears we are trying to do a GET request prior to doing a PUT request. Why is this necessary for BrowserStack?

nidhimj22:
@cnishina  The PUT request that updates the job status on BrowserStack doesn't return the public url of the session in the response. The GET request is being done to obtain the public url of the BrowserStack automate session. And hence we have 2 separate requests.

cnishina:
Sounds like a valid request. There is also a separate pull request in flight that will remove the `request` node module. See #3133.

cnishina:
`request` has been removed. Please see 68d2f7367984f0298295c87b8e1c710b57a3bfe5.

nidhimj22:
@cnishina I have updated the PR. Kindly have a look.

lib/driverProviders/browserStack.ts
```diff
@@ -43,6 +39,35 @@ export class BrowserStack extends DriverProvider {
...
+          });
+        });
+        req.end();
+        req.on('error', (e: Error) => {
+          logger.info(
+            'BrowserStack results available at ' +
+            'https://www.browserstack.com/automate');
+        });
+        var jobStatus = update.passed ? 'completed' : 'error';
+        options = {
```
cnishina: Nit: Instead of redefining `options`, the only thing we are changing is options.method. Maybe switch this to `options.method = 'PUT';`

lib/driverProviders/browserStack.ts
```diff
@@ -43,6 +39,35 @@ export class BrowserStack extends DriverProvider {
           hostname: 'www.browserstack.com',
           port: 443,
           path: '/automate/sessions/' + session.getId() + '.json',
+          method: 'GET',
+          headers: headers
+        };
+
+        var req = https.request(options, (res) => {
```
cnishina: Use `let` instead of `var`

lib/driverProviders/browserStack.ts
```diff
@@ -43,6 +39,35 @@ export class BrowserStack extends DriverProvider {
...
+            }
+          });
+        });
+        req.end();
+        req.on('error', (e: Error) => {
+          logger.info(
+            'BrowserStack results available at ' +
+            'https://www.browserstack.com/automate');
+        });
+        var jobStatus = update.passed ? 'completed' : 'error';
```
cnishina: This was previously `var` but should use `let` instead.

lib/driverProviders/browserStack.ts
```diff
@@ -43,6 +39,35 @@ export class BrowserStack extends DriverProvider {
           hostname: 'www.browserstack.com',
           port: 443,
           path: '/automate/sessions/' + session.getId() + '.json',
+          method: 'GET',
+          headers: headers
+        };
+
+        var req = https.request(options, (res) => {
+          res.on('data', (data: Buffer) => {
+            var info = JSON.parse(data.toString());
```
cnishina: Switch to `let`.

nidhimj22:
@cnishina I have made the changes.

cnishina:
LGTM.

PR #3269, opened 2016-06-14
chore(launcher): clean up multicapabilities typing error

cnishina:
fix for tsc error: "Argument of type 'Function' is not assignable to parameter"


cnishina:
@mdasberg A month ago we got this f2055181e60fd358c1764fe716af3d64fc64810b to fix the launcher; however, it is still giving typescript transpile errors. Could you try this fix out?

juliemr:
This LGTM but I'm confused about the error. Is it because webdriver typings are out of date?

cnishina:
This is an error from `tsc` when it transpiles the launcher typescript file when calling then with a resolve callback. Just double checked that the transpiled javascript file does not change with this PR but the error disappears. Merging.

PR #3271, opened 2016-06-14
chore(browser): clean up types

cnishina:
* clean up comments and clang issues
* move unnecessary exports to the browser class

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -1,21 +1,19 @@
 // Util from NodeJs
-import * as util from 'util';
-import * as url from 'url';
 import * as net from 'net';
-import * as helper from './util';
-import {Logger} from './logger2';
-import {ProtractorBy} from './locators';
+import * as url from 'url';
+import * as util from 'util';
+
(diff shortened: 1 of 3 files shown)


lib/browser.ts
```diff
@@ -100,6 +88,14 @@ function buildElementHelper(browser: Browser): ElementHelper {
...
 export class Browser {
+  /**
+   * @type {ProtractorBy}
+   */
+  static By = new ProtractorBy();
+
+  /**
+   * @type {ExpectedConditions}
+   */
   ExpectedConditions = new EC.ExpectedConditions();
```
juliemr: Should this one be static as well?

juliemr:
When we do a next release, we should probably have a breaking change note for those who are using Protractor without the runner (by directly doing `require('protractor/built/protractor')`) - could you add that to this one? Just say that it's now at `built/browser` and exports `By` and `ExpectedConditions` on the class.

lib/browser.ts
cnishina: Yup. Changed it to static.

PR #3275, opened 2016-06-16
chore(element): fix error thrown from browser to element

cnishina:


cnishina:
`npm test` is passing

lib/element.ts
```diff
@@ -1,8 +1,10 @@
 let webdriver = require('selenium-webdriver');
+let NoSuchElementError = require('selenium-webdriver/lib/error').NoSuchElementError;
 let clientSideScripts = require('./clientsidescripts');
 
 import {Logger} from './logger2';
 import {Browser} from './browser';
+import {ProtractorError} from './exitCodes';
```
juliemr: Unused (and above), remove.

lib/element.ts
```diff
@@ -1,8 +1,10 @@
 let webdriver = require('selenium-webdriver');
+let NoSuchElementError = require('selenium-webdriver/lib/error').NoSuchElementError;
```
juliemr: So, for internal reasons, any time we do `require('selenium-webdriver/lib/<anything>)` our import process for Google gets a tiny bit harder. Can we revert back to just using `webdriver.error.NoSuchElementError`?

PR #3279, opened 2016-06-17
chore(types): add function types to element and locators

cnishina:
- add 'gulp types' to output index.d.ts
- some clean up for functions to use fat arrow
- clang clean up

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -5,6 +5,9 @@ var clangFormat = require('clang-format');
 var gulpFormat = require('gulp-clang-format');
 var runSequence = require('run-sequence');
 var spawn = require('child_process').spawn;
+var fs = require('fs');
+var path = require('path');
+var glob = require('glob');
 
 var runSpawn = function(done, task, opt_arg) {
   opt_arg = typeof opt_arg !== 'undefined' ? opt_arg : [];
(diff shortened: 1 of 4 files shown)


gulpfile.js
```diff
@@ -65,3 +68,42 @@ gulp.task('pretest', function(done) {
...
+  files.forEach(file => {
+    contents += readFile(folder, file);
+  });
+
+  // add module declaration
+  contents += 'declare module "protractor" {\n';
+  contents += '  export = protractor\n';
+  contents += '}\n';
+
+  // remove files with d.ts
```
juliemr: Why this step? Do they interfere if left in?

gulpfile.js
```diff
@@ -65,3 +68,42 @@ gulp.task('pretest', function(done) {
...
+  glob.sync(folder + '/**/*.d.ts').forEach(file => {
+    fs.unlinkSync(path.resolve(file));
+  });
+
+  // write contents to 'built/index.d.ts'
+  fs.writeFileSync(outputFile, contents);
```
juliemr: Honestly, I think it's a little cleaner to just do:

`fs.writeFileSync('built/index.d.ts', contents)` and forget about keeping the file/output folder nicely in variables.

juliemr: Ah, I see that folder is used for more stuff. Nevermind.

gulpfile.js
```diff
@@ -65,3 +68,42 @@ gulp.task('pretest', function(done) {
...
+});
+
+var readFile = function(folder, file) {
```
juliemr: Can we have a less generic name? Maybe `parseTypingsFile`?

lib/browser.ts
```diff
@@ -612,8 +611,7 @@ export class Browser {
                       window['NG_PENDING_TIMEOUTS'] = {};
                     }
 
-                    let extendedTimeout: any = function() {
-                      let args = Array.prototype.slice.call(arguments);
+                    let extendedTimeout: any = (...args: any[]) => {
```
juliemr: These changes will cause issues because this code is being sent to the browser and run there. I assume this is clang format? We could move this whole module into lib/clientSideScripts.js (and probably should do that anyway).

lib/locators.ts
```diff
@@ -4,19 +4,19 @@ let clientSideScripts: any = require('./clientsidescripts');
 
 /**
  * webdriver's By is an enum of locator functions, so we must set it to
+ * +}
```
juliemr: Typo

juliemr:
Main concern is the fat-arrowing of the module that shouldn't be ES6ey. Other stuff looks reasonable.

gulpfile.js
cnishina: I thought this would be cleaner to have one d.ts file instead of many.

gulpfile.js
cnishina: Done.

lib/locators.ts
cnishina: Fixed.

gulpfile.js
juliemr: ok, sgtm.

gulpfile.js
```diff
@@ -65,3 +68,42 @@ gulp.task('pretest', function(done) {
 });
 
 gulp.task('default',['prepublish']);
+
+gulp.task('types', ['tsc'], function(done) {
```
juliemr: This task should be added to the sequence for 'prepublish' now.

juliemr:
Looks good after Travis finishes (I re-ran it after a sauce connect flake) and one more comment on the gulpfile.

gulpfile.js
cnishina: Adding the task to prepublish.

PR #3288, opened 2016-06-20
chore(types): add in types inherited from selenium-webdriver

cnishina:
- add element.ts to the gulp task for 'gulp types'
- clang clean up
- separate out selenium-webdriver types and extend them in protractor
- remove webdriver types for index.d.ts file

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -72,7 +72,7 @@ gulp.task('default',['prepublish']);
 
 gulp.task('types', function(done) {
   var folder = 'built';
-  var files = ['ptor', 'browser', 'locators', 'expectedConditions'];
+  var files = ['ptor', 'browser', 'element', 'locators', 'expectedConditions'];
   var outputFile = path.resolve(folder, 'index.d.ts');
   var contents = '';
   files.forEach(file => {
@@ -104,8 +104,29 @@ var parseTypingsFile = function(folder, file) {
       if (line.indexOf('export') !== -1) {
         line = line.replace('export', '').trim();
       }
+
+      // Remove webdriver types and plugins for now
+      line = removeTypes(line,'webdriver.ActionSequence');
+      line = removeTypes(line,'webdriver.promise.Promise');
+      line = removeTypes(line,'webdriver.util.Condition');
+      line = removeTypes(line,'webdriver.WebDriver');
+      line = removeTypes(line,'webdriver.Locator');
+      line = removeTypes(line,'webdriver.WebElement');
+      line = removeTypes(line,'Plugins');
       contents += line + '\n';
     }
+
   }
   return contents;
+}
+
+var removeTypes = function(line, webdriverType) {
+  var tempLine = line.trim();
+  if (tempLine.startsWith('/**') || tempLine.startsWith('*')) {
+    return line;
+  }
+  if (line.indexOf(webdriverType) !== -1) {
+    return line.replace(new RegExp(webdriverType,'g'), 'any');
+  }
+  return line;
 }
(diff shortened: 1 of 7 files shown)


cnishina:
Note: removing `webdriver.*` types from the `index.d.ts` file since they show up as `any`. This is because we do not define `selenium-webdriver` declarations and leaving them in will cause transpiling errors.

juliemr:
The travis failures look legitimate.

gulpfile.js
```diff
@@ -104,8 +104,29 @@ var parseTypingsFile = function(folder, file) {
...
       contents += line + '\n';
     }
+
   }
   return contents;
 }
+
+var removeTypes = function(line, webdriverType) {
+  var tempLine = line.trim();
+  if (tempLine.startsWith('/**') || tempLine.startsWith('*')) {
```
juliemr: They would usually start with ` *` for multiline comments.

lib/browser.ts
```diff
@@ -30,6 +30,23 @@ for (let foo in webdriver) {
   exports[foo] = webdriver[foo];
 }
 
+// explicitly extend selenium webdriver to define functions that
```
juliemr: Nit: comments should start with uppercase, end with period.

cnishina:
It looks like I went a little overboard cleaning up the ptor file.

gulpfile.js
cnishina: This should be fine because I am trimming the line prior to testing if it starts with a `*`. This should capture the comments in the index.d.ts file.

juliemr: Right you are.

lib/element.ts
```diff
@@ -14,6 +14,40 @@ let WEB_ELEMENT_FUNCTIONS = [
   'serialize', 'takeScreenshot'
 ];
 
+// Explicitly extend selenium webdriver for webdriver.WebElement.
+export class WebdriverWebElement {
```
juliemr: Would it make more sense for these to be interfaces instead of classes?

lib/globals.d.ts
```diff
@@ -61,12 +64,14 @@ declare namespace webdriver {
   }
 
   namespace promise {
-    interface Promise {
-      controlFlow: Function,
-      then: Function
-    }
+  interface Promise {
+    controlFlow: Function, then: Function
```
juliemr: Should this be `controlFlow: Function; then: Function;`?

lib/ptor.ts
```diff
@@ -1,39 +1,40 @@
...
-    fullyResolved: require('selenium-webdriver/lib/promise').fullyResolved,
-    isPromise: require('selenium-webdriver/lib/promise').isPromise,
-    rejected: require('selenium-webdriver/lib/promise').rejected,
-    thenFinally: require('selenium-webdriver/lib/promise').thenFinally,
-    when: require('selenium-webdriver/lib/promise').when
-  };
-  export let ActionSequence = require('selenium-webdriver/lib/actions').ActionSequence;
-  export let Key = require('selenium-webdriver/lib/input').Key;
+export namespace protractor {
+export let browser: Browser;
```
juliemr: looks like indentation got squashed here?

lib/globals.d.ts
cnishina: Interfaces are comma separated. The change here was to keep clang happy.

cnishina: Just kidding... interfaces can have comma separated. Changing to semi colon.

lib/element.ts
cnishina: Changing this to be an interface.

lib/ptor.ts
cnishina: This is clang... we could investigate removing it and using tslint instead. Clang is doing weird things.

juliemr: Eh, whatever, I'm OK with this one.

PR #3297, opened 2016-06-21
chore(test): add tests for declaration type file

cnishina:


scripts/typings/test_typings.js
```diff
@@ -0,0 +1,78 @@
...
+      path.resolve('scripts/typings', file)]);
+  var output = spawned.output.toString();
+  var lines = output.split('\n');
+  var pos = 0;
+  var pass = 0;
+
+  for (var currentError = start; currentError <= end; currentError++) {
+    var line = lines[pos];
+    if (line.indexOf(file + '(' + currentError) >= 0) {
+      pass++;
```
juliemr: pass and pos see unused?

scripts/typings/test_typings.js
```diff
@@ -0,0 +1,78 @@
+var spawnSync = require('child_process').spawnSync;
```
juliemr: Would be great to have an overview comment here about what this test is doing and why we need it.

scripts/typings/test_typings.js
cnishina: I was using `pass` previously and I am no longer using it. Removing it.

`pos` is still used to advance through the lines.

cnishina:
Changing the folder name since `typings` to `typings_tests` since `typings` is ignored in `.gitignore`

scripts/typings_tests/test_typings.js
```diff
@@ -0,0 +1,91 @@
...
+ * Test:
+ *   - Files include reference paths to the Protractor's generated typings file.
+ *   - Each file is transpiled using:
+ *        tsc {fileName}
+ *   - Files that follow the type declarations should produce no errors.
+ *   - Files that do not follow the type declarations should produce errors.
+ *     For error files, it is assumed that the format matches the following:
+ *        {fileName}({lineNumber}
+ *     and should have an error for each line from starting line using the types
+ *     to the ending line of the file.
```
juliemr: One more thing before merge - can you add WHY we need these? e.g. why we don't just trust that typescript works because we have these globals and ugly selenium wrapping.

juliemr:
Looks good with one note.

PR #3300, opened 2016-06-22
chore(testapp): upgrade to angular2 rc3

cnishina:


cnishina:
WIP note:

need to resolve testapp `GET /ng2/async`. The website works; however, navigating directly to `/ng2/async` fails.

cnishina:
Fixed: added back LocationStrategy to the bootstrap.

cnishina:
I believe this error is around trying to install node modules for the testapp

cnishina:
Going to split this into separate commits. One for upgrading package.json and the next one for upgrading to Angular 2 RC3.

PR #3302, opened 2016-06-23
chore(testapp): use testapp's package.json to install node_modules

cnishina:


.travis.yml
```diff
@@ -39,6 +39,7 @@ before_install:
   - g++-4.8 --version
 
 before_script:
+  - npm run testapp
```
juliemr: instead of `testapp`, `install_testapp`?

juliemr: `testapp` makes me think it will serve it.

juliemr:
One nitpick, LG

PR #3304, opened 2016-06-23
chore(testapp): upgrade to angular2 rc4

cnishina:


cnishina:
@juliemr Not sure about this failure. I might need help.

cnishina:
Looked at Travis errors, watched the Sauce Labs screencast:

- It appears it loads the Angular 2 page the first time
- Every time after that it times out. Going to make an update to print out the browser logs in an `afterEach`.

cnishina:
It appears that package.json used angular/router a version of "^3.0.0" for the router.

This downloaded the latest angular router which expected Angular 2 rc 4 but we were tied to Angular 2 rc 3...Locking down on the angular/router version will hopefully fix this problem.

cnishina:
Made a separate app and removed the router. Started increasing timeouts:

```
  getPageTimeout: 120000,
  jasmineNodeOpts: {
    defaultTimeoutInterval: 120000
  }
```

However, per http://www.protractortest.org/#/timeouts, to address the failures I need to add `allScriptsTimeout` to the configuration.

juliemr:
LG! Thanks.

PR #3305, opened 2016-06-24
chore(config): use thrown error stack trace for ConfigError

### LINKED ISSUE #3301: Failing require() in configuration file causes undescriptive error
Bug report
---

- Node Version: `4.4.5`
- Protractor Version: `3.3.0`
- Operating System and Version `Windows 10`

In my configuration file I had a dependency, which I imported with `require('modulename')`, but after updating packages.json this broke the call. However, Protractor responds with the following output:

```
[10:33:08] E/configParser - error code: 105
[10:33:08] E/configParser - description: failed loading configuration file C:\path\protractor.conf.js
```

As hinted per [this StackOverflow remark](http://stackoverflow.com/questions/37000131/protractor-error-105-when-trying-to-run-spec-js#comment63229521_37000131), adding `logger.error(e);` to https://github.com/angular/protractor/blob/master/lib/configParser.ts#L200, revealed the actual error:

```
[10:36:29] E/configParser - { [Error: Cannot find module 'modulename'] code: 'MODULE_NOT_FOUND' }
```

These parsing errors should not be masked by the current error message and should also be reported in the output.

cnishina:
closes #3301

### DIFF
--- a/lib/configParser.ts
+++ b/lib/configParser.ts
@@ -198,7 +198,7 @@ export class ConfigParser {
       fileConfig = require(filePath).config;
     } catch (e) {
       throw new ConfigError(
-          logger, 'failed loading configuration file ' + filename);
+          logger, 'failed loading configuration file ' + filename, e);
     }
     if (!fileConfig) {
       throw new ConfigError(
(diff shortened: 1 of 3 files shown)


sjelin:
Tests?

cnishina:
Tests updated...

PR #3307, opened 2016-06-27
chore(typings): remove ptor from gulp types

cnishina:
- after investigation, variables in the namespace cannot be assigned a value
- types will require import of the classes and then setting the value from global

  ```
  import {Browser} from 'protractor';
  var browser: Browser = global['browser'];
  ```


cnishina:
Here's a sample typescript file for the proposed change.

https://github.com/cnishina/protractor-typings/blob/master/src/spec.ts

PR #3319, opened 2016-06-30
chore(types): fix import of package.json typings

cnishina:
- fixes Exported external package typings file is not a module
 - allows for users to `import {browser,element,by,$,$$,ExpectedConditions} from 'protractor/globals';`


juliemr:
LGTM

PR #3321, opened 2016-07-01
chore(types): split out config interface from configParser

cnishina:
- split out config interface from configParser
- add comments from docs/referenceConf.js to lib/config.ts
- allows for users to `import {Config} from 'protractor';`

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -72,7 +72,7 @@ gulp.task('default',['prepublish']);
 
 gulp.task('types', function(done) {
   var folder = 'built';
-  var files = ['browser', 'element', 'locators', 'expectedConditions'];
+  var files = ['browser', 'element', 'locators', 'expectedConditions', 'config'];
   var outputFile = path.resolve(folder, 'index.d.ts');
   var contents = '';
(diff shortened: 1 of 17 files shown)


gulpfile.js
```diff
@@ -86,6 +86,10 @@ gulp.task('types', function(done) {
 
   // write contents to 'built/index.d.ts'
   fs.writeFileSync(outputFile, contents);
+
+  // copy over config.ts file so users just need to import from 'protractor/config';
```
juliemr: Shouldn't it be called `config.ts` then?

juliemr:
Mostly looks good except for some (probably silly) questions about where files go.

gulpfile.js
cnishina: 1. My comment should say `import {Config} from 'protractor/configuration';` instead...but we could also do `import {Config} from 'protractor/config';`
2. `/lib/globals.d.ts` is only for protractor internals to set up global variables / fix some of the missing interfaces. `/globals.ts` exports variables from variables in the global namespace. So users can do `import {browser} from 'protractor/globals';`

PR #3323, opened 2016-07-01
chore(example): add a protractor typescript example

cnishina:


cnishina:
package.json refers to protractor version 4.0.0 which is not published yet. This should work when 4.0.0 is released.

Xotabu4:
Hi! Can you also add example with pageObjects? Thanks!

cnishina:
@Xotabu4 Added a readme and a pageObjects example.

PR #3329, opened 2016-07-04
Added travis ci support for node 6

amilajack:


Xotabu4:
Currently using 6.2 as default nodeJS version for running tests. No problems detected so far. I think i should be fine.

Issue #2275, opened 2015-06-23
Incorrect flow with ignoreSynchronization set to true?


KrekkieD:
Consider this test:

```
	it ('should fill a form in the proper order', function () {

		browser.ignoreSynchronization = true;

		$('[ng-model="one"]').sendKeys('one');
		$('[ng-model="two"]').sendKeys('two');
		$('[ng-model="three"]').sendKeys('three');
		$('[ng-model="four"]').sendKeys('four');
		$('[ng-model="five"]').sendKeys('five');
		$('[ng-model="six"]').sendKeys('six');
		$('[ng-model="seven"]').sendKeys('seven');

	});
```

I would expect the inputs to be filled in this order, however they are filled in what appears to be a random order. 

To make this extra fun, if I add a line `browser.sleep(1);` after all `sendKeys` commands, the flow is corrected and inputs are filled in the order specified. This seems strange. Should it not always use the sequential flow?

This did not seem to occur with protractor `<2.0.0`.
I use ignoreSynchronization because I need to manually bootstrap my application, but the same behavior occurs for `ng-app` based bootstraps.

talcloudshare:
Same here, having problems with the flow. I've set ignoreSynchronization to true in the beginning of my test spec. In one of my page object's functions, which is only called at the end of the spec, I'm setting it to false. But, when the spec starts running it seems like ignoreSynchronization was set to false all along. Commenting out the "ignoreSynchronization = false" line resolves that issue.
For me it occurs also with protractor 1.8

talcloudshare:
Adding a simplified example:

'''
it('should be able to get page without failing', function () {
    browser.ignoreSynchronization = true;
    var loginButton = element(by.xpath("//input[@value = 'Login']"));
    browser.get("/");
    expect(loginButton.waitReady()).toBeTruthy();
    browser.ignoreSynchronization = false;
}
'''

Beginning to suspect the problem is separated from the issue and is related to waitReady somehow, since when I replaced it with isPresent() and now it seems to work

KrekkieD:
@talcloudshare your example is actually correct behavior, as your ignoreSynchronization change does not return a webdriver promise. The full `it` is read on start of the test, the actual execution in-browser follows after that. That means while reading your `it` it is queueing the webdriver commands (= promises) as well as executing other code (your ignoreSync change). I think in your situation you should change the ignoreSynchronization property in either a following `it`, or create a webdriver promise at the end of your `it` and set the property in the resolve of that promise.

[...]

Issue #1939, opened 2015-03-18
Make it easier to control when `ignoreSynchronization` is off or on


juliemr:
It's kind of difficult to know how it interacts with the control flow at the moment.

hankduan:
Can we just wrap all our command in control flow, and make a function that sets ignoreSynchronization in control flow. 

i.e. `browser.get` (simplified to capture the idea)
now:
```javascript
if (ignoreSynchronization) {driver.get(...)}
else {waitForAngular(); driver.get(...)}
```

after:
```javascript
flow.execute(function() {
  if (ignoreSynchronization) {driver.get(...)}
  else {waitForAngular(); driver.get(...)}
})
```

and make 
```javascript
setIgnoreSynchronization = function(val) {
  flow.execute(function() {
    ignoreSynchronization = val;
  })
}
```

This will make cloning browsers more robust too

juliemr:
Yeah, I think that's reasonable and probably the simplest way to do it. The hardest bit will be making clear docs about usage :)

smhc:
If this were implemented.. I wonder if synchronisation could be turned off automatically by 'get' operations, and on when an action is performed that interacts with the page. i.e 

```javascript
waitForAngular()
click() - turn on sync
waitForAngular()
getText() - turn off sync
getText() - turn off sync
getText() - turn off sync
getText() - turn off sync
getText() - turn off sync 
click() - turn on sync
waitForAngular()
getText()  - turn off sync
```

Does that make sense? I think in some cases it could result in a fairly reasonable performance improvement, without the overhead of having to manage the sync on and off manually.
I don't know how practical it is however.

juliemr:
@smhc I think that's out of scope for this change, but I'm working on an entirely new way of doing waits that will incorporate this sort of smart waiting. Stay tuned :)

martin770:
Any update on this? I was wondering about even just retrieving browser.ignoreSynchronization as a promise or setting/getting within the context of the control flow somehow.

    setIgnoreSynchronization = function(arg) {
        arg = arg === undefined ? true : arg;
        controlFlow().execute(function() { browser.ignoreSynchronization = arg };
    }
    getIgnoreSynchronization = function() {
        return controlFlow().execute(function() { return browser.ignoreSynchronization });
    }

So then in waitForAngular() you can have

    this.getIgnoreSynchronization.then(function(shouldNotSync) {
        if (shouldNotSync) {
          return
        } else {
          ... waitForAngularStuff ...  
[...]

PR #3330, opened 2016-07-05
docs: timeouts: ignoreSynchronization example should be asynchronous

laurence-myers:
ignoreSynchronization is synchronous, whereas other flow commands are asynchronous (promise-based). Therefore, to use ignoreSynchronization correctly, you need to chain promises using the control flow.

As per this issue/comment:
https://github.com/angular/protractor/issues/2275#issuecomment-118135431

### DIFF
--- a/docs/timeouts.md
+++ b/docs/timeouts.md
@@ -45,12 +45,23 @@ If you need to navigate to a page which does not use Angular, you can turn off w
 `browser.ignoreSynchronization = true`. For example:
 
 ```js
+var runNonAngular = function(fn) {
+    var flow = protractor.promise.controlFlow();
+    flow.execute(function () {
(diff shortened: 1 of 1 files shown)


laurence-myers:
I signed it!

juliemr:
This isn't actually how `ignoreSynchronization` works, the original docs are correct. Protractor checks for `ignoreSynchronization` when tasks are _scheduled_, not when they are executed. (See the code [here](https://github.com/angular/protractor/blob/e786ea1ad1a1183a8fee83640f4155932ec666d2/lib/browser.ts#L397))

I agree that this is confusing, and we'd like to fix up the API when we start using BlockingProxy. See https://github.com/angular/protractor/issues/1939

PR #3335, opened 2016-07-06
chore(tests): fix test for typings I broke with templated promises

sjelin:


cnishina:
Maybe fix the commit message... other than that LGTM.

cnishina:
will fix commit with merge squash.

PR #3338, opened 2016-07-07
chore(typings): clang + typings upgrade

cnishina:


cnishina:
Note to self: Fix commit message to typings upgrade.... has nothing to do with ambient typings tests

PR #3339, opened 2016-07-07
chore(logger): remove logger for logger2

cnishina:


cnishina:
part 1: remove logger from repo
part 2: rename logger2 to logger (coming soon)

PR #3368, opened 2016-07-13
chore(gitignore): ignore globals.js

sjelin:


juliemr:
Hm, my setup isn't generating a `globals.js`.

cnishina:
The tsconfig.json should prevent outputting `globals.js`.

sjelin:
After deleting globals.js, it appears to be gone forever.  Sorry!

Issue #3367, opened 2016-07-13
Typings for 'protractor' namespace missing from globals.ts in Protractor 4.0.0


jan-molak:
Although the `protractor/lib/ptor.ts` [defines a `protractor` namespace](https://github.com/angular/protractor/blob/a52207892b773dcd03e87543916ca10a6cf007f6/lib/ptor.ts), some of its contents don't seem to be exported in any of the typings?

The `globals.ts` file [defines](https://github.com/angular/protractor/blob/bed5b070c5b0dc8d39a5d5e9d20c98925ccf41bc/globals.ts):

```typescript
  Browser
  ElementArrayFinder
  ElementFinder
  ElementHelper
  ProtractorBy
  ProtractorExpectedConditions
```

but misses:

```typescript
 wrapDriver
 promise
 ActionSequence
 Key
 Command
 CommandName
```

It would be also great to have other public interfaces available, such as:

```
Locator
```

Thanks!
Jan

Droogans:
I was taking a look at [the `protractor.promise` namespace](https://github.com/angular/protractor/blob/575e41cf22a0ea16826bca86a02c490f929e8b38/lib/ptor.ts#L18-L31) and didn't see `protractor.promise.all`.

nizhu:
Attempting an upgrade to 4.0.0 and failing on `Failed: protractor.promise.all is not a function`. Is there an alternative?

alecxe:
@nizhu you can/should use the webdriver promise directly now, see http://stackoverflow.com/questions/38381292/failed-cannot-read-property-all-of-undefined.

Droogans:
If that is the case then why does `protractor.promise` namespace still get set up and exported? Shouldn't *all* promise functions be cut? Seems strange to keep `protractor.promise.when` but not `protractor.promise.all`.

This seems like a good workaround though! That is, until they either axe the entire promise namespace, or put `all` back.

alecxe:
@Droogans definitely is a workaround, I would prefer `all()` back to the `protractor.promise` namespace as well. Thanks.

cnishina:
related issue: #3391

cnishina:
Alright, changes have been made.. In reference to the missing items from global.ts: The `protractor` variable tied to the global namespace is now part of globals.ts. This should address the following list
```
 wrapDriver
 promise
 ActionSequence
 Key
 Command
 CommandName
```

with importing {protractor}:

```
import {protractor} from 'protractor/globals';
```

[...]

PR #3388, opened 2016-07-18
fix(types): use protractor from global namespace

cnishina:


cnishina:
Relates to https://github.com/angular/protractor/issues/3367

PR #3389, opened 2016-07-18
fix(types): output plugin typings

### LINKED ISSUE #3365: Typings for ProtractorPlugin missing from index.d.ts in Protractor 4.0.0
Hi there,

It seems like the definition of the `ProtractorPlugin` is no longer accessible to plugin developers.

This definition used to be available from the `protractor/built/plugins` package as per the [docs here](https://github.com/angular/protractor/blob/master/docs/plugins.md#writing-plugins-in-typescript):

```typescript
import {ProtractorPlugin} from 'protractor/built/plugins';
```

but as Protractor 4.0.0 does not include this definition in the main `built/index.d.ts` nor does it provide definitions at `protractor/built/plugins/`, the class is beyond reach.

Could you please expose it in the main definitions file please?

Many thanks,
Jan

cnishina:
closes #3365

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -334,17 +334,6 @@ module.exports = myPlugin;
 
 ```
 
-Then, in a separate file, export an instance of that class using an assignment
-export:
-
-```typescript
-import MyPlugin from './MyPlugin';
-
-export = new MyPlugin();
-```
-
-This instance is what you should point to in your config file.
-
 
 First Party Plugins
 -------------------
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -312,17 +312,26 @@ export function onPageLoad(): void {
 };
 ```
 
-However, if you want your code more heavily typed, you can write your plugin as
-an extension of the `ProtractorPlugin` class:
+If you want your code more heavily typed, you can write your plugin with
+the `ProtractorPlugin` interface:
 
 ```typescript
-import {ProtractorPlugin} from 'protractor/built/plugins';
+import {ProtractorPlugin} from 'protractor';
 
-export class MyPlugin extends ProtractorPlugin {
+// creating a "var module: any" will allow use of module.exports
+declare var module: any;
+
+let myPlugin: ProtractorPlugin = {
+  addSuccess(info: {specName: string}) {
+    console.log('on success: ' + info.specName);
+  },
   onPageLoad() {
     this.addSuccess({specName: 'Hello, World!'});
-  };
+  }
 };
+
+module.exports = myPlugin;
+
 ```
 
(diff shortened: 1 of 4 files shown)


cnishina:
change protractor plugins to an interface; however, more direction is needed for:

```
addSuccess: ({specName: 'hello world'})
```

cnishina:
@sjelin will need input for above ^^

docs/plugins.md
```diff
@@ -312,17 +312,26 @@ export function onPageLoad(): void {
...
     this.addSuccess({specName: 'Hello, World!'});
-  };
+  }
 };
+
+module.exports = myPlugin;
+
 ```
 
 Then, in a separate file, export an instance of that class using an assignment
```
sjelin: Delete the part about the separate file now yeah?

lib/plugins.ts
```diff
@@ -18,16 +18,217 @@ export interface PluginConfig {
...
+   * The plugin configuration object. Note that this is not the entire
+   * Protractor config object, just the entry in the plugins array for this
+   * plugin.
+   *
+   * @type {Object}
+   */
+  config?: PluginConfig;
+
+  /**
+   * Adds a failed assertion to the test's results.
```
sjelin: Mention that this is added by the protractor API - not to be implemented by the plugin author

lib/plugins.ts
```diff
@@ -18,16 +18,217 @@ export interface PluginConfig {
...
+   * @throws {Error} Throws an error if called after results have been reported
+   */
+  addFailure?:
       (message?: string,
        info?: {specName?: string, stackTrace?: string}) => void;
-  addSuccess: (info?: {specName?: string}) => void;
-  addWarning: (message?: string, info?: {specName?: string}) => void;
+
+  /**
+   * Adds a passed assertion to the test's results.
```
sjelin: Mention that this is added by the protractor API - not to be implemented by the plugin author

lib/plugins.ts
```diff
@@ -18,16 +18,217 @@ export interface PluginConfig {
...
+   *       - specName The name of the spec which this assertion belongs to.
+   *            Defaults to `PLUGIN_NAME + ' Plugin Tests'`.
+   *     Defaults to `{}`.
+   *
+   * @throws {Error} Throws an error if called after results have been reported
+   */
+  addSuccess?: (info?: {specName?: string}) => void;
+
+  /**
+   * Warns the user that something is problematic.
```
sjelin: Mention that this is added by the protractor API - not to be implemented by the plugin author

sjelin:
Should be good with some minor doc fixes

sjelin:
LGTM

PR #3393, opened 2016-07-19
fix(protractor): export class definitions under the protractor namespace

### LINKED ISSUE #3377: Cannot read property 'prototype' of undefined (Protractor 4 upgrade problem)
After upgrading to Protractor 4.0.0, we are starting to see out tests failing immediately in `onPrepare()` with:

> Error: TypeError: Cannot read property 'prototype' of undefined

failing on the following line:

```js
protractor.ElementArrayFinder.prototype.takewhile = function(whileFn) {
   // ...
};
```
(here we are adding `takewhile` - linking https://github.com/angular/protractor/issues/2521) 

Looks like `ElementArrayFinder` is no longer exposed on the `protractor` global object.

How can we extend `ElementArrayFinder` in Protractor 4? Thanks!

SO reference: http://stackoverflow.com/questions/38380465/cannot-read-property-prototype-of-undefined-after-upgrading-to-protractor-4.

cnishina:
closes #3377

### DIFF
--- a/lib/ptor.ts
+++ b/lib/ptor.ts
@@ -15,6 +15,15 @@ export namespace protractor {
   export let wrapDriver: Function;
   export let ExpectedConditions: ProtractorExpectedConditions;
 
+  // Export the protractor classes
+  export let Browser = require('./browser').Browser;
+  export let ElementFinder = require('./element').ElementFinder;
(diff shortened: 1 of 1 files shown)


cnishina:
Should probably test this out but this should work...

PR #3415, opened 2016-07-26
fix(sauce): sauceAgent passed incorrectly to sauce node module

### LINKED ISSUE #3410: sauceAgent passed incorrectly to SauceLabs.js
It looks like there is an issue with the way the sauceAgent config param is passed to the SauceLabs node module.  It is expecting 'proxy' to be passed in rather than 'agent'.

the sauceAgent config is a string according to this which is fine:
https://github.com/angular/protractor/blob/master/lib/config.ts

..and it looks like the intent is that this is the proxy address to be passed to SauceLabs node module when behind a proxy - docs are a bit unclear.  

when we just set the webDriverProxy, the tests run against saucelabs via the proxy ok, but the final result rest api call fails (outside webdriver) - even though we have the sauceAgent proxy set too.  This means that all tests are left in an error state.

When i locally update this file to match the expected constructor parameter it works fine:

https://github.com/angular/protractor/blob/master/lib/driverProviders/sauce.ts 

from this:
this.sauceServer_ = new SauceLabs({
            username: this.config_.sauceUser,
            password: this.config_.sauceKey,
            **agent**: this.config_.sauceAgent
        });

to this:
this.sauceServer_ = new SauceLabs({
            username: this.config_.sauceUser,
            password: this.config_.sauceKey,
            **proxy**: this.config_.sauceAgent
        });

this means that the parameter matches the expected constructor parameter - 'agent' does not seem to be referenced in SauceLabs.js:

https://github.com/danjenkins/node-saucelabs/blob/master/lib/SauceLabs.js

this options is then used to setup the proxy like this:
options.agent = new HttpsProxyAgent(url.parse(options.proxy));

happy to provide more details if required.

cnishina:
closes #3410

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -92,10 +92,23 @@ export interface Config {
    */
   sauceKey?: string;
   /**
-   * Use sauceAgent if you need customize agent for https connection to
-   * saucelabs.com (i.e. your computer behind corporate proxy)
+   * Use sauceAgent if you need custom HTTP agent to connect to saucelabs.com.
+   * This is needed if your computer is behind a corporate proxy.
+   *
+   * To match sauce agent implementation, use
+   * [HttpProxyAgent](https://github.com/TooTallNate/node-http-proxy-agent)
+   * to generate the agent or use sauceProxy as an alternative. If a
+   * sauceProxy is provided, the sauceAgent will be overridden.
    */
-  sauceAgent?: string;
+  sauceAgent?: HttpProxyAgent;
+  /**
+   * Use sauceProxy if you are behind a corporate proxy to connect to
+   * saucelabs.com.
+   *
+   * The sauceProxy is used to generate an HTTP agent. If a sauceProxy is
+   * provided, the sauceAgent will be overridden.
+   */
+  sauceProxy?: string;
   /**
    * Use sauceBuild if you want to group test capabilites by a build ID
    */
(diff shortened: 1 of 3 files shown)


lib/config.ts
```diff
@@ -93,9 +93,23 @@ export interface Config {
   sauceKey?: string;
   /**
    * Use sauceAgent if you need customize agent for https connection to
```
mgiambalvo: Sorry, this bugs me a little. "Use sauceAgent if you need a customized agent for the https connection..."

lib/config.ts
```diff
@@ -93,9 +93,23 @@ export interface Config {
...
+   *
+   * To match sauce agent implementation, use
+   * [HttpProxyAgent](https://github.com/TooTallNate/node-http-proxy-agent)
+   * to generate the agent or use sauceProxy as an alternative where
+   * sauce generates the agent for you based on the proxy. If a sauceProxy is
+   * provided, the sauceAgent will be overridden.
+   */
+  sauceAgent?: HttpProxyAgent;
+  /**
+   * Use sauceProxy that will get generated into an agent for https connection
```
mgiambalvo: Wording is a little awkward. Maybe

"Use sauceProxy to provide the address of a proxy for the connection to saucelabs.com (i.e. your computer is behind a corporate proxy).

PR #3436, opened 2016-08-01
chore(error): throwing strings

cnishina:
- fix throw string to throw new error
- check typeof error is string to create a new error


lib/util.ts
```diff
@@ -41,18 +41,22 @@ export function runFilenameOrFn_(
...
+      throw new Error('filenameOrFn must be a string or function');
     }
 
     if (typeof filenameOrFn === 'string') {
       filenameOrFn = require(resolve(configDir, filenameOrFn));
     }
     if (typeof filenameOrFn === 'function') {
       let results = when(filenameOrFn.apply(null, args), null, (err) => {
-        if (err.stack) {
+        if (typeof err === 'string') {
```
juliemr: what about

```js
if (typeof err === 'string') {
  err = new Error(err);
}
err.stack = exports.filterStackTrace(err.stack);
throw err;
```

juliemr: Should we check `if (!err.stac)` instead of `typeof`? In case we get some crazy error object without a stack or something?

PR #3451, opened 2016-08-05
chore(util): remove argument type for err

cnishina:
- Fix argument of type 'string | Error' is not assignable to parameter of type 'string'.

### DIFF
--- a/lib/util.ts
+++ b/lib/util.ts
@@ -48,19 +48,18 @@ export function runFilenameOrFn_(
       filenameOrFn = require(resolve(configDir, filenameOrFn));
     }
     if (typeof filenameOrFn === 'function') {
-      let results =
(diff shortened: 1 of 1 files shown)


lib/util.ts
```diff
@@ -48,19 +48,18 @@ export function runFilenameOrFn_(
...
-            } else {
-              err = err as Error;
-              if (err.stack) {
-                err.stack = new Error().stack;
-              }
-            }
-            err.stack = exports.filterStackTrace(err.stack);
-            throw err;
-          });
+      let results = when(filenameOrFn.apply(null, args), null, (err) => {
```
sjelin: Surprised no type is needed for `err` but alright

cnishina: Looked into this and DefinitelyTyped `q.when` has defined a callback `onRejected?: (reason: any)`.

PR #3467, opened 2016-08-11
feat(config): Add 'random' and 'seed' Jasmine config options

szimek:
## Feature Request

This adds handling of `random` and `seed` options in `jasmineNodeOpts` and allows to run specs in random order. The descriptions for both options are taken from Jasmine [docs](http://jasmine.github.io/2.4/node.html#section-14).
  Running specs in random order is usually preferable, as it ensures that the tests don't depend on each other and don't leak unwanted changes.
  Nope.

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -519,6 +519,14 @@ export interface Config {
      * Inverts 'grep' matches
      */
     invertGrep?: boolean;
+    /**
+     * If true, run specs in semi-random order
+     */
+    random?: boolean,
+    /**
+     * Set the randomization seed if randomization is turned on
+     */
+     seed?: string,
   };
 
   /**
--- a/lib/frameworks/jasmine.js
+++ b/lib/frameworks/jasmine.js
@@ -87,6 +87,16 @@ exports.run = function(runner, specs) {
     return true;
   };
 
+  // Run specs in semi-random order
+  if (jasmineNodeOpts.random) {
+    jasmine.getEnv().randomizeTests(true);
+
+    // Sets the randomization seed if randomization is turned on
+    if (jasmineNodeOpts.seed) {
+      jasmine.getEnv().seed(jasmineNodeOpts.seed);
+    }
+  }
+
   return runner.runTestPreparer().then(function() {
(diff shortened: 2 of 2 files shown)


szimek:
I signed it!

szimek:
The one thing that I'm not sure of is how to show a custom text once all specs are finished. In jasmine-npm console reporter they [show what seed was used to randomize specs](https://github.com/jasmine/jasmine-npm/blob/master/lib/reporters/console_reporter.js#L92), which may be useful if your tests fail and you want to restart them in the exact same order, thus using the same seed.

sjelin:
Good idea!  Why not throw it on to `err.message` on 120 of `lib/frameworks/jasmine.js`?

szimek:
Turns out that Protractor doesn't have to handle printing of the seed used, as Jasmine reporters should take care of this. I'm using a custom reporter ([SpecReporter](https://github.com/bcaudan/jasmine-spec-reporter)) instead of the default one, which doesn't seem to print it out, but when I switched back to the default one, I got:
```
Ran 3 of 7 specs
3 specs, 1 failure
Finished in 9.273 seconds
Randomized with seed 91703
[15:34:59] I/launcher - 0 instance(s) of WebDriver still running
[15:34:59] I/launcher - chrome #01 failed 1 test(s)
[15:34:59] I/launcher - overall: 1 failed spec(s)
[15:34:59] E/launcher - Process exited with error code 1
```
so it looks like everything is ok.

szimek:
@sjelin My [PR](https://github.com/bcaudan/jasmine-spec-reporter/pull/55) to jasmine-spec-reporter to show the seed used was merged recently and now I'm getting the following output with the latest jasmine-spec-reporter and protractor that includes this PR:
```
Executed 184 of 184 specs SUCCESS in 8 mins 41 secs.
Randomized with seed 22937.
[20:36:48] I/launcher - 0 instance(s) of WebDriver still running
[20:36:48] I/launcher - chrome #01 passed
```
so I think that this PR can be merged as it is.

cnishina:
Please rebase.

PR #3471, opened 2016-08-11
feat(browser): auto-unwrap ElementFinder into WebElement for selenium…

sjelin:
… funtions

Currently, if you do something like:

```js
browser.executeScript('', element(by.model('modern_major_general')))
```

you will get a `FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory` error because selenium doesn't know what to do with our `ElementFinder` class.  This solves that problem, and reduces the need for our users to know about `getWebElement`

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -51,7 +51,8 @@ export class Webdriver {
 
 /**
  * Mix a function from one object onto another. The function will still be
- * called in the context of the original object.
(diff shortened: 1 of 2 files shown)


sjelin:
Julie I'm assigning this to you because it seems like a pretty consequential change

PR #3472, opened 2016-08-12
chore(docs): document more of selenium's functions on our website

sjelin:


sjelin:
Michael I'm assigning this to you because I haven't assigned much to you in the past and I want to balance that out.

manoj9788:
can't we just link in someway to the original docs ? in that it will reduce the rework if incase something actually changes and we loose track to change it here ? Thoughts ?

sjelin:
We used to do that, but we decided to rewrite them because those docs don't use protractor-style code examples (e.g. `browser.findElement(webdriver.By.id("id"))` vs `element(by.id("id"))`)

mgiambalvo:
Sorry I missed this!

PR #3479, opened 2016-08-16
chore(exampleTypescript): use a tmp folder for compiled files

robinboehm:
keep your src folder clean

screen shot 2016-08-16 at 19 23 30

### DIFF
--- a/exampleTypescript/package.json
+++ b/exampleTypescript/package.json
@@ -9,7 +9,7 @@
     "typings": "typings install",
     "postinstall": "typings install",
     "pretest": "tsc",
-    "test": "protractor conf.js"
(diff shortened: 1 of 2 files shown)


cnishina:
The exampleTypescript folder has a `.gitignore` that ignores javascript. This shouldn't be a problem.

robinboehm:
Hey @cnishina, thanks for your quick reply!
I think it is an problem because people starting real projects based on this example and they don't want to mess up their codebase with compiled output artifacts.
In many editors and ide's it's confusing with both versions side by side.

cnishina:
I'm okay with this. LGTM.

PR #3483, opened 2016-08-18
feat(plugins): support onPrepare in plugins

sjelin:
Allows plugins to access framework-specific variables/funtions

A feature that seemed like it would be useful for some internal google users

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -95,6 +95,22 @@ Plugins are node modules which export an object with the following API:
 exports.setup = function() {};
 
 /**
+ * This is called before the test have been run but after the test framework has
+ * been set up.  Analogous to a config file's `onPreare`.
+ *
+ * Very similar to using `setup`, but allows you to access framework-specific
+ * variables/funtions (e.g. `jasmine.getEnv().addReporter()`)
+ *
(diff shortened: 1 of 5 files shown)


lib/plugins.ts
```diff
@@ -1,6 +1,7 @@
 var webdriver = require('selenium-webdriver');
 import {Logger} from './logger';
 import * as Q from 'q';
+import {Promise as Q_Promise} from 'q';
```
sjelin: This part needed because Q.promise no longer works oddly

docs/plugins.md
```diff
@@ -95,6 +95,22 @@ Plugins are node modules which export an object with the following API:
...
+ * variables/funtions (e.g. `jasmine.getEnv().addReporter()`)
+ *
+ * @throws {*} If this function throws an error, a failed assertion is added to
+ *     the test results.
+ *
+ * @return {Q.Promise=} Can return a promise, in which case protractor will wait
+ *     for the promise to resolve before continuing.  If the promise is
+ *     rejected, a failed assertion is added to the test results.
+ */
+explorts.onPrepare = function() {};
```
cnishina: `explorts`?  Did you mean `exports`?

Issue #3505, opened 2016-08-31
Protractor synchronization fails if multiple versions of selenium-webdriver are in node_modules


christiandreher:
Bug report
---

### Setup

- Node Version: `6.3.1`
- npm Version: `3.10.3`
- Protractor Version: `4.0.4 & 4.0.3`
- Browser(s): `Chrome (52.0.2743.116 (Official Build) (64-bit))`
- Operating System and Version `Ubuntu 16.04`
- Protractor configuration file
```javascript
var SpecReporter = require('jasmine-spec-reporter');

exports.config = {
    allScriptsTimeout: 11000,
    specs: [
        '../e2e/**/*.e2e-spec.ts'
    ],
    capabilities: {
        'browserName': 'chrome'
    },
    directConnect: true,
    baseUrl: 'http://localhost:4200/',
    framework: 'jasmine',
    jasmineNodeOpts: {
        showColors: true,
        defaultTimeoutInterval: 30000,
        print: function() {}
    },
    useAllAngular2AppRoots: true,
    beforeLaunch: function() {
        require('ts-node').register({
            project: 'e2e'
        });
    },
    onPrepare: function() {
        jasmine.getEnv().addReporter(new SpecReporter());
    }
};
```

### Bug description

Whenever I run my e2e tests, my browser opens and immediately closes. Protractor will then claim, that all tests have passed, which just can't be. I implemented a test like this to test this:

```typescript
    it('should fail', () => {
        let ec = protractor.ExpectedConditions;
        let e = element(by.css('some-html-tag-which-definetly-doesnt-exist-in-my-app'));

        browser.wait(ec.presenceOf(e));
    });
```

Although this test should typically timeout, it will pass just fine in a split second.

If I add a spec like `expect(true).toBe(false)`, the test will fail by the way.

I also had a `beforeEach` set up, with `browser.ignoreSynchronization = true`. After removing this line, all tests will pass like before, but protractor will fail anyway with this error:

```
/home/chris/workspace/mem-webclient/node_modules/protractor/built/exitCodes.js:87
if (e.message.indexOf(errMsg) !== -1) {                                                                                                         
              ^                                                                                                                                  
                                                                                                                                                                
TypeError: Cannot read property 'indexOf' of undefined
    at Function.ErrorHandler.isError (/home/chris/workspace/mem-webclient/node_modules/protractor/built/exitCodes.js:87:30)
[...]

PR #3506, opened 2016-08-31
fix(launcher): Handle uncaught exceptions that are strings.

mgiambalvo:


mgiambalvo:
This should stop us from swallowing exceptions (as in #3505, for example)

PR #3507, opened 2016-08-31
chore(types): webdriver typings for locators

cnishina:
- temporarily add typings for selenium-webdriver.d.ts
- include selenium-webdriver dependency to built/index.d.ts
- add webdriver typings to locators
- update example and spec/install to not use typings.json
  - spec test updated to get the tsc test to pass
- includes clang formatting fixes

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -88,7 +88,8 @@ gulp.task('types', function(done) {
   var files = ['browser', 'element', 'locators', 'expectedConditions',
     'config', 'plugins', 'ptor'];
   var outputFile = path.resolve(folder, 'index.d.ts');
-  var contents = '';
+  var contents = '/// <reference path="../typings/index.d.ts" />\n';
+  contents += 'import {By, WebDriver, WebElement, promise} from \'selenium-webdriver\';\n';
   files.forEach(file => {
     contents += parseTypingsFile(folder, file);
   });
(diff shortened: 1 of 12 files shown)


lib/element.ts
```diff
@@ -964,9 +961,9 @@ export class ElementFinder extends WebdriverWebElement {
    * @returns {ElementFinder} which resolves to whether
```
juliemr: Can you update this comment too?

gulpfile.js
```diff
@@ -88,7 +88,8 @@ gulp.task('types', function(done) {
   var files = ['browser', 'element', 'locators', 'expectedConditions',
     'config', 'plugins', 'ptor'];
   var outputFile = path.resolve(folder, 'index.d.ts');
-  var contents = '';
+  var contents = '/// <reference path="../typings/index.d.ts" />\n';
```
juliemr: Can you explain why we need this and didn't need it before?

typings/index.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference path="globals/selenium-webdriver/index.d.ts" />
```
juliemr: Can we add a note that these are forked from definitely typed so we can upkeep them?

lib/element.ts
cnishina: Done. Updated comment for both `isPresent` and `isElementPresent`.

gulpfile.js
cnishina: As discussed, triple slash reference path brings in the ambient typings. We are doing this specifically for selenium-webdriver typings.

typings/index.d.ts
cnishina: Added a readme file. We will remove this folder when we can install these typings with `npm install @types/selenium-webdriver@2.53.1`.

PR #3513, opened 2016-09-02
chore(types): webdriver typings for elements and browser

cnishina:
- include node and jasmine dependency to built/index.d.ts
- update example and spec/install to not need @types/jasmine and @types/node to install
- added an interface in globals for Error to include a code and stack
- improve webdriver typings to elements


cnishina:
@googlebot cla yes?

cnishina:
Thanks @googlebot

PR #3518, opened 2016-09-06
Update tutorial.md

hadrien-toma:
Fixing double quotes issue inside an <input> tag.


juliemr:
Fixed with https://github.com/angular/protractor/pull/3560

PR #3520, opened 2016-09-07
chore(types): clean up TypeScript support

### LINKED ISSUE #3430: Better typings Support for the webdriver api
Feature Request
---

Better typings Support for the webdriver api

Currently many of the webdriver functions specified in the selenium api are not possible to use with the new protractor typings. I see why you might want to hide some of the api, however a lot is missing right now.

We upgraded to 4.0.2 and want to use the new typings however we are unable to as long as parts of the api is missing. (at least in a easy and clean way)

We specifically use 
```typescript
  executeScript: (script: string) => webdriver.promise.Promise<any[]>;
  manage: () => Options;
```

But more can be found her: http://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_WebDriver.html

As a small note shouldn't 

```typescript
export class Webdriver {

}
```

be an interface instead? I believe an interface would be possible to extend without changes code in the protractor module.

Let me know if you are interested in a pull request

### LINKED ISSUE #3477: Property 'manage' doesn't exist on type 'ProtractorBrowser'
I am trying to use `browser.manage().window().maximize()` in my config and specs, The typescript compiler throws the above error. I think the method is not yet exported! Similar case is for `browser.switchTo().alert()` few of the properties are still not exported!

- Protractor Version: `4.0.3`

### LINKED ISSUE #3500: optional class members in protractor using typescript
Optional class members were introduced in `TypeScript 2.0` as a result older typescript `versions(1.8)` using new `Protractor 4.0.4` version would throw an compilation error : 
```
node_modules/protractor/built/index.d.ts(516,13): error TS1112: A class member cannot be declared optional.
```
due to this declaration in `index.d.ts` : 

screen shot 2016-08-27 at 9 32 33 pm

Suggestions:

* Either officially support protractor with typescript 2.0  'or'
* Switch to Interfaces having optional members rather than classes which would promote backward compatibility.

Any thoughts?

cnishina:
- fix exampleTypescript to have noGlobals
- change the exampleTypescript tsconfig.json to include @types
- alias promise as wdpromise everywhere
- refactor gulpfile task 'types' to use multiple files from tsc
- add browser method types applied from webdriver

this closes #3430, closes #3477, and closes #3500

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -49,9 +49,10 @@ gulp.task('checkVersion', function(done) {
   }
 });
 
-gulp.task('built:copy', function() {
-  return gulp.src(['lib/**/*.js','lib/globals.d.ts'])
+gulp.task('built:copy', function(done) {
+  return gulp.src(['lib/**/*.js','lib/globals.d.ts','lib/index.d.ts'])
       .pipe(gulp.dest('built/'));
+  done();
 });
 
 gulp.task('webdriver:update', function(done) {
@@ -88,29 +89,14 @@ gulp.task('tsc:globals', function(done) {
 });
 
 gulp.task('prepublish', function(done) {
-  runSequence('checkVersion', ['jshint', 'format'], 'tsc', 'tsc:globals', 'types',
+  runSequence('checkVersion', ['jshint', 'format'], 'tsc', 'tsc:globals',
     'built:copy', done);
 });
 
 gulp.task('pretest', function(done) {
   runSequence('checkVersion',
     ['webdriver:update', 'jshint', 'format'], 'tsc', 'tsc:globals',
-    'types', 'built:copy', done);
+    'built:copy', done);
 });
 
 gulp.task('default',['prepublish']);
-
-gulp.task('types', function(done) {
-  var outputFile = path.resolve('built', 'index.d.ts');
-  var contents = '';
-  contents += '/// <reference path="../typings/index.d.ts" />\n';
-  contents += '/// <reference path="./globals.d.ts" />\n';
-  contents += 'export { ElementHelper, ProtractorBrowser } from \'./browser\';\n';
-  contents += 'export { ElementArrayFinder, ElementFinder } from \'./element\';\n';
-  contents += 'export { ProtractorExpectedConditions } from \'./expectedConditions\';\n';
-  contents += 'export { ProtractorBy } from \'./locators\';\n';
-  contents += 'export { Config } from \'./config\';\n';
-  contents += 'export { Ptor } from \'./ptor\';\n';
-  fs.writeFileSync(outputFile, contents);
-  done();
-});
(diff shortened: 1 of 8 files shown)


ZurabWeb:
+1

lib/browser.ts
```diff
@@ -32,20 +33,36 @@ for (let foo in webdriver) {
   exports[foo] = webdriver[foo];
 }
 
-// Explicitly define webdriver.WebDriver.
+// Explicitly define webdriver.WebDriver
```
juliemr: Please add a TODO to extend Webdriver from our typings for it.

lib/element.ts
```diff
@@ -20,25 +20,25 @@ let WEB_ELEMENT_FUNCTIONS = [
 // Explicitly define webdriver.WebElement.
```
juliemr: Please add a TODO here as well

gulpfile.js
```diff
@@ -97,90 +97,28 @@ gulp.task('tsc:globals', function(done) {
...
 });
 
 gulp.task('default',['prepublish']);
 
 gulp.task('types', function(done) {
-  var folder = 'built';
-  var files = ['browser', 'element', 'locators', 'expectedConditions',
-    'config', 'plugins', 'ptor'];
-  var outputFile = path.resolve(folder, 'index.d.ts');
+  var outputFile = path.resolve('built', 'index.d.ts');
```
juliemr: Instead of having a gulp task for this, can this now just be a `lib/index.ts` file?

juliemr: Or do we still need this to handle the reference path business?

lib/globals.d.ts
```diff
@@ -25,7 +25,7 @@ declare namespace NodeJS {
     Command: any;
```
juliemr: Can you add a file comment to the top of this file explaining what it's for? Especially with `globals.ts` floating around, it's confusing.

juliemr:
Sorry this took me so long to review. Shame!

I think we could use some more documentation to make navigating our TS infrastructure easier, so I requested a bit here. Feel free to leave everything else for follow-up PRs.

lib/globals.d.ts
cnishina: Done. Also cleaning up the rest of the webdriver types

gulpfile.js
cnishina: Created a `lib/index.d.ts` instead of generating it.

lib/element.ts
cnishina: done.

lib/browser.ts
cnishina: done.

juliemr:
Looks good.

lib/globals.d.ts
juliemr: I was thinking something at the top of the file to describe why the entire file exists (e.g. ambient typings that aren't captured in the rest of our type dependencies, but we can revisit this.

PR #3523, opened 2016-09-07
fix(logger): Set the log level based on the config at startup

### LINKED ISSUE #3522: Is troubleshoot flag still working?
Apologize if there is an issue about that, but, according to my experiments `--troubleshoot` command-line argument or `troubleshoot` configuration option does not change the log level since protractor@3.2.1? (works in 3.2.1 but does not in 3.3.0 and 4) 

Could you please confirm if this is a bug, or am I missing something? Thanks so much!

mgiambalvo:
Fixes #3522. Also fix the mocha spec to stop yelling at us about ES6
arrow functions.

### DIFF
--- a/lib/launcher.ts
+++ b/lib/launcher.ts
@@ -106,7 +106,7 @@ let initFn = function(configFile: string, additionalConfig: Config) {
     configParser.addConfig(additionalConfig);
   }
   let config = configParser.getConfig();
-
+  Logger.set(config);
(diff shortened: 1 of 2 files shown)


cnishina:
`mocha/lib_spec.js` fixes the jshint issue?

mgiambalvo:
Yeah, that was my bad. Though, we should probably tell jshint ES6 stuff is ok once Node 6 is the LTS.

PR #3543, opened 2016-09-13
chore(build): fix version check in gulpfile

juliemr:
Previously, the version check would say that 5.2 was less than 4.3.
Fix by using semver.


juliemr:
Also, this task used to fail silently, and just not continue the gulp run. This change makes it actually error out.

PR #3546, opened 2016-09-13
fix(types): do not publish built/globals.d.ts

cnishina:


cnishina:
Breaks things in bad ways... needs investigation.

lib/browser.ts
```diff
@@ -679,8 +679,12 @@ export class ProtractorBrowser extends Webdriver {
                     let $timeout = $delegate;
 
                     let taskId = 0;
-                    if (!window['NG_PENDING_TIMEOUTS']) {
-                      window['NG_PENDING_TIMEOUTS'] = {};
+
```
juliemr: Thanks for fixing here - we should actually just move this whole function into `clientSideScripts` so we don't deal with TS here at all. 
I filed an issue at https://github.com/angular/protractor/issues/3550

lib/config.ts
```diff
@@ -1,3 +1,5 @@
+export interface HttpProxyAgent { constructor(opts: Object): HttpProxyAgent; }
```
juliemr: Why not just type that as `any`? I'm not sure why this particular one requires a special type.

lib/browser.ts
cnishina: This was really messy and I agree we should pull this out of `browser.ts`

lib/config.ts
cnishina: I agree, we should remove the type in the config. `HttpProxyAgent` does not have an ambient typings; however, we should not have to define the interface for the constructor.

juliemr:
Passes all my local tests, the TS looks good. Let's merge!

PR #3548, opened 2016-09-14
fix(typescript): do not export our global helper

juliemr:
lib/global.d.ts contains fixes necessary for Protractor to compile.
But it should not be shared with users - this change removes that.
It was problematic previously because it could conflict with
users' type definitions for global and others.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -50,7 +50,7 @@ gulp.task('checkVersion', function(done) {
 });
 
 gulp.task('built:copy', function(done) {
-  return gulp.src(['lib/**/*.js','lib/globals.d.ts','lib/index.d.ts'])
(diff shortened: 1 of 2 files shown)


juliemr:
See instead https://github.com/angular/protractor/pull/3546

jan-molak:
Hey @juliemr, qq: if globals.d.ts is no longer published, how would one get hold of global objects like protractor, browser, by, etc?

juliemr:
@jan-molak So this is extra confusing because we have two files named `globals`. The one where you do `import {browser} from 'protractor/global'` *still works*. This CL is for some internal cleanup of a file in `protractor/lib/global`. Hope that clarifies!

jan-molak:
Ah, gotcha! thanks @juliemr, I was a bit worried for a moment :)

PR #3555, opened 2016-09-14
chore(testapp): get the hybrid ngUpgrade test app working with rc7

juliemr:


juliemr:
I've now added a second commit which fixes the tests and turns them back on. (Though note the TODO in there, I had to do a hack to get it working).

PR #3558, opened 2016-09-14
chore(ts): remove most of globals.d.ts

juliemr:


cnishina:
I am not a fan of `<any>global` but it definitely is a start to getting rid of `lib/globals.d.ts`. We will need to have to move out `IError`.

Since we have the `lib/exitCodes.ts`,  it might be nice to stick both the `IError` interface and the `exitCodes` into an `error(s)/` folder.

PR #3569, opened 2016-09-16
chore(element): fix optional param types for element array finder

cnishina:


lib/element.ts
```diff
@@ -491,7 +491,7 @@ export class ElementArrayFinder extends WebdriverWebElement {
    * @returns {!webdriver.promise.Promise} A promise which will resolve to
    *     an array of ElementFinders represented by the ElementArrayFinder.
    */
-  then(fn: Function, errorFn: Function): wdpromise.Promise<any> {
+  then(fn: Function, errorFn?: Function): wdpromise.Promise<WebElement[]> {
```
juliemr: Change to `wdpromise.Promise<any[]>`

cnishina: done.

PR #3570, opened 2016-09-19
feat(elementexplorer): highlight elements

### LINKED ISSUE #3465: Could not find testability for element when use elementor or element explorer
Hello, 

I am trying to run elementor, in order to locate elements on my page. But i always got the same message :

 ERROR: TypeError: Cannot create property 'stack' on string 'Error while waiting for Protractor to sync with the page: "Could not find testability for element."'

image

image

Same error when trying to using Element explorer. 

image
## Bug report
- Node Version: `6.2.0`
- Protractor Version: `4.0.3`
- Browser(s): `Chrome`
- Operating System and Version `Windows 7`
- Your protractor configuration file
- A relevant example test
- Output from running the test
- Steps to reproduce the bug
- The URL you are running your tests against (if relevant)
## Feature Request

cnishina:
- Highlight elements by changing css
- Use the rootEl default 'body'. On second attempt, use a default rootEl ''.
  After the attempt is made, reset the rootEl.

closes #3465

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -1,10 +1,10 @@
 // Util from NodeJs
 import * as net from 'net';
 import {ActionSequence, Capabilities, Command as WdCommand, FileDetector, Options, promise as wdpromise, Session, TargetLocator, TouchSequence, until, WebDriver, WebElement} from 'selenium-webdriver';
-
 import * as url from 'url';
 import * as util from 'util';
 
+import {ExplorerScripts} from './debugger/clients/explorerScripts';
 import {build$, build$$, ElementArrayFinder, ElementFinder} from './element';
 import {ProtractorExpectedConditions} from './expectedConditions';
 import {Locator, ProtractorBy} from './locators';
@@ -376,11 +376,16 @@ export class ProtractorBrowser extends Webdriver {
    * available on the page when finding elements or waiting for stability.
    * Only compatible with Angular2.
    */
-  useAllAngular2AppRoots() {
+  useAllAngular2AppRoots(): void {
     // The empty string is an invalid css selector, so we use it to easily
     // signal to scripts to not find a root element.
     this.rootEl = '';
   }
+
+  /**
+   * Set the rootEl
+   */
+  setRootEl(rootEl: string): void { this.rootEl = rootEl; }
 
   /**
    * The same as {@code webdriver.WebDriver.prototype.executeScript},
@@ -994,17 +999,9 @@ export class ProtractorBrowser extends Webdriver {
       [key: string]: any;
     }
     let context: Context = {require: require};
-    global.list = (locator: Locator) => {
-      return (<Ptor>global.protractor)
-          .browser.findElements(locator)
-          .then((arr: webdriver.WebElement[]) => {
-            let found: string[] = [];
-            for (let i = 0; i < arr.length; ++i) {
-              arr[i].getText().then((text: string) => { found.push(text); });
-            }
-            return found;
-          });
-    };
+    (<any>global).list = ExplorerScripts.list;
+    (<any>global).highlight = ExplorerScripts.highlight;
+
     for (let key in global) {
       context[key] = global[key];
     }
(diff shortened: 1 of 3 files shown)


cnishina:
Per conversation... should pull out the string function and add a js file.

manoj9788:
Good to see this feature, however this was a simple hack that I am using currently: http://www.assertselenium.com/protractor/highlight-elements-during-your-protractor-test-run/

juliemr:
This feels like two separate features - can I be a pain and ask you to split it into two PRs?

lib/browser.ts
```diff
@@ -994,17 +999,9 @@ export class ProtractorBrowser extends Webdriver {
...
-          .browser.findElements(locator)
-          .then((arr: webdriver.WebElement[]) => {
-            let found: string[] = [];
-            for (let i = 0; i < arr.length; ++i) {
-              arr[i].getText().then((text: string) => { found.push(text); });
-            }
-            return found;
-          });
-    };
+    (<any>global).list = ExplorerScripts.list;
```
juliemr: with `declare var global` at the top you shouldn't need the casting here.

juliemr: (and throughout)

lib/browser.ts
```diff
@@ -1168,16 +1176,16 @@ export class ProtractorBrowser extends Webdriver {
...
+      logger.info('------- Element Explorer ---------------------------------');
+      logger.info('Starting WebDriver debugger in a child process. Element');
+      logger.info('Explorer is still beta, please report issues at ');
+      logger.info('github.com/angular/protractor');
       logger.info();
       logger.info('Type <tab> to see a list of locator strategies.');
-      logger.info(
-          'Use the `list` helper function to find elements by strategy:');
-      logger.info('  e.g., list(by.binding(\'\')) gets all bindings.');
+      logger.info('Use the `list` and `highlight` helper function to find');
```
juliemr: typo: helper functionS

lib/debugger/clients/explorerScripts.ts
```diff
@@ -0,0 +1,88 @@
+import {promise, WebElement} from 'selenium-webdriver';
+
+import {ProtractorBrowser} from '../../browser';
+import {Locator} from '../../locators';
+let absoluteXPath = require('./explorerXpath').absoluteXPath;
+
```
juliemr: Use the `declare var global: any` trick here too.

lib/debugger/clients/explorerScripts.ts
```diff
@@ -0,0 +1,88 @@
...
+  static highlight(locator: Locator) {
+    let xPaths: string[] = [];
+    let cssBorders: string[] = [];
+
+    return (<any>global)
+        .browser.findElements(locator)
+        .then((arr: WebElement[]) => {
+
+          for (let i = 0; i < arr.length; ++i) {
+            ExplorerScripts.getAbsoluteXPath(arr[i]).then(
```
juliemr: I think we can get around the whole xpath business by passing in the web element. Webdriver's `ExecuteScript` command lets you pass in a WebElement as an argument, which is converted to the proper HTMLElement on the browser side.

lib/browser.ts
```diff
@@ -1074,9 +1071,20 @@ export class ProtractorBrowser extends Webdriver {
       execute_: function(execFn_: Function) {
         this.execPromiseResult_ = this.execPromiseError_ = undefined;
 
-        this.execPromise_ = this.execPromise_.then(execFn_).then(
-            (result: Object) => { this.execPromiseResult_ = result; },
-            (err: Error) => { this.execPromiseError_ = err; });
+        (<any>global).tempEl = (<any>global).browser.rootEl;
```
juliemr: It makes me nervous to add this here. I think this should be a change in the general logic for how we look for the root Element, not something specific to the debugger (which is not just used for elementExplorer, but also other REPLs if I'm remembering correctly.

cnishina:
Closing this again. We'll get it in one day.

PR #3572, opened 2016-09-19
chore(globals): move error interface to exitCodes and remove globals.d.ts

cnishina:


juliemr:
Looks good! I'm reviewing the short easy one first :)

PR #3577, opened 2016-09-20
[testapp] ng1 Added iframe and windows to the testapp

manoj9788:
Works as it is, simple and no fancy stuff,  Intent is that we can write tests using Protractor API for scenarios like switching into iFrames and switching into new window. Tests will be in my next PR.


juliemr:
Thanks for the contribution. I have some questions about the goal. Why do we need to write tests for iframes and windows? There is nothing special about the way that Protractor handles these cases compared to anything else build on top of webdriver, or how Angular handles them as a framework. I don't want to pollute our tests with complexity that's not related to us.

If we're adding tests here because users have to deal with iframes and windows, and need to see examples, that would be better to put in a place like https://github.com/angular/protractor-cookbook

manoj9788:
@juliemr I still look at the tests for any open source tools/libraries/modules as a first point of place to see how its been implemented, had a look at our tests and the frames and windows part was missing so thought I can try some Angular dev work. - Hence the PR.

I agree that now we have an official cookbook it makes sense to write tests as an showcase there.

Our testapp is a very good place (I recommend in my training and conferences) as a source to start learning Protractor, so I think it would be good at least to have them merged in and users can practice (switching frames and windows related scenarios) working on the test app.

juliemr:
@cnishina had added a test app for the cookbook - I think that will be the best place for this to go. See https://github.com/angular/protractor-cookbook/tree/master/testapp

juliemr:
Closing this in here for now, please port it over to the cookbook. Thanks!!

manoj9788:
@juliemr that testapp doesn't have all controls like our testapp in protractor repo does, I still recommend this to be in here.

juliemr:
We should put effort towards growing the cookbook testapp to the point that it can be a good teaching tool, if it's lacking, rather than using the tests here. The testapp in this repository should be for our own validations primarily.

PR #3608, opened 2016-09-29
fix(launcher): Ignore uncaught exceptions from webdriver.

### LINKED ISSUE #3553: Option to disable uncaughtException listener
Please add an option to disable the `uncaughtException` listener from lib/launcher.ts:182 since it interferes with Cucumber as described in #3491.

Quick recap:
Protractor's `uncaughtException` listener registers first so it's fired before Cucumber's listener for the same event, which would normally handle whatever error occurred inside a step definition and gracefully fail the step. Which in turn leads to Protractor exiting the suite at the first exception (which could be, for example, a failed assertion).

mgiambalvo:
Fixes #3553.

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -457,6 +457,12 @@ export interface Config {
      */
     untrackOutstandingTimeouts?: boolean;
 
+    /**
+     * If set, Protractor will ignore uncaught exceptions instead of exiting
+     * without an error code. The exceptions will still be logged as warnings.
+     */
+    ignoreUncaughtExceptions?: boolean;
+
     // ---------------------------------------------------------------------------
     // ----- The test framework
     // --------------------------------------------------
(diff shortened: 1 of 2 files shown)


mgiambalvo:
I'd like to have a test for this. I tried making a minimal custom test framework that would duplicate Cucumber's behavior. However, I wasn't able to come up with something that would both wait for all the WebDriver control flow promises to resolve, and also somehow throw webdriver errors as uncaught exceptions.

lib/config.ts
```diff
@@ -457,6 +457,13 @@ export interface Config {
      */
     untrackOutstandingTimeouts?: boolean;
 
+    /**
+     * If set, Protractor will ignore uncaught exceptions instead of exiting
+     * with
```
juliemr: Comment indentation is weird.

cdaguerre:
Could you please release?

PR #3618, opened 2016-10-05
fix(conditions) Explicitly remove newlines from getText()

yaycmyk:
MSEdge does not properly remove newlines, which causes false negatives when using `textToBePresentInElement()`


mgiambalvo:
Thanks for the PR!

PR #3620, opened 2016-10-05
docs(webdriver): Add missing quotes in getAttribute doc.

### LINKED ISSUE #3619: webdriver.getAttribute documentation: Missing quotes around class parameter can be confusing
Hello everyone !

I am not much experienced with protractor and needed to use `getAttribute` function to check for a `class` attribute.

Reading the documentation helped me implement it but I just couldn't make it work because I got confused by some missing quotes around `class` parameter which is a `string`.

Here is the documentation extract at `lib/selenium-webdriver/webdriver.js:556` : `expect(foo.getAttribute(class))...`

I am going to reference a PR where I added quotes around the class word showing the paramater is a string and not a variable.

Thank you for your work ! :-)

thecuriousnoob:
In getAttibute function documentation :

`expect(foo.getAttribute(class))...` was written.

This got me confused when getting a class attribute.

I just added quotes around the class word showing class as a string
and not a variable.

Close #3619

### DIFF
--- a/lib/selenium-webdriver/webdriver.js
+++ b/lib/selenium-webdriver/webdriver.js
@@ -305,7 +305,7 @@ webdriver.WebDriver.prototype.takeScreenshot = function() {};
 
 /**
  * Used to switch WebDriver's focus to a frame or window (e.g. an alert, an
- * iframe, another window).  
+ * iframe, another window).
  *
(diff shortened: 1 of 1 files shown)


thecuriousnoob:
I signed it!

mgiambalvo:
CI failure is a flake, and this is only documentation changes, anyway. Merging.

PR #3621, opened 2016-10-06
Refactor the debugger out of browser.ts

mgiambalvo:
This pulls the debugger out of browser.ts and into a separate module. This is the first step in porting the debugger to typescript and adding unit testing around it.

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -995,7 +995,7 @@ export class ProtractorBrowser extends Webdriver {
    * process
    */
   pause(opt_debugPort?: number): webdriver.promise.Promise<any> {
-    if (!this.debugHelper.isAttached()) {
+    if (this.debugHelper.isAttached()) {
       logger.info(
           'Encountered browser.pause(), but debugger already attached.');
       return webdriver.promise.fulfilled(true);
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -264,7 +264,7 @@ export class ProtractorBrowser extends Webdriver {
   ng12Hybrid: boolean;
 
   /**
-   * A helper manages debugging.
+   * A helper that manages debugging tests.
    */
   debugHelper: DebugHelper;
 
@@ -387,8 +387,7 @@ export class ProtractorBrowser extends Webdriver {
    * @param {string} description A description of the command for debugging.
    * @param {...*} var_args The arguments to pass to the script.
    * @returns {!webdriver.promise.Promise.<T>} A promise that will resolve to
-   * the
-   *    scripts return value.
+   * the scripts return value.
    * @template T
    */
   public executeScriptWithDescription(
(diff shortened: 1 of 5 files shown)


lib/browser.ts
```diff
@@ -261,14 +253,7 @@ export class ProtractorBrowser extends Webdriver {
    * when running element explorer.
    * @private {number}
```
juliemr: Can you update the silly old closure typings to public as well?

lib/browser.ts
```diff
@@ -1000,191 +938,7 @@ export class ProtractorBrowser extends Webdriver {
...
-        }
-        if (this.execPromiseError_) {
-          throw this.execPromiseError_;
-        }
-        return this.execPromiseResult_;
-      }
-    };
-
-    return pausePromise;
+    return this.debugHelper.init(debuggerClientPath, onStartFn, opt_debugPort);
```
juliemr: Can we just call this directly in the two places where `initDebugger_` is used?

lib/browser.ts
```diff
@@ -1251,7 +1005,7 @@ export class ProtractorBrowser extends Webdriver {
       logger.info('Encountered browser.pause(). Attaching debugger...');
       if (firstTime) {
         logger.info();
-        logger.info('------- WebDriver Debugger -------');
+        logger.info('------- WebDriver DebugHelper -------');
```
juliemr: I'm not sure this change is meaningful to the user.

lib/debugger.ts
```diff
@@ -0,0 +1,273 @@
...
+  /**
+   * Set to true when we validate that the debug port is open. Since the debug
+   * port is held open forever once the debugger is attached, it's important
+   * we only do validation once.
+   */
+  debuggerValidated_: boolean;
+
+  dbgCodeExecutor: any;
+
+  constructor(browser: ProtractorBrowser) {
```
juliemr: Can shortcut to:

```ts
constructor(private browserUnderDebug_: ProtractorBrowser) {}
```

(and I think that the browser should be private)

lib/debugger.ts
```diff
@@ -0,0 +1,273 @@
...
+    let vm_ = require('vm');
+    let flow = webdriver.promise.controlFlow();
+
+    interface Context {
+      require: any;
+      [key: string]: any;
+    }
+    let context: Context = {require: require};
+    global.list = (locator: Locator) => {
+      return (<Ptor>global.protractor)
```
juliemr: I think this can use `this.browserUnderDebug_`

lib/debugger.ts
mgiambalvo: Good point, done.

lib/debugger.ts
mgiambalvo: Fixed where possible. We have to leave things the way they are for dbgCodeExecutor, but I can fix that when I untangle it later.

lib/browser.ts
mgiambalvo: Haha, this is Intellij's rename refactor being extremely thorough.

juliemr:
LGTM

PR #3632, opened 2016-10-11
Update spec.ts

holdfenytolvaj:
fix import and text check


holdfenytolvaj:
I signed it!

holdfenytolvaj:
dfdfd

holdfenytolvaj:
It seems the second await is only needed because of the last promise skip bug

holdfenytolvaj:
Double checked, so with the current typescript 2.0.3, it is needed to have the second async
otherwise, the expect - check wont be done. (e.g. if in the next test you write
`await browser.get('google.com')`
it will throw an exception because the previous test hasn't been finished.) It seems this is only needed if there is at least one await call before the expect.

mgiambalvo:
Thanks for the fix!

PR #3650, opened 2016-10-20
chore(types): fix transpile errors related to IThenable

cnishina:


lib/browser.ts
```diff
@@ -775,8 +775,8 @@ export class ProtractorBrowser extends Webdriver {
         .then(loadMocks, deferred.reject);
 
     let self = this;
-    function loadMocks(angularVersion: number) {
-      if (angularVersion === 1) {
+    function loadMocks(angularVersion: string) {
+      if (angularVersion == '1') {
```
thorn0: `testForAngular` returns `ver` as `number`. [This line](https://github.com/angular/protractor/blob/cc459402970afe5270e54c087a5347592cff9270/lib/browser.ts#L761) is where `string` needs to be replaced with `number`.

lib/element.ts
```diff
@@ -178,32 +178,28 @@ export class ElementArrayFinder extends WebdriverWebElement {
...
-                        (childrenList: webdriver.WebElement[],
-                         resolvedE: webdriver.WebElement) => {
-                          return childrenList.concat(resolvedE);
-                        },
-                        []);
-                  });
-            });
+          // Resolve the list of Promise<List<child_web_elements>> and merge
+          // into a single list
+          return wdpromise.all(childrenPromiseList).then((resolved: any) => {
```
thorn0: Why `any`? `resolved` seems to be `WebElement[][]`. But again I think it's better just to delete the type annotations here. Excessive type annotations are a kind of type assertions, they prevent the compiler from doing its job of finding type errors.

lib/element.ts
```diff
@@ -178,32 +178,28 @@ export class ElementArrayFinder extends WebdriverWebElement {
...
+            return resolved.reduce(
+                (childrenList: WebElement[],
+                 resolvedE: webdriver.WebElement) => {
```
thorn0: These annotations seems to be wrong and excessive too.

lib/browser.ts
cnishina: This change is to satisfy TypeScript, loadMocks must have a string value. Also, since angularVersion is a string, it must then be compared to a string value. After the files are transpiled, the value '1' and 1 should not matter.

lib/driverProviders/driverProvider.ts
```diff
@@ -61,7 +64,7 @@ export class DriverProvider {
     if (driver.getSession() === undefined) {
       deferred.resolve();
     } else {
-      driver.getSession().then((session_: string) => {
+      driver.getSession().then((session_: Session) => {
```
cnishina: Done.

lib/browser.ts
thorn0: But why must `loadMocks` have a string value? Because it's inferred from the type of the `angularTestResult` variable whose type is wrong.

lib/element.ts
```diff
@@ -751,14 +749,16 @@ export class ElementFinder extends WebdriverWebElement {
...
        */
-      this.then = (fn: Function, errorFn: Function) => {
-        return this.elementArrayFinder_.then((actionResults: any) => {
-          if (!fn) {
-            return actionResults[0];
-          }
-          return fn(actionResults[0]);
-        }, errorFn);
-      };
+      this.then =
```
thorn0: This function should be made generic I think.

lib/browser.ts
cnishina: Good catch. Thanks! Fixed.

lib/element.ts
cnishina: I'll defer to @juliemr

cnishina:
Thanks @thorn0 for the review! 🎉

lib/element.ts
juliemr: I don't think this is a good case for generics, because we don't actually know what value it will return based on the arguments to the function - it's based on what has been called in the chain.

PR #3656, opened 2016-10-24
fix(types): fix types for element.all(...)

thorn0:
Currently. `all` is marked as optional. It means `all` can be `undefined`, which isn't true. For this reason, it's impossible now to use `element.all` with TS 2.x if `strictNullChecks` is `true`. Simple code like
```
element.all(by.css('.info'))
```
leads to an error: `Object is possibly 'undefined'`.


thorn0:
Pls rebase this on top of #3650 to make Travis green.

thorn0:
Rebased, but something still fails on Travis. Seems unrelated to my commit.

Issue #3671, opened 2016-10-27
Documentation - Broken links in Choosing a Framework section 


rajesh825:
Hello, 
Few broken links present in "Choosing a Framework " Documentation 
 (http://www.protractortest.org/#/frameworks. ).  

Please refer to the screenshot for further reference.

![protractor_documentation_brokenlinks](https://cloud.githubusercontent.com/assets/5097539/19762888/af5cafc8-9c33-11e6-9b11-1cf0e2542fc9.png)

Thanks in advance.

sjelin:
Once @cnishina reviews #3674 and we get this merged in this might be fixed

PR #3674, opened 2016-10-27
Improvements to frameworks.md and added postinstall script

igniteram:
- Broken links in frameworks.md should be fixed with https://github.com/angular/protractor/commit/3d14cc10bdf6dce3f932ffa77a193a62072927f2. closes https://github.com/angular/protractor/issues/3671
- Added sample `cucumberConf.js` so that it could be referenced in frameworks.md https://github.com/angular/protractor/commit/61ae390c20fe3e7de3c32c2b407246ed5a850b3e, would be happy to  send a PR with protractor-cucumber tests for our testapp as well.
- Added a `postinstall` script in package.json as it would help remove the redundancy of manually installing dependencies(website,testapp & webdriver-manager update) for contributors. Updated the same in `README.md` https://github.com/angular/protractor/commit/e6187541285c0cfa9597c3c1632eecf21c6a1b4d

These could be squashed & merged into one commit if needed, let's see what travis has to say 😄 !


igniteram:
I see some changes in `README.md` so this PR would have to be rebased with master!

package.json
```diff
@@ -62,10 +62,12 @@
   "main": "built/index.js",
   "typings": "built/index.d.ts",
   "scripts": {
+    "postinstall": "cd bin && webdriver-manager update && npm run install_website && npm run install_testapp",
     "prepublish": "gulp prepublish",
     "pretest": "gulp pretest",
     "start": "node testapp/scripts/web-server.js",
     "test": "node scripts/test.js",
+    "install_website":"cd website && npm install",
```
sjelin: Why would someone outside the core team want to install the website?

spec/cucumberConf.js
```diff
@@ -0,0 +1,35 @@
+var env = require('./environment');
```
sjelin: We do not provide first party cucumber support, and the `.feature` files referenced here do not exist

package.json
```diff
@@ -62,10 +62,12 @@
   "main": "built/index.js",
   "typings": "built/index.d.ts",
   "scripts": {
+    "postinstall": "cd bin && webdriver-manager update && npm run install_website && npm run install_testapp",
```
sjelin: Adding a `postinstall` like this will cause anyone who used `npm install protractor` to install the testapp - something almost no one wants and won't even work since the testapp is ignored via `.npmignore`.

README.md
```diff
@@ -40,11 +40,11 @@ Clone the github repository:
...
     cd protractor
+
+Install all the dependencies in the folder:
+
     npm install
-    ./bin/webdriver-manager update
-    cd website
-    npm install
-    cd ..
+
```
cnishina: These instructions should be left in. These instructions are important for contributors.

We could do an `npm run install_website` here instead.

README.md
```diff
@@ -54,11 +54,7 @@ Start up a selenium server. By default, the tests expect the selenium server to
 
 Protractor's test suite runs against the included test application.
 
-Install the test application dependencies with:
-
-    npm run install_testapp
-
-Start that up with
+Start the test application with
```
cnishina: This is also important to get started for contributing. The user should install the angular 2 dependencies and serve the Angular app.

package.json
cnishina: I am also not a fan of this step. We should avoid `postinstall` because not everyone is a contributor.

spec/cucumberConf.js
igniteram: I know we do not provide first party cucumber support and there are no feature files, but users who use cucumber would have a look at the frameworks.md file in which we have mentioned a link to a sample cucumber conf.js. we need to either remove it or add a sample conf file. I would suggest let's remove it and i could add a reference link to our protractor-cookbook where users can have a look at the cucumber conf file.

package.json
igniteram: Yes I get it, users would not need it and not everyone is a contributor, my intention was to simplify things for contributors as its a 3 step npm install. I would remove this!

README.md
igniteram: This could be done! let me clean up things hear :+1:

spec/cucumberConf.js
sjelin: You could put something in `docs/`, just make sure it makes clear we don't provide first-party support

igniteram: I have added a basic cucumber config file in `docs/frameworks.md` so that users can refer to it.

README.md
igniteram: Reverted back and removed post_install and I see you have already added `npm run install_website` !

igniteram:
@cnishina can you please review and close this!

cnishina:
LGTM! Thanks @igniteram

PR #3682, opened 2016-10-28
chore(clang-format): increase ColumnLimit to 100

thorn0:
Does anybody use an 80x25 hardware text-mode terminal to edit this code? Why is this value so low? It doesn't help the readability at all.

The Angular 2 project has [has](https://github.com/angular/angular/blob/master/.clang-format) `ColumnLimit: 100`.


cnishina:
It looks like our Circle CI is having issues. LGTM.

PR #3685, opened 2016-10-29
chore(types): make plugins.ts more strongly-typed

thorn0:
Some code reformatting happened here because of changing `ColumnLimit` in `.clang-format`. I can extract it to a separate PR if needed. The main idea of this PR is to use the type definitions for the `selenium-webdriver` module and make the compiler infer better types for the members of the `Plugins` class.

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -1,4 +1,4 @@
-import {ActionSequence, Capabilities, Command as WdCommand, FileDetector, Options, promise as wdpromise, Session, TargetLocator, TouchSequence, until, WebDriver, WebElement} from 'selenium-webdriver';
+import {ActionSequence, By, Capabilities, Command as WdCommand, FileDetector, ICommandName, Options, promise as wdpromise, Session, TargetLocator, TouchSequence, until, WebDriver, WebElement} from 'selenium-webdriver';
 import * as url from 'url';
 
 import {DebugHelper} from './debugger';
@@ -9,10 +9,11 @@ import {Locator, ProtractorBy} from './locators';
 import {Logger} from './logger';
 import {Plugins} from './plugins';
 
-let clientSideScripts = require('./clientsidescripts');
-let webdriver = require('selenium-webdriver');
-let Command = require('selenium-webdriver/lib/command').Command;
-let CommandName = require('selenium-webdriver/lib/command').Name;
+const clientSideScripts = require('./clientsidescripts');
+const webdriver = require('selenium-webdriver');
+// TODO: fix the typings for selenium-webdriver/lib/command
+const Command = require('selenium-webdriver/lib/command').Command as typeof WdCommand;
+const CommandName = require('selenium-webdriver/lib/command').Name as ICommandName;
 
 // jshint browser: true
 
@@ -244,7 +245,7 @@ export class ProtractorBrowser extends Webdriver {
    * @type {Array<{name: string, script: function|string, args:
    * Array.<string>}>}
    */
-  mockModules_: any[];
+  mockModules_: {name: string, script: string|Function, args: any[]}[];
 
   /**
    * If specified, start a debugger server at specified port instead of repl
@@ -279,8 +280,8 @@ export class ProtractorBrowser extends Webdriver {
     let methodsToSync = ['getCurrentUrl', 'getPageSource', 'getTitle'];
 
     // Mix all other driver functionality into Protractor.
-    Object.getOwnPropertyNames(webdriver.WebDriver.prototype).forEach((method: string) => {
-      if (!this[method] && typeof(webdriverInstance as any)[method] == 'function') {
+    Object.getOwnPropertyNames(WebDriver.prototype).forEach(method => {
+      if (!this[method] && typeof(webdriverInstance as any)[method] === 'function') {
         if (methodsToSync.indexOf(method) !== -1) {
           ptorMixin(this, webdriverInstance, method, this.waitForAngular.bind(this));
         } else {
@@ -291,8 +292,8 @@ export class ProtractorBrowser extends Webdriver {
 
     this.driver = webdriverInstance;
     this.element = buildElementHelper(this);
-    this.$ = build$(this.element, webdriver.By);
-    this.$$ = build$$(this.element, webdriver.By);
+    this.$ = build$(this.element, By);
+    this.$$ = build$$(this.element, By);
     this.baseUrl = opt_baseUrl || '';
     this.rootEl = opt_rootElement || 'body';
     this.ignoreSynchronization = false;
@@ -442,7 +443,7 @@ export class ProtractorBrowser extends Webdriver {
 
     let runWaitForAngularScript: () => wdpromise.Promise<any> = () => {
       if (this.plugins_.skipAngularStability()) {
-        return webdriver.promise.fulfilled();
+        return wdpromise.fulfilled();
       } else if (this.rootEl) {
         return this.executeAsyncScript_(
             clientSideScripts.waitForAngular, 'Protractor.waitForAngular()' + description,
@@ -472,9 +473,7 @@ export class ProtractorBrowser extends Webdriver {
                   .then(() => {
                     return this.driver.wait(() => {
                       return this.plugins_.waitForCondition().then((results: boolean[]) => {
-                        return results.reduce((x, y) => {
-                          return x && y;
-                        }, true);
+                        return results.reduce((x, y) => x && y, true);
                       });
                     }, this.allScriptsTimeout, 'Plugins.waitForCondition()');
                   });
@@ -580,8 +579,7 @@ export class ProtractorBrowser extends Webdriver {
    * Add a module to load before Angular whenever Protractor.get is called.
    * Modules will be registered after existing modules already on the page,
    * so any module registered here will override preexisting modules with the
-   * same
-   * name.
+   * same name.
    *
    * @example
    * browser.addMockModule('modName', function() {
@@ -629,9 +627,7 @@ export class ProtractorBrowser extends Webdriver {
    * @returns {Array.<!string|Function>} The list of mock modules.
    */
   getRegisteredMockModules(): Array<string|Function> {
-    return this.mockModules_.map((module) => {
-      return module.script;
-    });
+    return this.mockModules_.map(module => module.script);
   };
 
   /**
@@ -661,9 +657,7 @@ export class ProtractorBrowser extends Webdriver {
    * @param {number=} opt_timeout Number of milliseconds to wait for Angular to
    *     start.
    */
-  get(destination: string, opt_timeout?: number) {
-    let timeout = opt_timeout ? opt_timeout : this.getPageTimeout;
-
+  get(destination: string, timeout = this.getPageTimeout) {
     destination = this.baseUrl.indexOf('file://') === 0 ? this.baseUrl + destination :
                                                           url.resolve(this.baseUrl, destination);
     let msg = (str: string) => {
@@ -672,17 +666,14 @@ export class ProtractorBrowser extends Webdriver {
 
     if (this.ignoreSynchronization) {
       this.driver.get(destination);
-      return this.driver.controlFlow().execute(() => {
-        return this.plugins_.onPageLoad();
-      });
+      return this.driver.controlFlow().execute(() => this.plugins_.onPageLoad()).then(() => {});
     }
 
-    let deferred = webdriver.promise.defer();
+    let deferred = wdpromise.defer<void>();
 
     this.driver.get(this.resetUrl).then(null, deferred.reject);
     this.executeScriptWithDescription(
             'window.name = "' + DEFER_LABEL + '" + window.name;' +
-
                 'window.location.replace("' + destination + '");',
             msg('reset url'))
         .then(null, deferred.reject);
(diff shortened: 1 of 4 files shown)


lib/plugins.ts
```diff
@@ -1,7 +1,15 @@
-var webdriver = require('selenium-webdriver');
-import {Logger} from './logger';
 import * as Q from 'q';
```
cnishina: initially I changed this from q to Q. This could be set to lowercase.

lib/plugins.ts
```diff
@@ -1,7 +1,15 @@
...
+import * as webdriver from 'selenium-webdriver';
+
 import {ConfigParser} from './configParser';
+import {Logger} from './logger';
+
+declare module 'q' {
```
cnishina: If this is a missing typing in `@types/q`, then this should be filed against DefinitelyTyped types-2.0 branch.  Also, a TODO should be added to remove this when the DefinitelyTyped PR is merged.

lib/plugins.ts
```diff
@@ -490,38 +482,29 @@ export class Plugins {
...
+  /**
+   * Generates the handler for a plugin function (e.g. the setup() function)
+   *
+   * @param {string} funName The name of the function to make a handler for
+   * @param {PromiseType} promiseType The type of promise (WebDriver or Q) that should be used
+   * @param {boolean=} failReturnVal The value that the function should return if the plugin crashes
+   *
+   * @return The handler
+   */
+  private pluginFunFactory(funName: string, promiseType: PromiseType.Q, failReturnVal?: boolean):
```
cnishina: `promiseType: PromiseType`. This could either be WebDriver or Q. I see some benefit defining several times to get the correct return type. If we define it this way:

```
private pluginFunFactory(funName: string, promsieType: PromiseType, 
    failReturnVal?: boolean): 
    (...args: any[]) => webdriver.promise.Promise<any[]> |
    (...args: any[]) => Q.Promise<any[]> {
  ...
```

does it have negative effects on the typings?

cnishina:
@thorn0 thanks for another great PR. Just a few comments.

lib/plugins.ts
thorn0: It's a deprecated and undocumented method that Protractor still uses. It can be difficult to get the reviewers to accept such a PR as they usually want to see a link to API docs. So I'm not sure it'd be the right thing to do.

thorn0:
> Some of these changes are clang related rather than making the plugins.ts more strongly-typed. This might be better as two separate PRs.

In the separate PR, would it better if I reformat just this one file or all the files?

lib/plugins.ts
thorn0: I'll look into whether we can just not use this method at all.

cnishina:
Thanks! Just saw the separate PR https://github.com/angular/protractor/pull/3686.

lib/plugins.ts
thorn0: The most elegant solution would be not to use the `PromiseType` enum at all, and instead make the function generic and pass the promise implementation itself (`q` or `webdriver.promise`) instead of an enum value. It'd look like this:
```ts
private pluginFunFactory<T extend typeof q | typeof webdriver.promise>(
  funName: string, promiseImplementation: T, failReturnVal?: boolean) {
...
}
```
Unfortunately, this doesn't work because of this: https://github.com/Microsoft/TypeScript/issues/11940

lib/plugins.ts
thorn0: I did submit a [PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/12358) to DefinitelyTyped.

cnishina:
Please rebase.

thorn0:
@cnishina please have a look

lib/browser.ts
```diff
@@ -672,17 +673,14 @@ export class ProtractorBrowser extends Webdriver {
 
     if (this.ignoreSynchronization) {
       this.driver.get(destination);
-      return this.driver.controlFlow().execute(() => {
-        return this.plugins_.onPageLoad();
-      });
+      return this.driver.controlFlow().execute(() => this.plugins_.onPageLoad()).then(() => {});
```
cnishina: `.then(() => {})` This appears to not do anything. Should this be removed?

lib/plugins.ts
```diff
@@ -436,11 +428,21 @@ export class Plugins {
    * @return {webdriver.promise.Promise|Q.Promise} A promise which resolves to the
    *     function's return value
    */
-  safeCallPluginFun(
-      pluginObj: ProtractorPlugin, funName: string, args: IArguments, promiseType: PromiseType,
-      failReturnVal: any): any {
-    var deferred = promiseType == PromiseType.Q ? Q.defer() : webdriver.promise.defer();
-    var logError = (e: any) => {
+  private safeCallPluginFun(
```
cnishina: @sjelin could you review the `safeCallPluginFun` changes?

lib/element.ts
```diff
@@ -1061,7 +1061,7 @@ export class ElementFinder extends WebdriverWebElement {
  * @returns {ElementFinder} which identifies the located
  *     {@link webdriver.WebElement}
  */
-export let build$ = (element: ElementHelper, by: ProtractorBy) => {
+export let build$ = (element: ElementHelper, by: typeof By) => {
```
cnishina: Let's revert this to `ProtractorBy`. `ProtractorBy` is composed with `By` functionality and currently does not inherit from `By`.

lib/element.ts
```diff
@@ -1092,7 +1092,7 @@ export let build$ = (element: ElementHelper, by: ProtractorBy) => {
  * @returns {ElementArrayFinder} which identifies the
  *     array of the located {@link webdriver.WebElement}s.
  */
-export let build$$ = (element: ElementHelper, by: ProtractorBy) => {
+export let build$$ = (element: ElementHelper, by: typeof By) => {
```
cnishina: Same here.

lib/plugins.ts
```diff
@@ -257,41 +259,42 @@ export interface ProtractorPlugin {
...
-      if (path) {
-        pluginObj = (<ProtractorPlugin>require(path));
-      } else if (pluginConf.inline) {
-        pluginObj = pluginConf.inline;
-      } else {
-        throw new Error(
-            'Plugin configuration did not contain a valid path or ' +
-            'inline definition.');
-      }
+        var pluginObj: ProtractorPlugin;
```
cnishina: nit: use `let` over `var`

lib/plugins.ts
```diff
@@ -1,7 +1,9 @@
-var webdriver = require('selenium-webdriver');
-import {Logger} from './logger';
-import * as Q from 'q';
+import * as q from 'q';
```
cnishina: Thanks for fixing this. This is leftover from experimenting to get the 'Q' module to work with our typings.

lib/plugins.ts
```diff
@@ -380,14 +377,11 @@ export class Plugins {
    * @return {Object} The results object
    */
   getResults() {
-    var results: {failedCount: number, specResults: any[]} = {failedCount: 0, specResults: []};
+    var results = {failedCount: 0, specResults: [] as SpecResult[]};
```
cnishina: nit: Let's fix this too. Use `let` over `var`

lib/plugins.ts
```diff
@@ -400,24 +394,22 @@ export class Plugins {
    * @return {boolean}
    */
   skipAngularStability() {
-    var result = this.pluginObjs.some((pluginObj: ProtractorPlugin) => {
-      return pluginObj.skipAngularStability;
-    });
+    var result = this.pluginObjs.some(pluginObj => pluginObj.skipAngularStability);
```
cnishina: nit: Same nit fix for `let` over `var`

lib/plugins.ts
thorn0: However, in the [docs](https://github.com/kriskowal/q/wiki/API-Reference) for that module, the uppercase Q is used...

lib/browser.ts
thorn0: That's a good question. First of all, it doesn't do nothing, it converts `Promise<any[]>` to `Promise<void>`, so that the return type of `ProtractorBrowser#get` is `Promise<void>`.

`onPageLoad` used to be typed simply as `Function`, but now that it's type is inferred to be `(...args: any[]) => webdriver.promise.Promise<any[]>`, it became apparent that the returned promise is resolved with an array (the result of the `.all(...)` call in `pluginFunFactory`), and this array seems to be totally useless. Moreover, the presence of this array can also have unexpected effects in run time. E.g. we may pass a function to `then` expecting that its argument would be undefined whereas in fact it'd get an array.

That's why I'm thinking about moving `.then(() => {})` from here to `Plugins#pluginFunFactory`, so that it creates functions of type `(...args: any[]) => q.Promise<void>`. What are your thoughts on this? Are probably among those functions (`setup`, `onPrepare`, etc.) such whose resolved array value isn't meaningless?

thorn0: Okay, I found that the array is meaningful for `waitForCondition`... It's used in `browser.ts`:
```ts
return this.plugins_.waitForCondition().then((results: boolean[]) => {
  return results.reduce((x, y) => {
    return x && y;
  }, true);
});
```
Which means `.then(() => {})` can't be moved to `pluginFunFactory`.

sjelin: None of the functions generated by `pluginFunFactory` are ever exposed to protractor users, so I'm not bothered by the fact that the promise resolves to something useless.  Instead of this `.then()` block - which is really only here for type checking - why not just cast this to a `() => void`?

thorn0: These functions are exposed indirectly. `ProtractorBrowser#get` is exposed. The very line we're discussing now is its return statement. So it's not just type checking, it's public API.

lib/element.ts
thorn0: The thing is this function is actually called with WebDriver's `By`, not with Protractor's one. See `browser.ts`:
```ts
this.$ = build$(this.element, By);
this.$$ = build$$(this.element, By);
```
Hence this change.

lib/plugins.ts
```diff
@@ -436,11 +428,21 @@ export class Plugins {
...
    */
-  safeCallPluginFun(
-      pluginObj: ProtractorPlugin, funName: string, args: IArguments, promiseType: PromiseType,
-      failReturnVal: any): any {
-    var deferred = promiseType == PromiseType.Q ? Q.defer() : webdriver.promise.defer();
-    var logError = (e: any) => {
+  private safeCallPluginFun(
+      pluginObj: ProtractorPlugin, funName: string, args: any[], promiseType: PromiseType,
+      failReturnVal: any) {
+    let resultPromise: webdriver.promise.Promise<any>|q.Promise<any>;
```
sjelin: Why this way instead of my way?  Couldn't you have just given `deferred` the type `webdriver.promise.Deferred | q.Deferred`?

lib/plugins.ts
```diff
@@ -436,11 +428,21 @@ export class Plugins {
...
+    let done: (result: any) => void;
+    if (promiseType == PromiseType.Q) {
+      const deferred = q.defer<any>();
+      resultPromise = deferred.promise;
+      done = deferred.resolve;
```
sjelin: Does this not have to be bound to the `deferred` object?

lib/plugins.ts
```diff
@@ -436,11 +428,21 @@ export class Plugins {
...
+    } else {
+      const deferred = webdriver.promise.defer<any>();
+      resultPromise = deferred.promise;
+      done = deferred.fulfill;
```
sjelin: Same as above

lib/plugins.ts
thorn0: In both libraries, these methods are bound.

lib/plugins.ts
cnishina: Yup, I think lowercase is fine.

lib/element.ts
cnishina: Yup. I agree with the change. Thanks for this fix.

lib/plugins.ts
cnishina: The `'q'` issue should be fixed when your PR to DefinitelyTyped is merged. Is `fulfill` deprecated (https://github.com/kriskowal/q/blob/v1/q.js#L1141)? I believe that comment should be fixed in your DefinitelyTyped PR (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/12358).

thorn0: Hm, probably you're right. I searched through the [history of edits](https://github.com/kriskowal/q/wiki/API-Reference/_history) of the API reference. It turns out it never was documented. So we can't really call it deprecated.

thorn0: I've updated that PR, but IMHO using undocumented methods isn't a much better thing than using deprecated ones.

lib/browser.ts
sjelin: Good point :smile:

lib/browser.ts
```diff
@@ -596,7 +594,7 @@ export class ProtractorBrowser extends Webdriver {
    *     the script and may be referenced using the `arguments` object.
    */
   addMockModule(name: string, script: string|Function, ...moduleArgs: any[]) {
-    this.mockModules_.push({name: name, script: script, args: moduleArgs});
+    this.mockModules_.push({name, script, args: moduleArgs});
```
juliemr: I dislike the shorthand here, let's continue to use the more explicit form.

lib/browser.ts
```diff
@@ -661,9 +657,7 @@ export class ProtractorBrowser extends Webdriver {
    * @param {number=} opt_timeout Number of milliseconds to wait for Angular to
    *     start.
    */
-  get(destination: string, opt_timeout?: number) {
-    let timeout = opt_timeout ? opt_timeout : this.getPageTimeout;
-
+  get(destination: string, opt_timeout = this.getPageTimeout) {
```
juliemr: This should just be called `timeout` now.

lib/plugins.ts
```diff
@@ -308,8 +311,8 @@ export class Plugins {
                 'reported.');
           }
           info = info || {};
-          var specName = info.specName || (obj.name + ' Plugin Tests');
-          var assertion: any = {passed: passed};
+          const specName = info.specName || (obj.name + ' Plugin Tests');
+          const assertion: AssertionResult = {passed};
```
juliemr: Here too with the object shortcut.

juliemr:
LGTM with cleanups.

thorn0:
done

Issue #3606, opened 2016-09-28
how i can run protractor using localhost


MahmoudSamik:
i'm running my AngularJs application on port 80 
so when i run my test i ask him to go to the URL 
browser.get('http://localhost/#/user/login');
but the application is not working right its just loading and nothing happen after that

mgiambalvo:
Hi there!

Since it involves usage and debugging (and you haven't used the [issue template](https://github.com/angular/protractor/blob/master/.github/ISSUE_TEMPLATE) to indicate otherwise), your question is better suited for StackOverflow or Gitter. Please [ask a question there with the 'protractor' tag](http://stackoverflow.com/questions/ask?tags=protractor) or post in the [Gitter Channel](https://gitter.im/angular/protractor) to get help.

From the the [getting help](https://github.com/angular/protractor#getting-help) section of the README:

> Please ask usage and debugging questions on StackOverflow (use the "protractor" tag) or in the Angular discussion group. (Please do not ask support questions here on Github.)

Thanks and good luck!

PR #3715, opened 2016-11-08
chore(deps): Downgrade jasmine to 2.4.1

mgiambalvo:
Upgrading to 2.5.2 causes #3606. We'll need to wait on a fix for
jasmine/jasmine-npm#95 before upgrading.


dreef3:
I had some code that used `jasmine.getEnv().clearReporters()`. This PR broke protractor startup because the method was added in 2.5.x

Issue #3605, opened 2016-09-28
Protractor 4.0.9 breaks benchpress


cnishina:
After investigation of https://github.com/angular/angular/pull/11816, we are breaking the angular benchpress.
- [ ] Add a sample benchpress into Protractor
- [ ] Investigate / fix the sample benchpress

Issue #3627, opened 2016-10-07
Including protractor in a protractor.conf.ts file breaks using importing protractor.browser and more


KrauseStefan:
Bug report
---

- Node Version: `6.7.0`
- Protractor Version: `4.0.9`
- Operating System and Version `Win 8.1`

Output from running the test
```
[14:04:55] I/direct - Using ChromeDriver directly...                                                                
[14:04:55] I/launcher - Running 1 instances of WebDriver                                                            
[14:04:57] E/launcher - Error: TypeError: Cannot read property 'manage' of undefined                                
    at onPrepare (D:\git\hq\SitaWareWeb\client\src\main\client_integration_test\protractor.conf.ts:64:12)           
    at D:\git\hq\SitaWareWeb\client\src\main\node_modules\protractor\built\util.js:48:49                            
    at Function.promise (D:\git\hq\SitaWareWeb\client\src\main\node_modules\q\q.js:682:9)                           
    at Object.runFilenameOrFn_ (D:\git\hq\SitaWareWeb\client\src\main\node_modules\protractor\built\util.js:38:16)  
    at D:\git\hq\SitaWareWeb\client\src\main\node_modules\protractor\built\runner.js:93:27                          
    at _fulfilled (D:\git\hq\SitaWareWeb\client\src\main\node_modules\q\q.js:834:54)                                
    at self.promiseDispatch.done (D:\git\hq\SitaWareWeb\client\src\main\node_modules\q\q.js:863:30)                 
    at Promise.promise.promiseDispatch (D:\git\hq\SitaWareWeb\client\src\main\node_modules\q\q.js:796:13)           
    at D:\git\hq\SitaWareWeb\client\src\main\node_modules\q\q.js:556:49                                             
    at runSingle (D:\git\hq\SitaWareWeb\client\src\main\node_modules\q\q.js:137:13)                                 
[14:04:57] E/launcher - Process exited with error code 100                                                          
```

Protractor.conf.ts
```typescript
import {UtilityFunctions as util} from './UtilityFunctions';
//import * as protractor from 'protractor';
import {browser, Config} from 'protractor';

const driverInfo = util.getDriverInfo(protractorVersion);

export const config: Config = {

  specs: [
    'client_integration_test/build/**/test/ui/**/*.test.js'
  ],
  framework: 'jasmine',

  beforeLaunch: () => {
    launchWebServer();
  },

  onPrepare: () => {
   // patchProtractorObject();

    browser.manage().window().maximize();
[...]

PR #3716, opened 2016-11-09
fix(globals): Resolve exported globals dynamically.

mgiambalvo:
Because globals are set in the runner, they may be undefined at import time. Also, the `browser` global object will change if `browser.reset()` is called. By making these globals exported as properties, they will be resolved dynamically. This also fixes issues importing protractor in config files written in typescript, since the globals are resolved at call time instead of at import time.

Note that until the Runner calls `setupGlobals_`, these exports will be undefined. That's just the nature of these objects, though, and the global objects have the same behavior.

### DIFF
--- a/spec/basic/restart_spec.js
+++ b/spec/basic/restart_spec.js
@@ -1,10 +1,10 @@
-fdescribe('browser.restart', function() {
-  fit('doesn\'t break ignoreSynchronization', function() {
+describe('browser.restart', function() {
+  it('doesn\'t break ignoreSynchronization', function() {
     browser.get('index.html#/polling');
     browser.restart();
 
     browser.ignoreSynchronization = true;
     // Get a non-angular page. It shouldn't fail if ignoreSynchronization is on.
-    browser.get("https://google.com/");
+    browser.get('https://google.com/');
   });
 });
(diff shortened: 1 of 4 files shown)


spec/basic/restart_spec.js
```diff
@@ -0,0 +1,10 @@
+fdescribe('browser.restart', function() {
+  fit('doesn\'t break ignoreSynchronization', function() {
+    browser.get('index.html#/polling');
+    browser.restart();
+
+    browser.ignoreSynchronization = true;
+    // Get a non-angular page. It shouldn't fail if ignoreSynchronization is on.
+    browser.get("https://google.com/");
```
cnishina: Nit: should use single quotes.

spec/basic/restart_spec.js
```diff
@@ -0,0 +1,10 @@
+fdescribe('browser.restart', function() {
```
cnishina: Probably do not need fdescribe and fit

mgiambalvo:
Also fixes #3605

KrauseStefan:
Great work, happy that this made it in. :thumbsup:
This solved my issue #3627 thank you

PR #3723, opened 2016-11-09
docs(timeout): Update timeout error message

mgiambalvo:
Also changed the way headers work in the timeouts doc, so now the error message has a direct link to the right spot.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -25,7 +25,7 @@ with more info.
 Angular can't be found on my page
 ---------------------------------
 
-Protractor supports angular 1.5.x and higher - check that your version of Angular is upgraded.
+Protractor supports angular 1.0.6/1.1.4 and higher - check that your version of Angular is upgraded.
 
 The `angular` variable is expected to be available in the global context. Try opening chrome devtools or firefox and see if `angular` is defined.
 
@@ -35,9 +35,9 @@ How do I deal with my log-in page?
 If your app needs log-in, there are a couple ways to deal with it. If your login
 page is not written with Angular, you'll need to interact with it via 
 unwrapped webdriver, which can be accessed like `browser.driver.get()`. You can also use
-`browser.ignoreSynchronization`, as explained in [this StackOverflow answer](http://stackoverflow.com/a/23198865/264229)
+`browser.ignoreSynchronization` as explained [here](/docs/timeouts.md#how-to-disable-waiting-for-angular).
 
-You can also put your log-in code into an `onPrepare` function, which will be run
+Another option is to put your log-in code into an `onPrepare` function, which will be run
 once before any of your tests. See this example ([withLoginConf.js](https://github.com/angular/protractor/blob/master/spec/withLoginConf.js))
 
 Which browsers are supported?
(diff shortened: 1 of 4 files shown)


docs/faq.md
```diff
@@ -25,7 +25,7 @@ with more info.
 Angular can't be found on my page
 ---------------------------------
 
-Protractor supports angular 1.0.6/1.1.4 and higher - check that your version of Angular is upgraded.
+Protractor supports angular 1.5.x and higher - check that your version of Angular is upgraded.
```
juliemr: Not sure we should change this yet - we should do this in an official deprecation and remove the code from clientSideScripts alongside changing the messaging.

docs/faq.md
```diff
@@ -34,9 +34,10 @@ How do I deal with my log-in page?
 
 If your app needs log-in, there are a couple ways to deal with it. If your login
 page is not written with Angular, you'll need to interact with it via 
-unwrapped webdriver, which can be accessed like `browser.driver.get()`. 
+unwrapped webdriver, which can be accessed like `browser.driver.get()`. You can also use
+`browser.ignoreSynchronization`, as explained in [this StackOverflow answer](http://stackoverflow.com/a/23198865/264229)
```
juliemr: Hah, we should probably have our own docs. But I'm fine with this for now, since I can't think off the top of my head where it would best live.

docs/timeouts.md
```diff
@@ -17,48 +17,49 @@ be loaded and the new URL to appear before continuing.
...
+Protractor waits for the `angular` variable to be present when loading a new page.
 
  - Looks like: an error in your test results - `Angular could not be found on the page: retries looking for angular exceeded`
 
  - Default timeout: 10 seconds
 
  - How to change: To change globally, add `getPageTimeout: timeout_in_millis` to your Protractor configuration file. For an individual call to `get`, pass an additional parameter: `browser.get(address, timeout_in_millis)`
 
-_*How to disable waiting for Angular*_
+### _How to disable waiting for Angular_
```
juliemr: Maybe above, we should link to this instead of the SO page?

docs/timeouts.md
```diff
@@ -17,48 +17,49 @@ be loaded and the new URL to appear before continuing.
...
-$interval service ([interval.js](https://github.com/angular/angular.js/blob/master/src/ng/interval.js)) for anything that polls continuously (introduced in Angular 1.2rc3).
+Before performing any action, Protractor waits until there are no pending asynchronous tasks in your Angular application. This means that all timeouts and http requests are finished. If your application continuously polls $timeout or $http, Protractor will wait indefinitely and time out. You should use the
+[$interval](https://github.com/angular/angular.js/blob/master/src/ng/interval.js) for anything that polls continuously (introduced in Angular 1.2rc3).
 
- - Looks like: an error in your test results - `Timed out waiting for Protractor to synchronize with the page after 11 seconds.`
+ - Looks like: an error in your test results - `Timed out waiting for asynchronous Angular tasks to finish after 11 seconds.`
 
  - Default timeout: 11 seconds
 
  - How to change: Add `allScriptsTimeout: timeout_in_millis` to your Protractor configuration file.
```
juliemr: Let's add a 'You can also disable waiting for Angular, see below' (and have a link to that header)

mgiambalvo: Done

juliemr:
Looks good!

PR #3724, opened 2016-11-10
chore(gulpfile): remove format from install

cnishina:
- will allow circle ci to do a format check


cnishina:
this should fail with format fixes

PR #3727, opened 2016-11-10
DO NOT MERGE -- chore(direct): fix directConnect for the last chromedriver downloaded

cnishina:


cnishina:
Should fail since webdriver-manager is not published.

cnishina:
With webdriver-manager@10.2.9: direct connect will read the update-config.json to locate the chromedriver binaries.... will have another PR to fix local driver provider to do similar.

circle.yml
```diff
@@ -1,10 +1,10 @@
 machine:
+  node:
+    version: 6.9.1
```
sjelin: I don't see that version in https://circleci.com/docs/build-image-trusty/#nodejs ?

lib/driverProviders/direct.ts
```diff
@@ -57,15 +55,12 @@ export class Direct extends DriverProvider {
    */
   getNewDriver(): webdriver.WebDriver {
     let driver: webdriver.WebDriver;
+    let updateJson = path.resolve(SeleniumConfig.getSeleniumDir(), 'update-config.json');
```
sjelin: We should document this somewhere in wdm

circle.yml
cnishina: 6.9.1 is the latest stable and Circle CI doesn't complain about it. For some reason I was having issues with spread syntax. This change will also happen when we adopt selenium-webdriver 3.

lib/driverProviders/direct.ts
cnishina: Yes, I'll create an issue to remind myself in wdm

cnishina:
Since this relies on webdriver-manager (11.0.0) and requires node 6, this PR should be against the beta branch for the next release.

PR #3726, opened 2016-11-10
fix(element): make Element[Array]Finder actually extend WebElement

thorn0:
An attempt to fix #3721. A change to [angular/jasminewd](https://github.com/angular/jasminewd) is needed for this to work:

**[index.js](https://github.com/angular/jasminewd/blob/35d2ac86623d79f1a3d99a60933ec30986253139/index.js):**
```diff
 var originalExpect = global.expect;
 global.expect = function(actual) {
-  if (actual instanceof webdriver.WebElement) {
+  if (actual instanceof webdriver.WebElement && typeof actual.then !== 'function' && actual.then !== null) {) {
     throw Error('expect called with WebElement argument, expected a Promise. ' +
         'Did you mean to use .getText()?');
   }
```

Issue #3721, opened 2016-11-09
types: ElementFinder isn't compatible with browser.actions()


thorn0:
This line doesn't compile:
```ts
browser.actions().doubleClick(element(by.css('.row_0'))).perform();
```
Error:
```
Argument of type 'ElementFinder' is not assignable to parameter of type 'string | WebElement | undefined'.
       Type 'ElementFinder' is not assignable to type 'WebElement'.
         Types of property 'findElement' are incompatible.
           Type '(subLocator: Locator) => Promise<any>' is not assignable to type '(locator: Function | By) => WebElementPromise'.
             Type 'Promise<any>' is not assignable to type 'WebElementPromise'.
               Property 'getDriver' is missing in type 'Promise<any>'.
```
Seems like the core problem is that `ElementFinder` isn't a subtype of `WebElement` (from `selenium- webdriver`).

cnishina:
```
browser.actions().doubleClick(element(by.css('.row_0')).getWebElement()).perform();
```

We are exploring this; however, there is a way to get around it via the existing API

cnishina:
What do you think @juliemr?

thorn0:
For #3726 to fix this, we need to modify jasminewd a bit.

thorn0:
As an alternative, see #3732.

thorn0:
#3732 has been merged

PR #3732, opened 2016-11-10
chore(types): make TS think that Element[Array]Finder extends WebElement

thorn0:
Another approach to #3721. The alternative is #3726.


mgiambalvo:
Thanks!

PR #3758, opened 2016-11-22
chore(tests): Testapp with `@angular/upgrade/static` (ngUpgrade for AoT apps)

sjelin:
These kids and their new fangled "ahead of time compilation."  They're upgrading
their apps using `@angular/upgrade/static` now.  I guess we should get with the
times and run tests against a webapp running this hip new API.


juliemr:
Looks good. Thanks!

PR #3766, opened 2016-11-22
fix(ExpectedConditions): non-static ExpectedConditions for browser

### LINKED ISSUE #3761: Issue with ExpectedConditions working with forked browser instances
Bug report
---

- Node Version: `v6.9.1`
- Protractor Version: `4.0.11`
- Angular Version: `1.5.6`
- Browser(s): `chrome`
- Operating System and Version `mac osx`

Code of my test is as following:

```

beforeAll(function(){
  browser2=browser.forkNewDriverInstance();
  return user.login(browser2);
});

beforeEach(function(){
  return browser2.get('/other-page');
});
```

and user.login() is:

```
User.prototype.login=function(browserToUse){
  browser.runLogin()
    .then(function(){
      return browserToUse.wait(EC.urlContains('/home'), 5000*2,'Waiting ofr URL /home');
    });
}
```

It looks like that it executes instructions in the wrong order, that is:

1. user.login(b)
2. browser2.get('/other-page')
3. browserToUse.wait(EC.urlContains('/home'), 5000*2,'Waiting ofr URL /home');

indeed, the 3rd instruction generates a time out exception. But the browser goes through /home URL, it is a problem of synchronization.

I wonder: how is it possible? 

Which is the way to tell protractor: don't use your algorithm for ordering instructions, but I manage the order?

I am using `.then()` for each instruction, but it does not seem enough.

It is worth to note that:
* this test (that includes the beforeAll) have other tests (describe and it) and that before it
* when i run this test by itself, it does not go in timeout
* when i run the whole file, it goes in timeout

cnishina:
closes #3761

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -131,7 +131,7 @@ export class ProtractorBrowser extends Webdriver {
   /**
    * @type {ExpectedConditions}
    */
-  static ExpectedConditions = new ProtractorExpectedConditions();
+  ExpectedConditions: ProtractorExpectedConditions;
 
   /**
    * The wrapped webdriver instance. Use this to interact with pages that do
@@ -320,6 +320,9 @@ export class ProtractorBrowser extends Webdriver {
     this.trackOutstandingTimeouts_ = !opt_untrackOutstandingTimeouts;
     this.mockModules_ = [];
     this.addBaseMockModules_();
+
+    // set up expected conditions
+    this.ExpectedConditions = new ProtractorExpectedConditions(this);
   }
 
   /**
--- a/lib/expectedConditions.ts
+++ b/lib/expectedConditions.ts
@@ -1,5 +1,5 @@
+import {ProtractorBrowser} from './browser';
 import {ElementFinder} from './element';
-import {Ptor} from './ptor';
 
 let webdriver = require('selenium-webdriver');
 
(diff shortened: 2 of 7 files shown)


cnishina:
It looks like:

* The basic tests pass locally and on sauce.
* when Protractor shuts down, `webdriver.quit` throws an error.
* After some investigation, forking a browser instance is causing an error only in sauce labs with the `spec/basic/expected_conditions_spec.js`:

```
var browser2 = browser.forkNewDriverInstance();
browser2.get('index.html#/form');
var EC2 = browser2.ExpectedConditions;
```

My guess is that we are not shutting down properly and the issue might be in `lib/driverProviders/driverProviders.ts` and `lib/driverProviders/sauce.ts`...

cnishina:
Also able to reproduce with `driverProviders/local` :confused:

cnishina:
In a related issue https://github.com/angular/protractor/issues/3772

lib/browser.ts
```diff
@@ -320,6 +320,9 @@ export class ProtractorBrowser extends Webdriver {
     this.trackOutstandingTimeouts_ = !opt_untrackOutstandingTimeouts;
     this.mockModules_ = [];
     this.addBaseMockModules_();
+
+    // set up expected conditions
```
juliemr: Nit: comments should be sentences.

cnishina: :disappointed: Will fix in next PR...

Issue #2138, opened 2015-05-13
Protractor invisibilityOf flakiness, throwing NoSuchElementError


hgted:
I am testing a **non angular** app where I have a login helper function in my tests similar to the following:

	function login() {
			return waitUntilVisible($('#txtUsername'))
				.then(function () {
					$('#txtUsername').sendKeys('tom');
				})
				.then(function () {
					$('#txtPassword').sendKeys('*');
				})
				.then(function () {
					$('#btnLogin').click();
				})
				.then(function () {
                    // this is the flaky line
					return browser.wait(protractor.ExpectedConditions.invisibilityOf($('#txtUsername')), 20000, 'Element never disappeared');
				});
		}

Where `browser.wait(protractor.ExpectedConditions.invisibilityOf($('#txtUsername')), 20000, 'Element never disappeared')` waits for a bootstrap modal containing the login form to close. However, this line is proving to be very flaky, occasionally failing with:

> NoSuchElementError: No element found using locator: By.cssSelector("#txtUsername")

Or even:

> StaleElementReferenceError: stale element reference: element is not attached to the page document

My first though was that the modal was closing before the line was called, however, the docs for `invisibilityOf` suggest it should handle the element not being present, stating it as:

> An expectation for checking that an element is either invisible or not present on the DOM.

**How can I make this line robust?**

hgted:
Switching `invisibilityOf` for `stalenessOf` has done the trick.

FaustXVI:
I have the same issue with an angular app.
Looking at the code (https://github.com/angular/protractor/blob/3.0.0/lib/expectedConditions.js#L332), `invisibilityOf` is defined as :
```
   return this.not(this.visibilityOf(elementFinder));
```

Which seems to be a good idea but `visibilityOf` is defined as :
```
  return this.and(
    this.presenceOf(elementFinder),
    elementFinder.isDisplayed.bind(elementFinder));
```

So, if the element is no more prensent, `invisibilityOf` fails because `visibilityOf` can't found the element.

I think `visibilityOf` should be defined as :

```
  return this.or(
    this.stalenessOf(elementFinder),
    this.not(elementFinder.isDisplayed.bind(elementFinder)));
```

But I'm afraid `elementFinder.isDisplayed` won't work for the same reason. I'll try to get some time to fix that and make a PR but I can't promise anything…

PR #3775, opened 2016-11-30
docs(expectedconditions): Fix description

massimocode:
- See issue https://github.com/angular/protractor/issues/2138 which I just experienced as well because the documentation here (http://www.protractortest.org/#/api?view=ProtractorExpectedConditions.prototype.invisibilityOf) for invisibilityOf said it would support the element not being visible on the DOM, but it doesn't.

### DIFF
--- a/lib/element.ts
+++ b/lib/element.ts
@@ -792,7 +792,9 @@ export class ElementArrayFinder extends WebdriverWebElement {
 export class ElementFinder extends WebdriverWebElement {
   parentElementArrayFinder: ElementArrayFinder;
   elementArrayFinder_: ElementArrayFinder;
-  then: (fn: Function, errorFn?: Function) => wdpromise.Promise<any> = null;
+  then?:
+      (fn: (value: any) => {} | wdpromise.IThenable<any>,
+       errorFn?: (error: any) => any) => wdpromise.Promise<any> = null;
 
   constructor(public browser_: ProtractorBrowser, elementArrayFinder: ElementArrayFinder) {
     super();


massimocode:
The build failed because apparently it stopped responding? Not sure how my changes could have caused that

lib/expectedConditions.ts
```diff
@@ -404,8 +403,8 @@ export class ProtractorExpectedConditions {
   }
 
   /**
-   * An expectation for checking that an element is either invisible or not
-   * present on the DOM. This is the opposite of 'visibilityOf'.
+   * An expectation for checking that an element is invisible on the DOM.
```
cnishina: This should still have "either invisible or not present on the DOM". If `invisibilityOf` returns true, that means `visibility` returned false. So either `presenceOf` is false or `isDisplayed` is false.

cnishina: Please fix.

massimocode: I initially opened this PR exactly to change this piece of text. The minor formatting fix above was just some "gardening".

You are quite right, on closer examination of the source code, the intent of `invisibilityOf` is as you say.

However, I have since opened this issue https://github.com/angular/protractor/issues/3777 because invisibilityOf seems to have a race condition, which is why it cannot be used in its current state for waiting for elements to be removed from the DOM.

sjelin: I would rather fix the bug than document it

massimocode:
Closing this PR as I should create a new PR with the fix @sjelin suggested

PR #3785, opened 2016-12-02
fix(element): Fix typing of then function

massimocode:
Using `await myElemFinder` in typescript returns "any" instead of the `ElementFinder[]` type


cnishina:
Please fix formatting. `gulp format`.

massimocode:
Sorry been a very busy week at work. Have just amended PR.

PR #3788, opened 2016-12-02
WIP - start fixing the debugger w/ selenium 3.0

juliemr:


juliemr:
OK here's where I've got so far (changes pushed):

The initial hooks work fine, but things start to fall apart after a few commands are entered by the user. This is because the empty hook in-between commands isn't getting executed properly - it's listed as blocked in the control flow, and it isn't sent to the selenium server. Then the commands can get sent all at once, causing a very unhappy selenium server.

I'm tempted to say that this just isn't working with the control flow anymore and we need to burn it and re-think how we do debugging.

juliemr:
I think that this incremental update is hopeless - trying a new approach at https://github.com/angular/protractor/pull/3828

juliemr:
Closing in favor of 3828

PR #3791, opened 2016-12-04
Proposal: add seleniumServerStartTimeout in Local DriverProvider

akirakoyasu:
It's nice to specify how long time the Local DriverProvider waits for the selenium-webdriver server to start.

The Local DriverProvider starts local selenium-webdriver server, and waits for the server in 30,000ms now. That's because `DriverService#start()` of selenium-webdriver/remote has an arg `opt_timeoutMs` default 30,000ms.

30,000ms is almost enough long, but that depends on environment.

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -29,6 +29,13 @@ export interface Config {
   seleniumServerJar?: string;
 
   /**
+   * The timeout milliseconds waiting for a local standalone Selenium Server to start.
+   *
+   * default: 30000ms
+   */
+  seleniumServerStartTimeout?: number;
+
+  /**
    * Can be an object which will be passed to the SeleniumServer class as args.
    * See a full list of options at
    * https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/remote/index.js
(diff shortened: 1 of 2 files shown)


akirakoyasu:
@googlebot I signed it!

lib/config.ts
```diff
@@ -29,6 +29,13 @@ export interface Config {
   seleniumServerJar?: string;
 
   /**
+   * The timeout milliseconds waiting for standalone Selenium Server to start.
```
sjelin: Add a note that it's only for local selenium servers

lib/config.ts
```diff
@@ -29,6 +29,13 @@ export interface Config {
   seleniumServerJar?: string;
 
   /**
+   * The timeout milliseconds waiting for the local standalone Selenium Server to start.
```
sjelin: nit: should be "waiting for ***a*** local Selenium Standalone Server

akirakoyasu:
I'm sorry for my broken English. 😢

lib/config.ts
akirakoyasu: I fixed it.

akirakoyasu:
And squashed all commits.

PR #3794, opened 2016-12-05
chore(scripts): clean up package scripts

cnishina:
- npm start - installs dependencies and starts the testapp
- npm run website - installs dependencies and starts the website
- cleanup testapp package scripts that are not used
- npm run format - shortcut for 'gulp format' for contributors that failed format check test

### DIFF
--- a/testapp/package.json
+++ b/testapp/package.json
@@ -2,10 +2,9 @@
   "name": "testapp",
   "version": "0.0.1",
   "scripts": {
-    "start": "concurrently \"npm run tsc:w\" \"npm run lite\" ",
-    "tsc": "tsc",
-    "tsc:w": "tsc -w",
-    "lite": "lite-server"
+    "postinstall": "npm run tsc",
+    "start": "npm install && node scripts/web-server.js",
+    "tsc": "tsc"
   },
   "dependencies": {
     "@angular/common": "2.2.1",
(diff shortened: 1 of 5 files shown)


testapp/package.json
```diff
@@ -2,10 +2,10 @@
   "name": "testapp",
   "version": "0.0.1",
   "scripts": {
-    "start": "concurrently \"npm run tsc:w\" \"npm run lite\" ",
-    "tsc": "tsc",
-    "tsc:w": "tsc -w",
-    "lite": "lite-server"
+    "install_testapp": "npm install && npm run tsc",
```
sjelin: This should probably be `postinstall`.  Obviously we're installing the testapp, since we're in that folder.  And `postinstall` isn't a problem since we never publish the testapp.

testapp/package.json
```diff
@@ -2,10 +2,10 @@
...
+    "start": "npm run install_testapp && node scripts/web-server.js",
+    "start_detach": "node scripts/web-server.js &",
```
sjelin: it's weird that `start` does an install but `start_detach` doesn't.

sjelin: Also, what is the point of this command?  Presumably people could do `npm start &` on their own?

testapp/package.json
cnishina: Yup, `postinstall` would be better. Changed.

testapp/package.json
cnishina: Removing `install_testapp` from `start` in favor of `npm install`. Removing `start_detach` since its only used in Travis.

PR #3808, opened 2016-12-08
fix(revert): element(locator).then reverted.

cnishina:
- change comments to not be in JSDoc format
- revert change


cnishina:
Where oh where are my Travis tests. Where oh where could they be?

PR #3813, opened 2016-12-10
feat(blockingproxy): Add synchronization with BlockingProxy.

mgiambalvo:
This adds support for BlockingProxy behind the flag --useBlockingProxy. 

If set, the driver providers will start a proxy during their setup phase, passing the selenium address to the proxy and starting a webdriver client that talks to the proxy.

Starting a proxy for each driver provider isn't strictly necessary. However, when we run with multiple capabilities it's easier to handle the logging if each Protractor instance has it's own proxy. 

Known issues:
- Doesn't work with directConnect. You can get the address of chromedriver by mucking around in Selenium internals, but this probably changed for Selenium 3.0 and I doubt it's worth figuring out until we upgrade.
- Doesn't yet work with webDriverProxy (but it's an easy fix)
- Doesn't work with wrapDriver. The new browser instance will need to know somehow that the flag is set, still working on this.


lib/browser.ts
```diff
@@ -188,7 +195,21 @@ export class ProtractorBrowser extends Webdriver {
    *
    * @type {boolean}
    */
-  ignoreSynchronization: boolean;
+  set ignoreSynchronization(value) {
```
juliemr: I think we should take this opportunity to improve the naming here.  I think we had discussed `waitForAngularEnabled(true/false)`

juliemr: I'd also be OK with keeping these changes that keep `ignoreSynchronization` working but adding the new API as the 'recommended way' as well.

lib/browser.ts
```diff
@@ -668,6 +693,12 @@ export class ProtractorBrowser extends Webdriver {
       return 'Protractor.get(' + destination + ') - ' + str;
     };
 
+    if (this.bpClient) {
```
juliemr: Hm, why is this necessary? Because we do executeScript calls, and BpClient isn't smart enough  to know that it doesn't need to wait for those?

lib/config.ts
```diff
@@ -88,6 +88,13 @@ export interface Config {
    */
   webDriverProxy?: string;
 
+  /**
+   * If specified, connect to webdriver through a proxy that manages client-side
+   * synchronization. This option is intended for testing an  experimental feature
+   * and is not meant to be used by end users.
```
juliemr: This is a bit scary - maybe just 

"Blocking Proxy is an experimental feature and may change without notice."?

Also, I think this should be boolean?

lib/driverProviders/driverProvider.ts
```diff
@@ -28,17 +31,28 @@ export class DriverProvider {
     return this.drivers_.slice();  // Create a shallow copy
   }
 
+  getBPUrl() {
+    return `http://localhost:${this.bpRunner.port}`;
```
juliemr: use os.hostname()?

lib/driverProviders/driverProvider.ts
```diff
@@ -89,22 +103,33 @@ export class DriverProvider {
...
   /**
-   * Default setup environment method.
-   * @return a promise
+   * Default setup environment method, common to all driver providers.
    */
   setupEnv(): q.Promise<any> {
-    return q.fcall(function() {});
+    let driverPromise = this.setupDriverEnv();
+    if (this.config_.useBlockingProxy) {
+      // TODO(heathkit): If set, pass the webDriverProxy to BP.
```
juliemr: You should just grab this github username :)

lib/driverProviders/driverProvider.ts
```diff
@@ -89,22 +103,33 @@ export class DriverProvider {
...
+      return q.all([driverPromise, this.bpRunner.start()]);
+    }
+    return driverPromise;
   };
 
   /**
+   * Set up environment specific to a particular driver provider. Overridden
+   * by each driver provider.
+   */
+  protected abstract setupDriverEnv():
```
juliemr: Also update the documentation in driverProviders/README.md

lib/driverProviders/driverProvider.ts
```diff
@@ -89,22 +103,33 @@ export class DriverProvider {
...
+      /**
    * Teardown and destroy the environment and do any associated cleanup.
    * Shuts down the drivers.
    *
    * @public
    * @return {q.promise} A promise which will resolve when the environment
    *     is down.
    */
-  teardownEnv(): q.Promise<q.Promise<webdriver.WebDriver>[]> {
+      teardownEnv(): q.Promise<q.Promise<webdriver.WebDriver>[]> {
```
juliemr: This extra spacing doesn't look right...

package.json
```diff
@@ -72,7 +73,7 @@
   },
   "license": "MIT",
   "engines": {
-    "node": ">=4.2.x"
+    "node": ">=6.9.x"
```
juliemr: Let's avoid this change until 5.x

Out of curiosity, what in this PR depends on it?

spec/blockingProxyConf.js
```diff
@@ -0,0 +1,34 @@
+var env = require('./environment.js');
```
juliemr: Add this to the test.sh script for `npm test`?

spec/blockingProxyConf.js
```diff
@@ -0,0 +1,34 @@
...
+
+// The main suite of Protractor tests.
```
juliemr: Update comment.

spec/blockingProxyConf.js
```diff
@@ -0,0 +1,34 @@
...
+  specs: [
+    'basic/synchronize_spec.js',
+    'ng2/async_spec.js',
+    'ng2/timeout_spec.js',
+    'hybrid/async_spec.js'
+  ],
+
+  blockingProxyUrl: 'http://localhost:8111',
+
+  // Exclude patterns are relative to this directory.
```
juliemr: Unnecessary.

spec/blockingProxyConf.js
```diff
@@ -0,0 +1,34 @@
...
+
+  // Spec patterns are relative to this directory.
+  specs: [
+    'basic/synchronize_spec.js',
+    'ng2/async_spec.js',
+    'ng2/timeout_spec.js',
+    'hybrid/async_spec.js'
+  ],
+
+  blockingProxyUrl: 'http://localhost:8111',
```
juliemr: I think this should be `useBlockingProxy: true` now

lib/browser.ts
mgiambalvo: Good point. I'll add the new API, but I'd like to leave `ignoreSynchronization` in place and announce that it's deprecated.

lib/browser.ts
mgiambalvo: Exactly. We could skip waiting for executeScript calls in general, but I wanted to keep behavior the same.

lib/config.ts
mgiambalvo: Done

lib/driverProviders/driverProvider.ts
mgiambalvo: Done

lib/bpRunner.ts
```diff
@@ -0,0 +1,62 @@
...
+      })
+    })
+  }
+
+  checkSupportedConfig() {
+    if (this.config.directConnect) {
+      throw new Error('BlockingProxy not yet supported with directConnect!');
+    }
+  }
+}
```
mgiambalvo: Done

lib/bpRunner.ts
```diff
@@ -0,0 +1,62 @@
+import {ChildProcess, fork} from 'child_process';
+import * as q from 'q';
+
+import {Config} from './config';
+import {Logger} from './logger';
+
+const BP_PATH = 'node_modules/blocking-proxy/built/lib/bin.js';
```
mgiambalvo: Done

package.json
mgiambalvo: Rebased to the beta branch.

lib/driverProviders/driverProvider.ts
mgiambalvo: That is weird, don't know why clang-format didn't catch this...

lib/driverProviders/driverProvider.ts
mgiambalvo: I think localhost is the right thing to do here, since BP is binding to localhost.

lib/driverProviders/driverProvider.ts
mgiambalvo: Yup

juliemr:
Failing test is the one we decided to kill - so after this is gone and passing, LGTM.

juliemr:
Dunno why CLA-bot is so unhappy today.

Issue #3816, opened 2016-12-12
Incorrect Config definition for cucumberOpts.require


mdasberg:
When configuring the protractor configuration with the  [protractor-cucumber-framework](https://github.com/mattfritz/protractor-cucumber-framework) the cucumberOpts.require is defined incorrectly.

The cucumberOpts definition in config.ts defines the following:

```js
cucumberOpts?: {
    /** Require files before executing the features. */
    require?: string;
}
```
Which is incorrect as the [protractor-cucumber-framework](https://github.com/mattfritz/protractor-cucumber-framework#implementation) defines it as an array of strings as seen in the implementation.

PR #3817, opened 2016-12-12
fix(config): cucumberOpts.require should be an optional Array of strings

mdasberg:
fixes https://github.com/angular/protractor/issues/3816

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -553,7 +553,7 @@ export interface Config {
     /**
      * Require files before executing the features.
      */
-    require?: string[]; 
+    require?: string[] | string;
     /**
      * Only execute the features or scenarios with tags matching @dev.
      * This may be an array of strings to specify multiple tags to include.
@@ -562,7 +562,7 @@ export interface Config {
     /**
      * How to format features (default: progress)
      */
-    format?: string;
+    format?: string[] | string;
     // Other options include `coffee`, `noSnippets`, and `dryRun`
     coffee?: any;
     noSnippets?: any;
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -553,7 +553,7 @@ export interface Config {
     /**
      * Require files before executing the features.
      */
-    require?: Array<string>; 
+    require?: string[]; 
     /**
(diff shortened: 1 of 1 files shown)


lib/config.ts
```diff
@@ -553,7 +553,7 @@ export interface Config {
     /**
      * Require files before executing the features.
      */
-    require?: string;
+    require?: Array<string>;
```
cnishina: Nit: Prefer `string[]` over `Array<string>`

mdasberg: done

wswebcreation:
@mdasberg ,

Can you also check this issue? https://github.com/angular/protractor/issues/3822. It also refers to the`format` in line 565

lib/config.ts
```diff
@@ -553,7 +553,7 @@ export interface Config {
     /**
      * Require files before executing the features.
      */
-    require?: string;
+    require?: string[];
```
calendir: This is still wrong as it can be _either_ a string _or_ an array of strings.

calendir:
Since you're fixing things could you fix a couple more? :)
- `tags` should also be `string[] | string`
- `coffee`, `dryRun`, `noSnippets` are not Cucumber CLI options (at least not anymore) so IMO they should be removed

mdasberg:
Will do that.

dryRun is still a cucumber.js option, but [protractor-cucumber-framework](https://github.com/mattfritz/protractor-cucumber-framework) does not convert dryRun to --dry-run but converts it to --dryRun, which does not match, hence it will not work.

lib/config.ts
```diff
@@ -562,11 +562,9 @@ export interface Config {
     /**
      * How to format features (default: progress)
      */
-    format?: string;
-    // Other options include `coffee`, `noSnippets`, and `dryRun`
-    coffee?: any;
-    noSnippets?: any;
-    dryRun?: any;
+    format?: string[] | string;
+    // Other options include `dryRun`
```
cnishina: instead of having 
```
// Other options include `dryRun`
```

could we have a definition? 
```
/**
 * use dry run for ...
 */
dryRun?: boolean;
```

PR #3821, opened 2016-12-13
Fixes grammar issue in page title

Asheboy:
"end-to-end" is a noun phrase and therefore should be hyphenated.


Asheboy:
I signed it!

juliemr:
Thanks!

PR #3826, opened 2016-12-14
chore(deps): Update Travis Selenium 3.0 and latest browsers

mgiambalvo:
This also drops support for Safari 9, which doesn't use safaridriver.

### DIFF
--- a/spec/basic/elements_spec.js
+++ b/spec/basic/elements_spec.js
@@ -465,7 +465,7 @@ describe('ElementArrayFinder', function() {
     ]);
   });
 
-  fit('should map and resolve multiple promises', function() {
+  it('should map and resolve multiple promises', function() {
     browser.get('index.html#/form');
     var labels = element.all(by.css('#animals ul li')).map(function(elm) {
       return {
--- a/spec/ciFullConf.js
+++ b/spec/ciFullConf.js
@@ -25,7 +25,7 @@ exports.config = {
     'version': '54',
     'selenium-version': '3.0.1',
     'chromedriver-version': '2.26',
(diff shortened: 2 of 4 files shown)


spec/ciFullConf.js
```diff
@@ -22,17 +22,17 @@ exports.config = {
...
     'platform': 'OS X 10.9'
   }, {
     'browserName': 'firefox',
     'tunnel-identifier': process.env.TRAVIS_JOB_NUMBER,
     'build': process.env.TRAVIS_BUILD_NUMBER,
     'name': 'Protractor suite tests',
-    'version': '47',
-    'selenium-version': '2.53.1'
+    'version': '50',
+    'selenium-version': '3.0.1'
```
cnishina: I'm assuming this is the selenium standalone server version. If we upgrade to selenium standalone server 3.0.x, Firefox 47 tests will not work. Can we get our test to work with FF 48+?

Also, since this would test sauce labs, the most that they support for selenium standalone server 2.53.1 and FF 47. See https://wiki.saucelabs.com/display/DOCS/Test+Configuration+Options

If this does work on saucelabs with selenium standalone server 3.0.x and FF 50, I'll need to see how this work locally.

spec/basic/elements_spec.js
```diff
@@ -465,7 +465,7 @@ describe('ElementArrayFinder', function() {
     ]);
   });
 
-  fit('should map and resolve multiple promises', function() {
+  it('should map and resolve multiple promises', function() {
```
cnishina: This looks like something i did. Thanks for fixing this.

PR #3828, opened 2016-12-15
Refactor element explorer to work with selenium-webdriver 3

juliemr:
This implementation now relies mostly on promises explicitly,
so the control flow is only used to add one large task to the queue.
This should pave the way for the eventual removal of the control flow,
as well as getting element explorer to work immediately.

Note that there are several missing items and TODOs at the moment:

- [x] Needs rebasing on top of current beta branch
- [x] This is only for elementExplorer, I haven't yet fixed browser.pause()
- [x] There's some clean-up todos to make the breakpoint more robust
- [x] Need to ensure that the test suite for interactive tests is still working and can still break
- [x] Need to ensure that error cases (e.g. bad syntax in a statement) are handled gracefully

### DIFF
--- a/lib/breakpointhook.js
+++ b/lib/breakpointhook.js
@@ -1,0 +1,10 @@
+module.exports = function() {
+  return true;
+};
+
+/**
+ * The reason this file exists is so that we can set a breakpoint via
+ * script name, and then control when that breakpoint is set in
+ * our library code by importing and calling this function. The
+ * breakpoint will always be on line 2.
+ */
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -972,37 +972,46 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
   }
 
   /**
-   * Beta (unstable) enterRepl function for entering the repl loop from
-   * any point in the control flow. Use browser.enterRepl() in your test.
+   * See browser.explore().
+   */
+  enterRepl(opt_debugPort?: number) {
+    return this.explore(opt_debugPort);
+  }
+
+  /**
+   * Beta (unstable) explore function for entering the repl loop from
+   * any point in the control flow. Use browser.explore() in your test.
    * Does not require changes to the command line (no need to add 'debug').
    * Note, if you are wrapping your own instance of Protractor, you must
    * expose globals 'browser' and 'protractor' for pause to work.
    *
    * @example
    * element(by.id('foo')).click();
-   * browser.enterRepl();
+   * browser.explore();
    * // Execution will stop before the next click action.
    * element(by.id('bar')).click();
    *
    * @param {number=} opt_debugPort Optional port to use for the debugging
    * process
    */
-  enterRepl(opt_debugPort?: number) {
+  explore(opt_debugPort?: number) {
     let debuggerClientPath = __dirname + '/debugger/clients/explorer.js';
-    let onStartFn = () => {
+    let onStartFn = (firstTime: boolean) => {
       logger.info();
-      logger.info('------- Element Explorer -------');
-      logger.info(
-          'Starting WebDriver debugger in a child process. Element ' +
-          'Explorer is still beta, please report issues at ' +
-          'github.com/angular/protractor');
-      logger.info();
-      logger.info('Type <tab> to see a list of locator strategies.');
-      logger.info('Use the `list` helper function to find elements by strategy:');
-      logger.info('  e.g., list(by.binding(\'\')) gets all bindings.');
-      logger.info();
+      if (firstTime) {
+        logger.info('------- Element Explorer -------');
+        logger.info(
+            'Starting WebDriver debugger in a child process. Element ' +
+            'Explorer is still beta, please report issues at ' +
+            'github.com/angular/protractor');
+        logger.info();
+        logger.info('Type <tab> to see a list of locator strategies.');
+        logger.info('Use the `list` helper function to find elements by strategy:');
+        logger.info('  e.g., list(by.binding(\'\')) gets all bindings.');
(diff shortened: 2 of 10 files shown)


juliemr:
We can greatly simplify our code by making a change to separate the 'step by execute' and 'enter repl' debuggers more clearly. This would mean that if you do `browser.pause` you can no longer enter a repl loop - but you can do that by _instead_ placing `browser.enterRepl` in your code. 

I'll go ahead and implement this to show how much easier it is to reason about. I've seen many users not know they can get into a repl from `browser.pause` anyway, so I think we can use this as an opportunity to improve clarity and docs.

juliemr:
One more item to fix - the tab completion for the explorer is currently broken.

juliemr:
This is now ready for review! 

I'm also considering aliasing `browser.enterRepl()` as `browser.explore()`. Thoughts?

lib/breakpointhook.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function() {
```
mgiambalvo: It'd be good to have a comment here explaining why this is awesome. Oh, and maybe it should go under debugger/

lib/debugger.ts
```diff
@@ -27,18 +28,21 @@ export class DebugHelper {
...
    *  2) Enter process into debugger mode. (i.e. process._debugProcess).
    *  3) Invoke the debugger client specified by debuggerClientPath.
    *
    * @param {string} debuggerClientPath Absolute path of debugger client to use.
+   * @param {boolean} blockUntilExit Whether to block until process exit or resume immediately.
    * @param {Function} onStartFn Function to call when the debugger starts. The
    *     function takes a single parameter, which represents whether this is the
    *     first time that the debugger is called.
-   * @param {number=} opt_debugPort Optional port to use for the debugging
+   * @param {number=} opt_debugPort Optional port to u`se for the debugging
```
mgiambalvo: tpyo

lib/browser.ts
```diff
@@ -910,7 +910,7 @@ export class ProtractorBrowser extends Webdriver {
       logger.info('  e.g., list(by.binding(\'\')) gets all bindings.');
       logger.info();
     };
-    this.debugHelper.init(debuggerClientPath, onStartFn, opt_debugPort);
+    this.debugHelper.init(debuggerClientPath, true, onStartFn, opt_debugPort);
```
mgiambalvo: Up to you, but it'd be nice if this were an enum, or at least had a comment explaining what this flag does.

lib/debugger.ts
```diff
@@ -95,34 +102,30 @@ export class DebugHelper {
...
+    // into the control flow, via debugger 'evaluate' protocol.
+    // In order to achieve this, we maintain a task at the top of the control
     // flow, so that we can insert frames into it.
     // To be able to simulate callback/asynchronous code, we poll this object
-    // for a result at every run of DeferredExecutor.execute.
-    let browserUnderDebug = this.browserUnderDebug_;
+    // whenever `breakpointHook` is called.
     this.dbgCodeExecutor = {
-      execPromise_: pausePromise,     // Promise pointing to current stage of flow.
+      execPromise_: null,             // Promise pointing to currently executing command.
```
mgiambalvo: Why does this need to be null instead of undefined?

lib/debugger/clients/wddebugger.js
```diff
@@ -75,28 +38,25 @@ WdDebugger.prototype.stepEval_ = function(cmd, context, filename, callback) {
  */
 WdDebugger.prototype.initRepl_ = function() {
   var self = this;
-  this.cmdRepl = new CommandRepl(this.client);
   this.dbgRepl = new DebuggerRepl(this.client);
-  this.currentRepl = this.dbgRepl;
 
   // We want the prompt to show up only after the controlflow text prints.
   this.dbgRepl.printControlFlow_(function() {
     // Backward compatibility: node version 0.8.14 has a number of built in
```
mgiambalvo: Can we just get rid of this?

lib/debugger/modes/debuggerRepl.js
```diff
@@ -76,19 +84,22 @@ DebuggerRepl.prototype.stepEval = function(cmd, callback) {
  * @param {function} callback
  */
 DebuggerRepl.prototype.complete = function(line, callback) {
+  console.log('calling debuggerrepl complete');
```
mgiambalvo: Should these be here?

lib/breakpointhook.js
juliemr: Done adding a comment. I don't think this belongs under debugger/ because all of the code in there runs in the separate debugger process, whereas this is part of the same process as browser + friends.

lib/debugger.ts
juliemr: fixed

lib/debugger.ts
juliemr: No reason, changed.

lib/browser.ts
juliemr: Renamed into two different functions on the theory that the best documentation is good naming.

lib/debugger/clients/wddebugger.js
juliemr: Yes, thanks for catching that.

lib/debugger/modes/debuggerRepl.js
juliemr: Nope, gone.

juliemr:
Comments addressed, and rebased on top of beta. PTAL!

juliemr:
Rebase successful! Just waiting for CI to be green now.

PR #3833, opened 2016-12-15
chore(tslint): add tslint to gulpfile

cnishina:
- add gulp task tslint and lint
- update `**/*.ts` with tslint fixes
- update circle.yml to run `gulp lint`
- update `gulp pretest` to run both jshint and tslint

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -12,7 +12,7 @@ dependencies:
   cache_directories:
     - testapp/node_modules
   post:
-    - ./node_modules/.bin/gulp format:enforce
+    - ./node_modules/.bin/gulp lint
     - ./node_modules/.bin/webdriver-manager update
     - ./node_modules/.bin/webdriver-manager start:
           background: true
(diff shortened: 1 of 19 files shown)


mgiambalvo:
As part of this, could you put the tslint check on Circle? And please move the format check along with tslint down to the test phase, so they won't block testing if they fail. Thanks!

circle.yml
```diff
@@ -12,7 +12,7 @@ dependencies:
   cache_directories:
     - testapp/node_modules
   post:
-    - ./node_modules/.bin/gulp format:enforce
+    - ./node_modules/.bin/gulp lint
```
mgiambalvo: I think we still need the format check too, right? Also, let's move this down to before `npm test` so lint failures don't block test runs.

cnishina: I've added the format enforce into the lint similar to blocking-proxy. I also would rather have the test fail immediately if lint / format has errors.

Issue #3838, opened 2016-12-18
Add second() command for selecting the second element


Marketionist:
It would be great to have readable second() command for selecting the second element. Some additional evidences:
- http://stackoverflow.com/questions/33279813/locating-second-third-fourth-eigth-element-in-protractor-using-css-locator
- http://stackoverflow.com/questions/37529654/how-to-select-second-objects-second-element-with-same-classname-in-protractor

This not a breaking change. It will extend the existing functionality and make elements interactions more readable.

juliemr:
As I mentioned in the PR, I think this adds complexity, and `get(1)` is simpler and more generic. Thanks for looking into this!

Marketionist:
As I replied in https://github.com/angular/protractor/pull/3839 - it looks like 90% most often used cases are: `first()`, `get(1)` and `get(2)`. So adding `second()` and `third()` shortcuts looked like a reasonable addition that will give more readability for most of the cases. Anyway, it is up to you to decide if this should become a part of protractor syntax or not. Thanks for reviewing!

Marketionist:
**Update**: created a separate package - https://github.com/Marketionist/protractor-numerator - that can be used with protractor to select second, third, etc. elements. Would be great to hear what you think about it @juliemr

PR #3839, opened 2016-12-18
feat(element): add second() command for selecting the second element

Marketionist:
**feat(element): add second() command for selecting the second element**

### DIFF
--- a/lib/driverProviders/direct.ts
+++ b/lib/driverProviders/direct.ts
@@ -42,7 +42,8 @@ export class Direct extends DriverProvider {
         break;
       default:
         throw new BrowserError(
-            logger, 'browserName ' + this.config_.capabilities.browserName +
+            logger,
+            'browserName ' + this.config_.capabilities.browserName +
                 ' is not supported with directConnect.');
     }
     return q.fcall(function() {});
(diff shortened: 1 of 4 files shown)


Marketionist:
Fixes https://github.com/angular/protractor/issues/3838

Marketionist:
Circle CI failed with "Error: does not exist in logs: Cannot run in debug mode with multiCapabilities, count > 1, or sharding" in protractor/scripts/exitCodes.js:13 - which is not related to current PR

Marketionist:
@sjelin @cnishina please review

juliemr:
I think that this adds complexity without adding any real value. `.get(1)` is just as easy to understand - and where do we draw the line here? `third()`? `sixteenth()`? I think first and last have real value, I'd prefer to not add anything in between.

Marketionist:
@juliemr as far as I've seen (including questions from users of stackoverflow), most often used cases are: `first()`, `second()` and `third()`. This three shortcuts looked like a reasonable addition that will give more readability for 90% of cases.

But, anyway, it's up to you to decide if it should become a part of protractor syntax.

Issue #3836, opened 2016-12-17
Automatically detect hybrid apps most of the time


sjelin:
Before the `ng12Hybrid` flag, our wait logic was as follows:

```js
if (window.getAngularTestability) {
  // We are in an ng2 app and should use ng2 testability
} else {
  // We are in an ng1 app and should use ng1 testability
}
```

This causes problems because *sometimes*, but not consistently, the ngUpgrade API would publish both testability APIs, causing us to use the ng2 testability API, but really we wanted the ng1 testability API (which is what you use in hybrid apps).  So we introduced the `ng12Hybrid` flag to resolve this ambiguity.  But was it really ambiguous?  Instead we should have just reversed the original logic:
 
```js
if (window.angular && !(window.angular.version && window.angular.version.major > 1)) {
  // We are in either an ng1 app or an ngUpgrade app and should use ng1 testability
} else {
  // We are in an ng2 app and should use ng2 testability
}
```

I propose we do the following:

* In `master`, if `ng12Hybrid` is not set, used the new check (which makes `ng12Hybrid` needless)
* In `beta`, rebase the above change in and then remove the `ng12Hybrid` flag entierly

mgiambalvo:
I like this, and it's a good time to do this kind of breaking change. It would also be nice to just have a single waitForAngular function that detects the Angular version and does the right thing.

sjelin:
@mgiambalvo discussed this a bit offline, and this works fine for `waitForAngular`, but there's one snag during the bootstrap process: If you see the ng2 testability API, but not the ng1 API, it is possible that ng1 simply hasn't loaded yet.  This is extremely unlikely, and would require the app to be loading ng1 asynchronously *after* `platformBrowserDynamic()` was called, as opposed to just having a `<script>` tag at the top of the page with ng1.  So this is the new proposed solution:

1) Use the new wait logic instead of the `ng12Hybrid` flag everywhere except `browser.get`'s [`testForAngular`](https://github.com/angular/protractor/blob/master/lib/clientsidescripts.js#L596).
2) Change `testForAngular`'s [`check()`](https://github.com/angular/protractor/blob/master/lib/clientsidescripts.js#L602) logic to the following:
   ```js
   var definitelyNg1 = !!ng12Hybrid;
   var definitelyNg2 = false;
   function check(n) {
     // Figure out which version of angular we're waiting on
     if(!definitelyNg1 && !definitelyNg2) {
[...]

PR #3847, opened 2016-12-19
feat(ngUpgrade): Auto detect ngUpgrade apps and make the ng12Hybrid f…

sjelin:
…lag unnecessary for most users

Closes https://github.com/angular/protractor/issues/3836

### DIFF
--- a/lib/clientsidescripts.js
+++ b/lib/clientsidescripts.js
@@ -44,33 +44,38 @@ function wrapWithHelpers(fun) {
  * Asynchronous.
  *
  * @param {string} rootSelector The selector housing an ng-app
- * @param {boolean} ng12Hybrid Flag set if app is a hybrid of angular 1 and 2
  * @param {function(string)} callback callback. If a failure occurs, it will
  *     be passed as a parameter.
  */
-functions.waitForAngular = function(rootSelector, ng12Hybrid, callback) {
+functions.waitForAngular = function(rootSelector, callback) {
   var el = document.querySelector(rootSelector);
 
   try {
-    if (!ng12Hybrid && window.getAngularTestability) {
+    if (window.angular && !(window.angular.version &&
+        window.angular.version.major > 1)) {
+      if (angular.getTestability) {
+        angular.getTestability(el).whenStable(callback);
+      } else if (angular.element(el).injector()) {
+        angular.element(el).injector().get('$browser').
+            notifyWhenNoOutstandingRequests(callback);
+      } else {
+        throw new Error('root element (' + rootSelector + ') has no injector.' +
+           ' this may mean it is not inside ng-app.');
+      }
+    } else if (window.getAngularTestability) {
       window.getAngularTestability(el).whenStable(callback);
-      return;
-    }
-    if (!window.angular) {
+    } else if (!window.angular) {
       throw new Error('window.angular is undefined.  This could be either ' +
           'because this is a non-angular page or because your test involves ' +
           'client-side navigation, which can interfere with Protractor\'s ' +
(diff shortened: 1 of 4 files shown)


lib/clientsidescripts.js
```diff
@@ -44,33 +44,37 @@ function wrapWithHelpers(fun) {
...
-    if (!window.angular) {
+    } else if (!window.angular) {
       throw new Error('window.angular is undefined.  This could be either ' +
           'because this is a non-angular page or because your test involves ' +
           'client-side navigation, which can interfere with Protractor\'s ' +
           'bootstrapping.  See http://git.io/v4gXM for details');
-    }
-    if (angular.getTestability) {
-      angular.getTestability(el).whenStable(callback);
+    } else if (window.angular.version == 2) {
```
juliemr: Since Angular versions are going to keep advancing now, we should change this to:

```js
else if (window.angular.version.major >= 2) {
  throw new Error('You appear to be using angular, but window.getAngularTestability was ' + 
     'never set. This may be due to bad obfuscation.')
```

lib/clientsidescripts.js
```diff
@@ -599,15 +603,36 @@ functions.testForAngular = function(attempts, ng12Hybrid, asyncCallback) {
...
   var check = function(n) {
     try {
-      if (!ng12Hybrid && window.getAllAngularTestabilities) {
-        callback({ver: 2});
-      } else if (window.angular && window.angular.resumeBootstrap) {
-        callback({ver: 1});
-      } else if (n < 1) {
-        if (window.angular) {
+      // Figure out which version of angular we're waiting on
+      if(!definitelyNg1 && !definitelyNg2) {
```
juliemr: nit: space after `if`

lib/clientsidescripts.js
```diff
@@ -599,15 +603,36 @@ functions.testForAngular = function(attempts, ng12Hybrid, asyncCallback) {
       asyncCallback(args);
     }, 0);
   };
+  var definitelyNg1 = !!ng12Hybrid;
+  var definitelyNg2 = false;
```
juliemr: We should probably rename this as well, given that angular version thing. What about just `definitelyNg`

lib/clientsidescripts.js
```diff
@@ -599,15 +603,36 @@ functions.testForAngular = function(attempts, ng12Hybrid, asyncCallback) {
...
   var check = function(n) {
     try {
-      if (!ng12Hybrid && window.getAllAngularTestabilities) {
-        callback({ver: 2});
-      } else if (window.angular && window.angular.resumeBootstrap) {
-        callback({ver: 1});
-      } else if (n < 1) {
-        if (window.angular) {
+      // Figure out which version of angular we're waiting on
```
juliemr: These comments should be `/** */` style because some versions of webdriver will glomp lines together and `//` comments break stuff.

lib/clientsidescripts.js
```diff
@@ -599,15 +603,36 @@ functions.testForAngular = function(attempts, ng12Hybrid, asyncCallback) {
...
+        }
+      }
+      // See if our version of angular is ready
+      if (definitelyNg1) {
+        if (window.angular && window.angular.resumeBootstrap) {
+          return callback({ver: 1});
+        }
+      } else if (definitelyNg2) {
+        if (true /** ng2 has no resumeBootstrap() **/) {
+          return callback({ver: 2});
```
juliemr: Leaving this as `ver: 2` is fine for now but we should make this more future-proof later.

sjelin:
@juliemr All comments addressed!

juliemr:
LGTM

PR #3849, opened 2016-12-20
feat(rootEl): auto-detect the root element better

### LINKED ISSUE #1742: Improve rootElement default
At the moment, rootElement defaults to `body`, which is not particularly smart. It can be set manually, but we can improve the experience here.

sjelin:
Closes #1742

### DIFF
--- a/lib/clientsidescripts.js
+++ b/lib/clientsidescripts.js
@@ -16,6 +16,13 @@
 /* global angular */
 var functions = {};
 
+///////////////////////////////////////////////////////
+////                                               ////
+////                    HELPERS                    ////
+////                                               ////
+///////////////////////////////////////////////////////
+
+
 /* Wraps a function up into a string with its helper functions so that it can
  * call those helper functions client side
  *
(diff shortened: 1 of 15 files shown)


sjelin:
I *might* have a much better idea than this, pending talking with @juliemr later today, so don't both reviewing right now

sjelin:
As per talking to @juliemr, this will will not be part of a larger change after all.  Please review! :smile:

lib/clientsidescripts.js
```diff
@@ -36,6 +43,83 @@ function wrapWithHelpers(fun) {
       '  return (' + fun.toString() + ').apply(this, arguments);');
 }
 
+/* Tests if an ngRepeat matches a repeater
+ *
+ * @param {string} ngRepeat The ngRepeat to test
+ * @param {string} repeater The repeater to test against
+ * @param {boolean} exact If the ngRepeat needs to be more or less exactly the
```
juliemr: Why "more or less"? Just 'exactly'

lib/clientsidescripts.js
```diff
@@ -36,6 +43,83 @@ function wrapWithHelpers(fun) {
...
+/* Tries to find $$testability and possibly $injector for an ng1 app
+ *
+ * By default, doesn't care about $injector if it finds $$testability.  However,
+ * these priorities can be reversed.
+ *
+ * @param {string=} selector The selector for the element with the injector.  If
+ *   falsy, tries a variety of methods to find an injector
+ * @param {boolean=} injectorPlease Prioritize finding an injector
+ * @return {$$testability?: Testability, $injector?: Injector} Returns what
+ *   the app hooks it finds
```
juliemr: This description seems to have gotten a bit mangled.

lib/clientsidescripts.js
```diff
@@ -36,6 +43,83 @@ function wrapWithHelpers(fun) {
...
+    try {
+      if (!injectorPlease && angular.getTestability) {
+        var $$testability = angular.getTestability(el);
+        if ($$testability) {
+          return {$$testability: $$testability};
+        }
+      } else {
+        var $injector = angular.element(el).injector();
+        if ($injector) {
+          return {$injector: angular.element(el).injector()};
```
juliemr: just `return {$injector: $injector}` should work here, yeah?

lib/clientsidescripts.js
sjelin: Well, it removes ` track by ` and ` as ` and everything after `|` or `=`.  Is it still *exactly* the repeater?  I don't actually understand ng1 well enough to know which parts of the `ng-repeat` count.

lib/clientsidescripts.js
```diff
@@ -44,66 +128,60 @@ function wrapWithHelpers(fun) {
...
  */
-functions.waitForAngular = function(rootSelector, ng12Hybrid, callback) {
-  var el = document.querySelector(rootSelector);
-
+functions.waitForAngular = function(rootSelector, callback) {
   try {
-    if (!ng12Hybrid && window.getAngularTestability) {
+    if (window.angular && !(window.angular.version &&
+          window.angular.version.major > 1)) {
+      // ng1
```
juliemr: use `/** */` for comments in here.

lib/clientsidescripts.js
```diff
@@ -44,66 +128,60 @@ function wrapWithHelpers(fun) {
...
+      });
+    } else if (!window.angular) {
       throw new Error('window.angular is undefined.  This could be either ' +
           'because this is a non-angular page or because your test involves ' +
           'client-side navigation, which can interfere with Protractor\'s ' +
           'bootstrapping.  See http://git.io/v4gXM for details');
-    }
-    if (angular.getTestability) {
-      angular.getTestability(el).whenStable(callback);
+    } else if (window.angular.version == 2) {
```
juliemr: As in other PR, let's change this to be future proof for angular versions > 2

lib/config.ts
```diff
@@ -356,8 +356,18 @@ export interface Config {
   baseUrl?: string;
 
   /**
-   * CSS Selector for the element housing the angular app - this defaults to
-   * 'body', but is necessary if ng-app is on a descendant of <body>.
+   * A CSS selector for the DOM element you want Protractor to hook into your
+   * app at.
```
juliemr: Maybe

"A CSS Selector for a DOM element within your Angular application. Protractor will attempt to automatically find your application, but It is necessary to set rootElement in certain cases.

In Angular 1, Protractor will use bootstrapping and then search for `body` or `ng-app` elements (details here: https://git.io/v1b2r).

In later versions, Protractor will try to hook into all angular apps on the page. Use rootElement to limit the scope of which apps Protractor waits for and searches within.

lib/clientsidescripts.js
juliemr: What about `If the ngRepeat expression should exactly match`.... it's not 100% precise because of `track by` but I think it gets the idea across more clearly than "more or less"

lib/config.ts
sjelin: "Protractor will use bootstrapping" is a little unclear but that's definitely an improvement

lib/clientsidescripts.js
```diff
@@ -36,6 +43,84 @@ function wrapWithHelpers(fun) {
...
+  } else if (window.__TESTABILITY__NG1_APP_ROOT_INJECTOR__) {
+    var $injector = window.__TESTABILITY__NG1_APP_ROOT_INJECTOR__;
+    var $$testability = null;
+    try {
+      $$testability = $injector.get('$$testability');
+    } catch (e) {}
+    return {$injector: $injector, $$testability: $$testability};
+  } else {
+    return tryEl(document.body) ||
+        trySelector('[ng-app]') || trySelector('[ng:app]') ||
```
sjelin: @juliemr should this include `ng_appp` too?

sjelin:
@juliemr all comments addressed

lib/clientsidescripts.js
juliemr: I don't think anyone uses that... I'd leave it out. We can always add it later if there's a clamor.

juliemr:
LGTM

PR #3852, opened 2016-12-20
docs(element): Fix docs for #896 (clear() and sendKeys())

adrienverge:
Commit 3c0e727 (PR #896) allowed chaining of actions (i.e. `element(By.x).clear().sendKeys('abc)`). Since then, the return type of `clear()` and `sendKeys()` is not longer a `Promise<void>`, but the WebElement itself.

This commit fixes the docs.

It will allow typings for selenium-webdriver to be corrected. Until now, TypeScript complains with:

    error TS2339: Property 'sendKeys' does not exist on type 'Promise<void>'.

### DIFF
--- a/lib/selenium-webdriver/webdriver.js
+++ b/lib/selenium-webdriver/webdriver.js
@@ -479,8 +479,7 @@ webdriver.WebElement.prototype.click = function() {};
  *
  * @param {...(string|!webdriver.promise.Promise<string>)} var_args The sequence
  *     of keys to type. All arguments will be joined into a single sequence.
- * @returns {!webdriver.promise.Promise.<void>} A promise that will be resolved
- *     when all keys have been typed.
+ * @returns {!webdriver.WebElement} The element itself.
  */
 webdriver.WebElement.prototype.sendKeys = function(var_args) {};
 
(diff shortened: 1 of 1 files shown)


adrienverge:
I signed it!

cnishina:
Closing this PR since the documentation return types are promises and are also in their JSDoc.



- [WebElement sendKeys](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/webdriver.js#L2060). JSDoc is slightly different in the current version of selenium-webdriver Protractor uses:
```
   * @return {!promise.Promise<void>} A promise that will be resolved
   *     when all keys have been typed.
```
- [WebElement clear](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/webdriver.js#L2255). JSDoc is slightly different in the current version of selenium-webdriver Protractor uses:
```
   * @return {!promise.Promise<void>} A promise that will be resolved
   *     when the element has been cleared.
```

adrienverge:
@cnishina I don't understand. How can return types be promises and be chainable like this in the same time?

[spec/basic/elements_spec.js](https://github.com/angular/protractor/blob/83694f5e/spec/basic/elements_spec.js#L37):
```js
usernameInput.clear().sendKeys('Jane');
```
[docs/style-guide.md](https://github.com/angular/protractor/blob/83694f5e/docs/style-guide.md#L98):
```js
userProperties.name.clear().sendKeys('Teddy C');
```

(These examples come from protractor official tests and docs.)

It seems that since commit 3c0e727, Selenium and Protractor return types are not the same.

adrienverge:
@cnishina?

PR #3855, opened 2016-12-21
feat(config): allow option for suites to use different capabilities

### LINKED ISSUE #3027: Allow suites to be declared at a capability level
Like specs, it'd be nice to be able to configure suites at a capability level so that suites can be run against specific browser sessions as opposed to its current behavior where suites are declared globally and forced to run across all browser sessions.

I don't imagine this will break anything

sjelin:
Closes #3027

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -1,6 +1,10 @@
+import * as q from 'q';
+
 import {PluginConfig} from './plugins';
 
-export interface Config {
+export interface Config extends SpecsSettings {
+  [key: string]: any;
+
   // ---------------------------------------------------------------------------
   // ----- How to connect to Browser Drivers -----------------------------------
   // ---------------------------------------------------------------------------
@@ -170,50 +174,11 @@ export interface Config {
   // ----- What tests to run ---------------------------------------------------
   // ---------------------------------------------------------------------------
 
-  /**
-   * Use default globals: 'protractor', 'browser', '$', '$$', 'element', 'by'.
-   * These also exist as properties of the protractor namespace:
-   * 'protractor.browser', 'protractor.$', 'protractor.$$',
-   * 'protractor.element', 'protractor.by', and 'protractor.By'.
-   *
-   * When no globals is set to true, the only available global variable will be
-   * 'protractor'.
-   */
-  noGlobals?: boolean;
-
-  /**
-   * Required. Spec patterns are relative to the location of this config.
-   *
-   * Example:
-   * specs: [
-   *   'spec/*_spec.js'
-   * ]
-   */
-  specs?: Array<string>;
-
-  /**
-   * Patterns to exclude specs.
-   */
-  exclude?: Array<string>|string;
-
-  /**
-   * Alternatively, suites may be used. When run without a command line
-   * parameter, all suites will run. If run with --suite=smoke or
-   * --suite=smoke,full only the patterns matched by the specified suites will
-   * run.
-   *
-   * Example:
-   * suites: {
-   *   smoke: 'spec/smoketests/*.js',
-   *   full: 'spec/*.js'
-   * }
-   */
-  suites?: any;
-  /**
-   * If you would like protractor to use a specific suite by default instead of
-   * all suites, you can put that in the config file as well.
-   */
-  suite?: string;
+  // See the `SpecsSettings` interface below
+  // specs?: string[];
+  // exclude?: string[];
+  // suites?: {[suiteName: string]: string | string[] | Suite};
+  // suite?: string;
 
   // ---------------------------------------------------------------------------
   // ----- How to set up browsers ----------------------------------------------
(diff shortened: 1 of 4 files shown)


lib/config.ts
```diff
@@ -319,7 +269,7 @@ export interface Config {
    * `beforeLaunch` is run, and before any driver is set up. If this is
    * specified, both capabilities and multiCapabilities will be ignored.
    */
-  getMultiCapabilities?: any;
+  getMultiCapabilities?: () => q.IWhenable<Array<Capabilities>>;
```
juliemr: Could the type here be `Promise<Array<Capabilities>>`? `q.IWhenable` is gross to encounter while reading documentation.

lib/configParser.ts
```diff
@@ -100,19 +103,41 @@ export class ConfigParser {
...
   }
 
   /**
+   * Run alterCapabilities() if it is defined for the current suite.
+   *
+   * @param {Config} config The config file
+   * @return {Capabilities} The capabilities for the suite
+   */
+  static getSuiteCapabilities(config: Config): Array<Capabilities> {
+    let suite: any = config.suite && config.suites && config.suites[config.suite];
```
juliemr: This doesn't account for input like `--suite=suite1,suite2`. I'm not even sure what the correct behavior would be in that situation.

lib/config.ts
sjelin: It may or may not be a promise, and I'm not sure but I think `Promise` is a class not an interface, but I'll check and figure something better out

lib/configParser.ts
sjelin: I guess you'd run them both, one after the other

sjelin: Except that it's a union, so maybe not.  Hmm...

sjelin:
@juliemr how would the feature as you describe it work?  A capability will only be used for tests matching that suite?  Or that capability will used if that suite is specified?  If the latter and `config.specs.length == 0`, do we use all capabilities then?  Also, more often I think users will want the capability being used *except* against certain suites.  And you could imagine some suites wanting extra capabilities too.

The idea was that this would be a cleaner experience, but the point you make in your `configParser` comment undermines that.

juliemr:
My current idea is that `suite` is conceptually just `array<spec>`, so a `suite` inside a capability would expand to the same meaning as a list of specs for that capability.

Precisely, we have tests a, b, c, etc.

```
specs: 'a',
suites: {
  'foo': 'x',
  'bar': 'y'
},
multiCapabilities: [{
  browserName: 'chrome',
  specs: 'b'
  // chrome runs 'a', 'b'
}, {
  browserName: 'firefox',
  // firefox runs 'a'
}, {
  browserName: 'ie',
  suite: 'foo'
 // ie runs 'a', 'x'. *** It's debatable whether ie should run 'a', I'm not sure.
}, {
  browserName: 'edge',
  suite: 'foo,bar'
  // edge runs 'a', 'x', 'y'
```

juliemr:
Responding to my *** above, I think that I would amend that to ie and edge NOT running spec a, since that's how it works at the base level (if you specify config.specs and config.suite, config.suite wins and whatever was in specs doesn't run)

sjelin:
Alright, that seems reasonable

sjelin:
I haven't added tests yet or totally cleaned up this PR because I think we should talk about the logic it implements.

After re-reading the issue it's clear that he meant capabilities should be able to use the same suites API that the config does.  i.e.:

```js
  capabilities: {
    browserName: 'chrome',
    suites: {
      smoke: 'chrome_smoke/*_spec.js',
      full: 'chrome_*/*_spec.js',
  }
```

This change is an attempt to do that.  I know it's rather more complicated than you might expect, but here are some use cases I was considering:

* The config does not use suites directly but a `capabilities` does
* A `capabilities` uses suites but the config does not
* The user has existing configs which use `capabilities.specs` to add tests for that `capabilities` to all suites.
* A user wants to add some tests to be run against all suites for a `capabilities`, and other tests to only be run when certain suites are used

The resulting logic is basically:

* For the `config` tests:
  * `config.specs` becomes the default suite, who's name is the empty string (i.e. `''`).
  * If `config.suite` is `undefined`, and `config.specs` is a non-empty array, the default suite will be run.
  * If `config.suite` is not `undefined`, and `config.specs` is `[]` or `undefined`, all suites will be run. This logic is to avoid a breaking change.
* For the `capabilities` tests:
  * `capabilities.specs` becomes the global suite, who's name is `'*'`.  The global suite is always run, in addition to what suites the user specifies.
  * If `config.suite` is defined, that suite is also used for each `capabilities`.  This overrides `capabilities.suite` (so that the user can easily pick a suite from the command line). 
  * If `config.suites` is `undefined` and `capabilities.suite` is not, `capabilities.suite` is used as the suite *for this `capabilities` only*.
  * If both `config.suites` and `capabilities.suite` are `undefined`, and `capabilities.specs` is a non-empty array, only the global suite is run for this `capabilities`.  This is true even if `config` defaulted to running all suites (i.e. defaults from `config` do not carry on to each `capabilities`).
  * If both `config.suites` and `capabilities.suite` are `undefined`, and `capabilities.specs` is `[]` or `undefined`, all suites will be run *for this `capabilities` only*.

I am definitely open to changing this logic, or refactoring how it is implemented.

sjelin:
After talking to @juliemr offline, we decided that since there has been no follow up to the original issue, the solution isn't worth the complexity, so I'm closing this PR.  Should I delete the `iss3027` branch from `sjelin/protractor`, the contents of this change should still be visible at https://github.com/angular/protractor/pull/3855/commits/2c633884b0082a57b145edd9745bb213b6652a34.  In case they aren't, I have also created a gist with the result of `git diff HEAD^`: https://gist.github.com/sjelin/f86edeed0775257f593f0b5c1c98a6f5

Issue #1940, opened 2015-03-18
Support Appium's additional WebDriverJS commands


juliemr:

LeninChristopher:
How to select date which is in date picker using protractor

chrisfelix82:
Can't wait for this to be ready!

idaneliz:
Is this solution ready for usage?

LeninChristopher:
May we know the release date of protractor support for hybrid apps.

juliemr:
See https://github.com/angular/protractor/pull/3326/files

yfebles:
@juliemr  i was facing the same issue when trying to test an hybrid ionic app on a webview on android.  

It seems to me an  autoWebView : true  error using it  inside protractor. First i cant even start the tests cause the request were  been made using asycn methods of the webdriver api and on webviews should use implicitlyWait. 

I change the runner.js file on the createBrowser method to be able to load the tests and execute them  but of course inside the tests any call that is made by asycn methods has the same error from server. 

In the appium [forum ](https://discuss.appium.io) there was other situations like this one. I will try to fix the issue to use protractor for the e2e test, so any help on docs is welcome.

gkohen:
@yfebles and @juliemr  , could you share a configuration for testing a hybrid app with Protractor and Appium?
The one article in the Protractor repository about the matter is for mobile browsers, not hybrid app.
I do manage to get to the step up to the point I'm getting:

> [debug] [JSONWP Proxy] Got response with status 200: {"sessionId":"579045442705f7f0e10147786203060d","status":13,"value":{"message":"unknown error: unable to discover open pages\n  (Driver info: chromedriver=2.25.426935 (820a95b0b81d33e42712f9198c215f703412e1a1),platform=Mac OS X 10.11.6 x86_64)"}}

I've been going at it for days now and ready to back out of testing my Angular Hybrid app with Protractor which is a pretty big disadvantage considering Protractor semantic knowledge of an NG app.

sjelin:
We don't have a setup for hybrid apps yet, just mobile browsers.  Hybrid apps will probably require quite a bit of work on top of that.  I wouldn't hold your breath unless someone already has a clean solution

maciejkamela:
Is it possible to use Protractor + Appium for native android/IOS apps?

asifetisalat:
@juliemr - Please let me know if it is possible to test Hybrid mobile app (not a web app) using Protractor + Appium.

ansreenath:
[...]

PR #3326, opened 2016-07-01
feat(mobile): add extended wd commands for appium

sjelin:

PR #3860, opened 2016-12-21
feat(mobile): add extended wd commands for appium

sjelin:
Had to make some minor changes to the website to handle longer inheritance
chains

Closes https://github.com/angular/protractor/issues/1940


lib/webdriver-js-extender/index.js
```diff
@@ -0,0 +1,63 @@
...
+// //  webdriver_extensions.ExtendedWebDriver
+// //
+// /////////////////////////////////////////////////////////////////////////////
+/**
+ * Protractor's `browser` object is a wrapper for an instance of
+ * `ExtendedWebDriver`, provided by `webdriver-js-extender`, which itself is
+ * just an instance of `selenium-webdriver`'s WebDriver with some extra methods
+ * added in. The `browser` object inherits all of WebDriver's and
+ * ExtendedWebDriver's methods, but only the methods most useful to Protractor
+ * users are documented here.
```
juliemr: This comment should mention that these are available only if your server has appium.

juliemr:
Other than one comment LGTM, and CI is happy, yay!

lib/webdriver-js-extender/index.js
sjelin: Done (see below)

sjelin:
I can't quite check docgen due to #3869, but it looks about as good as it could considering, and before I switch to the beta branch (back in https://github.com/angular/protractor/pull/3326) docgen was good.  I'll merge as soon as I hear back from CI.

lib/webdriver-js-extender/index.js
gkohen: @sjelin , do you see this latest commit as something that can assist testing hybrid app with Protractor?

thyagab:
@juliemr are  selectors to identify elements with accessability id's for native ios app are supported?

wswebcreation:
@thyagab ,

I dont think it's wise to `@` mention individual members. There are a lot contributers that collaborate on the issue PR.

As an answer for your question. [This is the browser support list](https://github.com/angular/protractor/blob/master/docs/browser-support.md). Protractor doesn't (officially) supports Native apps with native selectors.

In a previous project I worked with a Hybrid App with native elements. The app opened with a webview, but some screens were native. To access the native elements with Protractor I used `wd` and `wd-bridge`. I then needed to:

- change the context from `webview` to `native` with for example `wdBrowser.context(context)` where `context` was or `webview` or `native`
- click on an element with for example `wdBrowser.elementById(`something:id/key_cell0`).click()`

Hopefully this helps you

Grtz

Wim

thyagab:
Thank you for response. We are looking for native app support which seems not there at thi spoint.

PR #3866, opened 2016-12-21
feat(element): clearing inputs for type date, time, week, month, and datetime-local

### LINKED ISSUE #562: clear() or sendKeys('') not working on input types: date, time, datetime-local, month, or week
This directly relates to the documentation tests on a pull request on the Angular.js project: https://github.com/angular/angular.js/pull/5864

For now I'm going to work around it, I guess. But the issue seems to be that due to the custom controls in browsers like Chrome, protractor simply doesn't know what to do to interact with them in order to clear them. It's sort of late, and I haven't dug any further into than that for now.

I did, however see a few StackOverflow questions that seemed to relate, although none of the workarounds I found there worked.

http://stackoverflow.com/questions/15360362/clear-date-input-fails-on-chromewebdriver

Also possibly related?  

http://stackoverflow.com/questions/21295314/how-to-set-html5-type-date-input-fields-e-g-in-chrome-using-selenium-protra

cnishina:
closes #562


cnishina:
No no Craig... you should add tests....

Issue #1205, opened 2014-08-18
Drag and Drop is confusing. Document better, or create a better syntax to work with it


juliemr:
Using an action sequence is hard, and causes a lot of confusion. Let's discuss how to make this easier.

pidupuis:
Any advance on this topic ?
I am currently not able at all to simulate a drag and drop... I would really like to have a better flexibility and a better documentation

mgiambalvo:
We're not going to make any API changes. However, I'll do the following

- [] Add the browser.actions() API to our docs
- [] mention this in the FAQ (and also mention that HTML5 DnD events do not work in Chrome due to a known [chromedriver issue](https://bugs.chromium.org/p/chromedriver/issues/detail?id=841))
- [] add a cookbook example for doing drag and drop

PR #3867, opened 2016-12-21
docs(actions): Improve documentation around drag and drop.

mgiambalvo:
Addresses #1205

### DIFF
--- a/lib/selenium-webdriver/webdriver.js
+++ b/lib/selenium-webdriver/webdriver.js
@@ -7,6 +7,20 @@
 
 goog.provide('webdriver');
 
+/**
+ * Class for defining sequences of complex user interactions.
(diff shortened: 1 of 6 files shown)


lib/selenium-webdriver/webdriver.js
```diff
@@ -25,20 +39,37 @@ goog.provide('webdriver');
 webdriver.WebDriver = function() {};
 
 /**
- * Creates a new action sequence using this driver. The sequence will not be
+ * Creates a sequence of mouse actions using this driver. The sequence will not be
```
juliemr: Can't this include other things like `sendKeys`? I don't think we should specify 'mouse' actions.

mgiambalvo: Yup, I'll change it.

PR #3871, opened 2016-12-22
fix(isElementPresent): for un-wrapped `WebElement`s, `browser.isElementPresent` was broken

### LINKED ISSUE #3864: Bug - browser.d.ts isElementPresent using ProtractorBy instead of Locator
Just started converting our older Protractor tests to start using TypeScript and I noticed I was getting errors when using `browser.isElementPresent(by.css('example-css'))` saying that it should be a type of `ProtractorBy`.

I'm pretty sure this is supposed to be `Locator` like it is in the element.d.ts file.

I'm using protractor version 4.0.13. And looking at 4.0.14 doesn't look like that was updated to match either. 

If I'm wrong please let me know what I should be using for `browser.isElementPresent` or should I just abstain from using that and using `element(by.id('parent-element')).isElementPresent(by.css('example-css')`. Going to continue to see if I can actually get the `element(by.id('parent-element')).isElementPresent(by.css('example-css')` working.

sjelin:
Closes #3864

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -589,10 +589,16 @@ export class ProtractorBrowser extends Webdriver {
    * @returns {!webdriver.promise.Promise} A promise that will resolve to whether
    *     the element is present on the page.
    */
-  isElementPresent(locatorOrElement: ProtractorBy|WebElement): wdpromise.Promise<any> {
-    let element =
-        ((locatorOrElement as any).isPresent) ? locatorOrElement : this.element(locatorOrElement);
-    return (element as any).isPresent();
+  isElementPresent(locatorOrElement: Locator|WebElement|ElementFinder): wdpromise.Promise<any> {
+    let element: ElementFinder;
+    if (locatorOrElement instanceof ElementFinder) {
+      element = locatorOrElement;
+    } else if (locatorOrElement instanceof WebElement) {
+      element = ElementFinder.fromWebElement_(this, locatorOrElement);
+    } else {
+      element = this.element(locatorOrElement);
+    }
+    return element.isPresent();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lib/browser.ts
```diff
@@ -589,10 +589,16 @@ export class ProtractorBrowser extends Webdriver {
...
    */
-  isElementPresent(locatorOrElement: ProtractorBy|WebElement): wdpromise.Promise<any> {
-    let element =
-        ((locatorOrElement as any).isPresent) ? locatorOrElement : this.element(locatorOrElement);
-    return (element as any).isPresent();
+  isElementPresent(locatorOrElement: Locator|WebElement): wdpromise.Promise<any> {
+    let element: ElementFinder;
+    if ((locatorOrElement as any).isPresent) {
+      element = locatorOrElement as ElementFinder;
+    } else if (locatorOrElement instanceof WebElement) {
```
sjelin: Raw `WebElement`s don't have `isPresent`, hence the need for this fix

lib/browser.ts
```diff
@@ -589,10 +589,16 @@ export class ProtractorBrowser extends Webdriver {
    * @returns {!webdriver.promise.Promise} A promise that will resolve to whether
    *     the element is present on the page.
    */
-  isElementPresent(locatorOrElement: ProtractorBy|WebElement): wdpromise.Promise<any> {
-    let element =
-        ((locatorOrElement as any).isPresent) ? locatorOrElement : this.element(locatorOrElement);
-    return (element as any).isPresent();
+  isElementPresent(locatorOrElement: Locator|WebElement): wdpromise.Promise<any> {
+    let element: ElementFinder;
+    if ((locatorOrElement as any).isPresent) {
```
cnishina: Either the method signature has to also include ElementFinder or we should remove this conditional.

Add in ElementFinder to signature:
`locatorOrElement: Locator|ElementFinder|WebElement`

Change conditional for ElementFinder:
`if (locatorOrElement instanceof ElementFinder) {`

sjelin: Done

PR #3882, opened 2016-12-27
fix(ci): Log test output on CircleCI

mgiambalvo:
CircleCI will timeout the test if it doesn't see any output for 10
minutes. When running in CI, we need to inherit stdio so we don't get
killed if the tests run slowly due to timeouts. This will also give us
more info for debugging.

### DIFF
--- a/scripts/test.js
+++ b/scripts/test.js
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+var path = require('path');
 
 var Executor = require('./test/test_util').Executor;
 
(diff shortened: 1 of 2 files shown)


sjelin:
The circle ci output is *really* long now and hard to skim, especially because of the failure tests.  I would rather see each line of output replaces with a `'.'`.  This would avoid timeouts while preserving the compactness of the output

mgiambalvo:
I changed it to dots and also saved the test output as a build artifact.  I'm hoping this will help debug the intermittent timeouts we're seeing for the beta branch, which I'm thinking are due to a problem with Blocking Proxy and alert handling.

mgiambalvo:
Test failure is the javascript dialog problem from chromedriver - going to go ahead and merge.

Issue #3879, opened 2016-12-26
protractor.Key.chord() not transpiling


pittgoose:
I just upgraded to protractor 4.0.14 and I'm getting a Typescript compiler error on the following code:
`browser.actions().sendKeys(protractor.Key.chord(protractor.Key.CONTROL, "a")).perform();`
error:
 `Error:(47, 55) TS2339:Property 'chord' does not exist on type 'IKey'.`
I know that 4.0.14 addressed transpiling issues with webdriver. Not sure if this could've broken it.

Any help here would be greatly appreciated.

Node: 6.9.1
OS: Windows 10

cnishina:
If you are using the @types/selenium-webdriver that comes with 4.0.14, it is because chord method does not exist in the typings. This has been fixed with https://github.com/DefinitelyTyped/DefinitelyTyped/pull/13513. The fix should be in the latest version 2.53.38.

### Option 1. Set protractor.Key to any:

```
browser.actions().sendKeys((protractor.Key as any).chord(protractor.Key.Control, "a")).perform();
```

We just set protractor.Key to any so the chord method will just work.

### Option 2. May or may not work (needs investigation):

Install the latest version of @types/selenium-webdriver: 2.53.38+. I am not what happens with different installed version of @types/selenium-webdriver.

cnishina:
Verified: Use option 2 and install a newer version of @types/selenium-webdriver. This appears to override the installed types by Protractor.

Closing this issue since the last item for investigation has been answered.

PR #3886, opened 2016-12-28
Update @types/selenium-webdriver dependency

pittgoose:
Fixes issue #3879 (https://github.com/angular/protractor/issues/3879)
Updating @types/selenium-webdriver dependency to @types/selenium-webdriver@2.53.38 which added back the protractor.Key.chord method.


pittgoose:
I signed it!

sjelin:
LGTM but @cnishina is the type expert

PR #3893, opened 2016-12-29
chore(docs): new framework requirements in Protractor 6.0

sjelin:
This will allow us to implement `restartBrowserBetweenTests` without the control flow, and in the future maybe add more features users have been asking for.

### DIFF
--- a/lib/frameworks/README.md
+++ b/lib/frameworks/README.md
@@ -3,47 +3,80 @@ Framework Adapters for Protractor
 
 Protractor can work with any test framework that is adapted here.
 
-Each file details the adapter for one test framework. Each file must export a `run` function with the interface:
+Each file details the adapter for one test framework. Each file must export a
+`run` function with the interface:
 
-```js
+```ts
 /**
  * @param {Runner} runner The Protractor runner instance.
  * @param {Array.<string>} specs A list of absolute filenames.
- * @return {q.Promise} Promise resolved with the test results
+ * @return {Promise.<Object>} Promise resolved with the test results.  See
+ *   "Requirements" section for details.
  */
-exports.run = function(runner, specs)
+export let run: (runner: Protractor.Runner, specs: string) => Promise<Object>
 ```
 
 Requirements
 ------------
 
- - `runner.emit` must be called with `testPass` and `testFail` messages.  These
-   messages must be passed a `testInfo` object, with a `name` and `category`
-   property.  The `category` property could be the name of the `describe` block
-   in jasmine/mocha, the `Feature` in cucumber, or the class name in something
-   like jUnit.  The `name` property could be the name of an `it` block in
-   jasmine/mocha, the `Scenario` in cucumber, or the method name in something
-   like jUnit.
+- `runner.emit` must be called with `testPass` and `testFail` messages. These
+  messages must be passed a `testInfo` object with the following structure:
 
- - `runner.runTestPreparer` must be called before any tests are run.
+  ```ts
+  testInfo: {
+    category: string,
+    name: string
+  }
+  ```
 
- - `runner.getConfig().onComplete` must be called when tests are finished.
-   It might return a promise, in which case `exports.run`'s promise should not
-   resolve until after `onComplete`'s promise resolves.
+  The `category` property could be the name of the `describe` block in
+  jasmine/mocha, the `Feature` in cucumber, or the class name in something like
+  jUnit.
+  The `name` property could be the name of an `it` block in jasmine/mocha, the
+  `Scenario` in cucumber, or the method name in something like jUnit.
 
- - The returned promise must be resolved when tests are finished and it should return a results object. This object must have a `failedCount` property and optionally a `specResults`
- object of the following structure:
- ```
-   specResults = [{
-     description: string,
-     assertions: [{
-       passed: boolean,
-       errorMsg: string,
-       stackTrace: string
-     }],
-     duration: integer
-   }]
- ```
+- `runner.runTestPreparer` must be called after the framework has been
+  initialized but before any spec files are run.  This function returns a
+  promise which should be waited on before executing tests.
+
+- `runner.getConfig().onComplete` must be called when tests are finished.
+  It might return a promise, in which case `exports.run`'s promise should not
+  resolve until after `onComplete`'s promise resolves.
+
+- The returned promise must be resolved when tests are finished and it should
+  return a results object. This object must have a `failedCount` property and
+  optionally a `specResults` object of the following structure:
+
+  ```ts
+  specResults: [{
+    description: string,
+    assertions: [{
+      passed: boolean,
+      errorMsg: string,
+      stackTrace: string
+    }],
+    duration: integer
+  }]
+  ```
+
+### Future requirements
+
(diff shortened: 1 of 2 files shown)


juliemr:
See issue at https://github.com/angular/protractor/issues/3894

lib/frameworks/README.md
```diff
@@ -17,33 +17,74 @@ exports.run = function(runner, specs)
...
+  `testInfo` object from the old requirements.
+
+- `runner.beforeAll` must be called after the framework has been initialized but
+  before any spec files are run.  This function returns a promise which should
+  be waited on before executing tests.
+
+- The returned promise must follow the same rules as before.
+
+Note that `testPass` and `testFail` events will be replaced by
+`runner.afterEach`, `runner.runTestPreparer` will be replaced by `beforeAll`,
```
juliemr: Why change the API for `runTestPreparer`? This seems like unnecessary churn.

lib/frameworks/README.md
```diff
@@ -17,33 +17,74 @@ exports.run = function(runner, specs)
...
+  The `category` and `name` properties as the same as the ones that were in the
+  `testInfo` object from the old requirements.
+
+- `runner.beforeAll` must be called after the framework has been initialized but
+  before any spec files are run.  This function returns a promise which should
+  be waited on before executing tests.
+
+- The returned promise must follow the same rules as before.
+
+Note that `testPass` and `testFail` events will be replaced by
```
juliemr: I think I'd prefer to make minimal changes here. I think we can keep the `testPass` and `testFail` events, since there are good and useful patterns for wanting to do event-based handling of those things without blocking the asynchronous flow. I think we should just add the `afterEach` (or `beforeEach`, not sure it really matters where we put the edge case).

lib/frameworks/README.md
```diff
@@ -17,33 +17,74 @@ exports.run = function(runner, specs)
...
+`runner.afterEach`, `runner.runTestPreparer` will be replaced by `beforeAll`,
+and `runner.getConfig().onComplete` will be handled automatically.
+
+Frameworks following the old specification will still work until at least
+Protractor 7.0.  If you write a backwards-compatible framework, Protractor will
+automatically detect that and handle deduplication.  However, do not *partially*
```
juliemr: If we go with my comments above and only add the `afterEach` function, then we don't need this warning.

lib/frameworks/README.md
```diff
@@ -17,33 +17,74 @@ exports.run = function(runner, specs)
...
+  }>
+  ```
+
+
+### Future changes
+
+In Protractor 6.0, the requirements will be changed to the following:
+
+- `runner.afterEach` will have to be called after each test finishes.  It must
+  be called with a `testInfo` object of the following structure:
```
juliemr: Should mention that this needs to wait on a returned promise.

lib/frameworks/README.md
sjelin: We would still have the runner emit `testPass` and `testFail`, we would just do that in `runner.afterEach`.  So instead of having the framework handle event emission we'd do it ourselves.  And I think it's pretty redundant to have the frameworks do both.

lib/frameworks/README.md
sjelin: Just to make it more consistent

lib/frameworks/README.md
juliemr: Another potential problem: I think frameworks should treat `afterEach` as part of the test, so if that errors, it could potentially change the test outcome from pass to fail. And the framework might not even have access to pass/fail stats before it's fully done.

sjelin: Ok, I think you're right

lib/frameworks/README.md
juliemr: I don't think that's worth the braking change, if we aren't updating the emitters.

sjelin:
@juliemr All comments addressed, but you're going to need to take a look again 😛

lib/frameworks/README.md
```diff
@@ -3,47 +3,91 @@ Framework Adapters for Protractor
...
+- `runner.getConfig().onComplete` must be called when tests are finished.
+  It might return a promise, in which case `exports.run`'s promise should not
+  resolve until after `onComplete`'s promise resolves.
+
+- The returned promise must be resolved when tests are finished and it should
+  return a results object. This object must have a `failedCount` property and
+  optionally a `specResults` object of the following structure:
+
+  ```ts
+  specResults: Array<{
```
juliemr: I don't think the switch to `Array` makes this more readable.

lib/frameworks/README.md
```diff
@@ -3,47 +3,91 @@ Framework Adapters for Protractor
...
+
+In Protractor 6.0, the following additional requirement will be added:
+
+- `runner.afterEach` will have to be called after each test finishes.  It will
+  return a promise, which should be waited for before moving onto the next test.
+
+  If your framework does not export a `protractorVersion` property which is at
+  least 6.0.0, and `testPass` or `testFail` are emitted before
+  `runner.afterEach` is called, Protractor will assume your framework is old and
+  does not support `runner.afterEach`.  In this case, Protractor will call
```
juliemr: Why don't we just check for presence of `afterEach` on the object? This seems simpler than requiring a version export.

sjelin: The framework calls us, not the other way around.  There's no way to check that they plan on calling us

lib/frameworks/README.md
sjelin: I was trying to make it an actual typescript type.  But I'll change it back, since this isn't real code anyway.

lib/frameworks/README.md
sjelin: We could also add a requirement that they call `afterEach` before they emit `testPass` or `testFail`.  Jasmine and Mocha will certainly do this.  But I'm concerned that this might be difficult for other frameworks.

lib/frameworks/README.md
sjelin: Done

lib/frameworks/README.md
juliemr: Ah of course, my bad. Asking for a version still feels like a very blunt solution though - we could instead warn if `afterEach` isn't being called at all (might not get real-time errors, but we could certainly warn if, say, `afterEach` wasn't called during an entire run)

sjelin: So let's say we see a `testPass` event but haven't seen `afterEach` yet and `restartBrowserBetweenTests` is on.  Do we assume it's an old framework and call `browser.restart` now?  Do we wait in hopes that `afterEach` will be called in the future?

juliemr: I'd say we log a warning if we see that behavior and move on without trying to fix it. Too harsh?

sjelin: I guess it's fine.  So we'd just break `restartBrowserBetweenTests` on old frameworks then?

juliemr: Right. And obviously we'll update frameworks that we control before we make the breaking change.

sjelin:
@juliemr all comments addressed.

juliemr:
LGTM!

PR #3906, opened 2016-12-30
chore(frameworks): allow frameworks to call `afterEach`.

sjelin:
Follow up to https://github.com/angular/protractor/pull/3893.

Part of #3893

I could have just made it a noop I guess but I think it makes sense to actually use it.

### DIFF
--- a/lib/runner.ts
+++ b/lib/runner.ts
@@ -33,6 +33,8 @@ export class Runner extends EventEmitter {
   driverprovider_: DriverProvider;
   o: any;
   plugins_: Plugins;
+  restartPromise: q.Promise<any>;
+  frameworkUsesAfterEach: boolean;
 
   constructor(config: Config) {
     super();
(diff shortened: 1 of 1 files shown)


lib/runner.ts
```diff
@@ -325,8 +347,12 @@ export class Runner extends EventEmitter {
           }
 
           if (this.config_.restartBrowserBetweenTests) {
+            // TODO(sjelin): replace with warnings once `afterEach` support is required
             let restartDriver = () => {
-              browser_.restart();
+              if (!this.afterEachWasCalled) {
```
juliemr: Maybe if we name this boolean `frameworkUsesAfterEach` it would be clearer? I confused myself for a bit thinking we'd need to reset it to false (which we don't, obviously).

juliemr:
LGTM with one consideration.

PR #3920, opened 2017-01-03
chore(index): publish interfaces instead of classes

### LINKED ISSUE #3907: Publish interfaces instead of classes in `index.ts`
I recently discovered a fun trick to only publish the interface instead of a class:

```ts
import {Runner} from './runner';
...
export type Runner = Runner;
```

Isn't that fun?  I think we should probably use it for most of our exports in `index.ts`.

cnishina:
closes #3907


cnishina:
Looks like the typings don't work. Closing this.

Issue #3216, opened 2016-05-12
Throw an error in the launcher if the user sets an unknown flag


mgiambalvo:
Some of our flags are very easy to mistype. Rather than silently ignoring unknown flags, we should at least show a warning and probably exit with an error if a user invokes Protractor with an unknown flag.

juliemr:
Hm, but users might be processing args themselves

cnishina:
Instead of throwing an error, should we warn the user that we encountered something unexpected?

PR #3921, opened 2017-01-03
chore(cli): throw errors on unknown flags

cnishina:
clsoes #3216


cnishina:
Test?

sjelin:
Yes, tests please! :stuck_out_tongue:

lib/cli.ts
```diff
@@ -31,49 +31,97 @@ process.argv.slice(2).forEach(function(arg: string) {
...
-      if (arg._.length > 1) {
-        throw new Error('Error: more than one config file specified');
-      }
-    });
+export interface Option {
+  key: string;
+  value?: string;
+}
+
+let describes: Option[] = [
```
sjelin: if you use an object instead of an array of `Option`, you can really simplify the code at the end

lib/cli.ts
```diff
@@ -31,49 +31,87 @@ process.argv.slice(2).forEach(function(arg: string) {
...
 
+// Check to see if additional flags were used.
+let unknownKeys: string[] = [];
+for (let key of Object.keys(argv).slice(2)) {
+  if (key === '$0') {
+    continue;
+  }
+  let found = false;
+  for (let optimistKey of Object.keys(optimistOptions)) {
+    for (let optionKey of Object.keys(optimistOptions[optimistKey])) {
```
sjelin: Replace inner loop with `found = found || (optimistOptions[optimistKey][key] !== undefined)`

lib/cli.ts
```diff
@@ -31,49 +31,82 @@ process.argv.slice(2).forEach(function(arg: string) {
...
+  if (arg._.length > 1) {
+    throw new Error('Error: more than one config file specified');
+  }
+});
 
 let argv: any = optimist.parse(args);
 
+// Check to see if additional flags were used.
+let unknownKeys: string[] = [];
+for (let key of Object.keys(argv).slice(2)) {
```
sjelin: Please use [`Array.prototype.filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) instead of rolling your own version

lib/cli.ts
```diff
@@ -31,49 +31,82 @@ process.argv.slice(2).forEach(function(arg: string) {
...
+  if (key === '$0') {
```
sjelin: I'd rather you just initialized `found` to `key !== '$0'`, since this check is pretty fast.

lib/cli.ts
```diff
@@ -31,48 +31,83 @@ process.argv.slice(2).forEach(function(arg: string) {
...
+  if (arg._.length > 1) {
+    throw new Error('Error: more than one config file specified');
+  }
+});
+
+let argv: any = optimist.parse(args);
+
+// Check to see if additional flags were used.
+let unknownKeys: string[] = [];
+Object.keys(argv)
```
sjelin: noooo..... Ok, I can see where the miscommunication happened.  There's what I was trying to say:

```ts
let unknownKeys = Object.keys(argv).slice(2).filter((key) => {
  let found = element !== '$0';
  for (let optimistKey of Object.keys(optimistOptions)) {
    found = found || (optimistOptions[optimistKey][currentValue] !== undefined);
  }
  return found
});
```

lib/cli.ts
```diff
@@ -84,6 +175,17 @@ if (argv.version) {
   process.exit(0);
 }
 
+if (!argv.disableChecks) {
+  // Check to see if additional flags were used.
+  let unknownKeys: string[] = Object.keys(argv).filter((element: string) => {
```
sjelin: (I'm not gonna make you do anymore with this, but it just occurred to me that [`Array.prototype.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) is designed exactly for this situation:

```ts
let hasUnknownKeys = Object.keys(argv).some((element: string) => {
  return element !== '$0' && element !== '_' && allowedNames.indexOf(element) === -1;
});

if (hasUnknownKeys) {
  throw new Error('Found extra flags: ' + unknownKeys.join(', '));
}
```
It's so beautiful)

cnishina: So beautiful...and.....merging.

PR #3922, opened 2017-01-03
chore(local): if no local drivers provided in config, use update-config.json

### LINKED ISSUE #3782: Local driverProvider should use the update-config.json for binaries
The binaries downloaded from webdriver-manager should be referenced from `update-config.json`

cnishina:
closes #3782

### DIFF
--- a/spec/unit/driverProviders/direct_test.js
+++ b/spec/unit/driverProviders/direct_test.js
@@ -1,17 +1,21 @@
 var fs = require('fs'),
     os = require('os'),
     path = require('path');
-var BrowserError = require('../../../built/exitCodes').BrowserError,
-    Logger = require('../../../built/logger2').Logger,
-    WriteTo = require('../../../built/logger2').WriteTo;
+var BrowserError = require('../../../built/exitCodes').BrowserError;
+var ProtractorError = require('../../../built/exitCodes').ProtractorError;
+var Logger = require('../../../built/logger').Logger;
+var WriteTo = require('../../../built/logger').WriteTo;
+var Direct = require('../../../built/driverProviders').Direct;
 var webdriver, file;
 
 describe('direct connect', function() {
   beforeEach(function() {
+    ProtractorError.SUPRESS_EXIT_CODE = true;
     Logger.setWrite(WriteTo.NONE);
   });
 
   afterEach(function() {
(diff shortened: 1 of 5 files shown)


cnishina:
Test?

sjelin:
We already have tests using `local` right?  They should be fine

cnishina:
This is a special case that does not have a high probability of happening. If the binaries are not specified in the config, the binaries were downloaded but the `update-config.json` was not written, it would throw a similar error to run `webdriver-manager update`.

cnishina:
As discussed offline, no test required.

spec/unit/driverProviders/direct_test.js
```diff
@@ -57,9 +64,43 @@ describe('direct connect', function() {
...
+    beforeEach((done) => {
+      // webdriver-manager clean
+      var argv = {
+        '_': ['update'],
+        'standalone': false,
+        'gecko': false
+      };
+      webdriverUpdate.run(JSON.parse(JSON.stringify(argv))).then(() => {
+        const SeleniumConfig = require('webdriver-manager/built/lib/config').Config;
+        rimraf.sync(path.resolve(SeleniumConfig.getSeleniumDir(), 'update-config.json'));
```
sjelin: I don't like this because:
1) You're actually changing the file system, which is an unexpected side effect of running these tests
2) Unit tests shouldn't require you to have the filesystem set up in any particular  setup

Please either use a mock file system of some kind or change `lib/driverProviders/local.ts` so you can call `addDefaultBinaryLocs_()` without it interacting with the file system.

sjelin: Per @juliemr's recommendation, the best solution for now is probably to use a jasmine spy on `fs.readFileSync` and `fs.existsSync`

lib/driverProviders/index.ts
```diff
@@ -49,6 +49,7 @@ export let buildDriverProvider = (config: Config): DriverProvider => {
     driverProvider = new Mock(config);
     logWarnings('mock', config);
   } else {
+    config.seleniumAddress = 'http://127.0.0.1:4444/wd/hub';
```
sjelin: Remove this temp code

PR #3927, opened 2017-01-04
fix(browser): Set rootEl in WebDriver ControlFlow.

### LINKED ISSUE #3905: Wrap `useAllAngular2AppRoots` and `rootEl` in control flow
Like we did for `ignoreSynchronization` (see `beta` branch if we haven't merged `beta` into `master` yet).  @mgiambalvo I'm assigning you, since it seems blocking-proxy related.  This will be a breaking change, so we might want to hold off on the 5.0 release until we do this?

mgiambalvo:
This makes sure that people who change rootEl in their tests (which is
uncommon, but potentially useful) will see the right behavior. Fixes #3905

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -191,7 +191,19 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
    *
    * @type {string}
    */
-  rootEl: string;
+  set rootEl(value) {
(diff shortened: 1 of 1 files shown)


mgiambalvo:
Actually, this is a bad idea. See #3905

PR #3963, opened 2017-01-12
fix: no longer use es6 let statement

DevVersion:
* Currently in Protractor v5 the Angular detection script uses ES6 features like the `let` modifier. 

This can break Protractor on browsers, which doesn't support those statements. 
> See https://saucelabs.com/beta/tests/275f75091dac40a0a3374d29d912caee/commands#11

**Note**: This currently makes Protractor v5 hard to use on older browsers. (On Material 2 we will update browsers)

cc. @sjelin @cnishina


RomanGotsiy:
@DevVersion do you have any estimates when new protractor version with this fix will be released?
thanks

mgiambalvo:
The plan is Monday or early next week.

PR #3964, opened 2017-01-12
display disableChecks option in extra flags error message

igniteram:
With reference to this recent [SO question](http://stackoverflow.com/questions/41592870/found-extra-flags-error-in-new-version-of-protractor) , it would be good to display the `--disableChecks` option in the error message thrown in case users have their own custom flags set in the cli.

Users are not aware of this option until they check the changelog, it would help if this is displayed along with the error message thrown in the console itself. 

@cnishina Since you worked on this, any thoughts?

### DIFF
--- a/lib/cli.ts
+++ b/lib/cli.ts
@@ -181,7 +181,7 @@ if (!argv.disableChecks) {
   });
 
   if (unknownKeys.length > 0) {
-    throw new Error('Found extra flags: ' + unknownKeys.join(', '));
+    throw new Error('Found extra flags: ' + unknownKeys.join(', ') + ', please use --disableChecks flag to disable the custom flag check!');
   }
 }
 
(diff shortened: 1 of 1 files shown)


lib/cli.ts
```diff
@@ -181,7 +181,7 @@ if (!argv.disableChecks) {
   });
 
   if (unknownKeys.length > 0) {
-    throw new Error('Found extra flags: ' + unknownKeys.join(', '));
+    throw new Error('Found extra flags: ' + unknownKeys.join(', ') + ', please use --disableChecks flag to disable the custom flag check!');
```
cnishina: The wording could be improved.
`please use --disableChecks flag to disable the Protractor CLI flag checks.`

cnishina: Also, CircleCI does not like your typescript formatting. Please run `npm run format` before submitting.

igniteram: Yup, noticed that let me change & format it!

Issue #3744, opened 2016-11-15
bug(element): el.serialize has been removed from WD


sjelin:
Judging by https://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_WebElement.html, there is no replacement.  We should remove it from the list of functions we wrap

PR #3966, opened 2017-01-12
chore(webdriver): remove `element.serialize`, since it is no longer p…

sjelin:
…art of webdriver

Closes https://github.com/angular/protractor/issues/3744


cnishina:
Format errors on circle Ci

sjelin:
Fixed, merging

PR #3970, opened 2017-01-13
chore(blockingproxy): Allow using a pre-existing Blocking Proxy instance

mgiambalvo:
Only intended for debugging.

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -106,8 +106,7 @@ export interface Config {
 
   /**
    * If specified, Protractor will connect to the Blocking Proxy at the given
-   * url instead of starting it's own. Intended for debugging Protractor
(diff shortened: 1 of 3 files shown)


lib/config.ts
```diff
@@ -104,6 +104,13 @@ export interface Config {
    */
   useBlockingProxy?: boolean;
 
+  /**
+   * If specified, Protractor will connect to the Blocking Proxy at the given
+   * url instead of starting it's own. Intended for debugging Protractor
+   * features only.
```
juliemr: Do we need this 'intended for' warning? There's no big issue with users doing it this way, right?

mgiambalvo: Fair enough, done.

PR #3958, opened 2017-01-11
Adding test case to show visibilityOf problem with race condition (#3…

tilmanpotthof:
Test case for #3777

I'll add the fix after a build failed to show the test case is correct.

Issue #3777, opened 2016-12-01
ExpectedConditions.invisibilityOf cannot reliably be used to wait for an element to be removed from the DOM


massimocode:
Protractor build 4.0.11

invisibilityOf states the following in the API documentation:
`An expectation for checking that an element is either invisible or not present on the DOM. This is the opposite of 'visibilityOf'`

Please see the following stack trace that we get intermittently from our build server when using invisibilityOf with browser.wait to wait for an element to be removed from the DOM:
```
NoSuchElementError: No element found using locator: By(css selector, .new-position-modal)
    at WebDriverError (C:\BuildAgent\work\5d4c0dd4e428c81b\node_modules\selenium-webdriver\lib\error.js:27:5)
    at NoSuchElementError (C:\BuildAgent\work\5d4c0dd4e428c81b\node_modules\selenium-webdriver\lib\error.js:242:5)
    at C:\BuildAgent\work\5d4c0dd4e428c81b\node_modules\protractor\built\element.js:698:27
    at ManagedPromise.invokeCallback_ (C:\BuildAgent\work\5d4c0dd4e428c81b\node_modules\selenium-webdriver\lib\promise.js:1379:14)
    at TaskQueue.execute_ (C:\BuildAgent\work\5d4c0dd4e428c81b\node_modules\selenium-webdriver\lib\promise.js:2913:14)
    at TaskQueue.executeNext_ (C:\BuildAgent\work\5d4c0dd4e428c81b\node_modules\selenium-webdriver\lib\promise.js:2896:21)
    at asyncRun (C:\BuildAgent\work\5d4c0dd4e428c81b\node_modules\selenium-webdriver\lib\promise.js:2775:27)
    at C:\BuildAgent\work\5d4c0dd4e428c81b\node_modules\selenium-webdriver\lib\promise.js:639:7
    at process._tickCallback (internal/process/next_tick.js:103:7)Error
    at ElementArrayFinder.applyAction_ (C:\BuildAgent\work\5d4c0dd4e428c81b\node_modules\protractor\built\element.js:395:27)
    at ElementArrayFinder._this.(anonymous function) [as isDisplayed] (C:\BuildAgent\work\5d4c0dd4e428c81b\node_modules\protractor\built\element.js:97:30)
    at ElementFinder._this.(anonymous function) [as isDisplayed] (C:\BuildAgent\work\5d4c0dd4e428c81b\node_modules\protractor\built\element.js:721:22)
[...]

Issue #3578, opened 2016-09-20
Provide documentation on how to use correctly borwser.wait with protractor.ExpectedConditions


evilaliv3:
I'm trying to use the protractor.ExpectedConditions together with browser.wait() but i'm sporadically failing with the error: ```stale element reference: element is not attached to the page document.```

The way i'm using the function is the following:
```
function (locator, timeout) {
  var t = timeout === undefined ? exports.browserTimeout() : timeout;
  var EC = protractor.ExpectedConditions;
  browser.wait(EC.invisibilityOf(element(locator)), t);
};
```

the reason that is causing the failure is probably that between the call to element(locator) and the usage of the element i've applications refresh that remove the element and cause the issue when the refernece is used.

i've not found any good documentation to use this routines without having this issue.

could you please provide some guidelines?

evilaliv3:
@juliemr do you have by the chance a tip for me?

evilaliv3:
@juliemr it seems to me that this is related to https://github.com/angular/protractor/issues/2277

is it possible that there is a regression?

evilaliv3:
Solved using element.isVisible in place of the ExpectedCondition:

```
function (locator, timeout) {
  var t = timeout === undefined ? exports.browserTimeout() : timeout;
  return browser.wait(function() {
    return element(locator).isDisplayed().then(function(present) {
      return present;
    });
  }, t);
};
```

i would suggest @cnishina that also for EC.invisibilityOf you adopt the same returning false in place of raising an exception.

evilaliv3:
No damn, i'm still getting the error in a sporadic way even using the isDisplayed().

i confirm so that we are facing a regression on #2277

tilmanpotthof:
@evilaliv3 Hi, this issue is caused by a race condition between the isPresent and isDisplayed check. If the element is removed between both calls or while isDisplayed is executed you get a NoSuchElementError or a StaleElementReferenceError.

This problem is solved for visibilityOf (https://github.com/angular/protractor/pull/3958/files), but it still exists for textToBePresentInElement and probably others.

Before these changes are release in a future version your can monkey patch your ExpectedContions and use workaround functions that I put in an npm module

https://www.npmjs.com/package/protractor-save-expected-conditions

PR #3972, opened 2017-01-13
fix(ExpectedConditions): allow ExpectedConditions to handle missing e…

sjelin:
…lements

Expected conditions used `presenceOf` and `visibilityOf` to check that it's
referencing elements which actually exist on the page, but there is a race
condition with this strategy: an element could disappear after the
`presenceOf`/`visibilityOf` check but before other checks, causing an error
to be thrown.  This PR handles this race condition in two ways:

1. `ElementFinder`'s `isEnabled`, `isDisplayed`, and `isSelected` functions now
   return false if no such element exists, rahter than throwing an error
2. `ExpectedConditions`'s `textToBePresent` and `textToBePresentInElementValue`
   now check for errors and also return false in those cases

This is a general solution to the problem referenced in
https://github.com/angular/protractor/issues/3777 and
https://github.com/angular/protractor/pull/3958.

I'm assigning @juliemr, since it's a low level change, and @mgiambalvo, since we discussed this issue earlier today

### DIFF
--- a/lib/element.ts
+++ b/lib/element.ts
@@ -4,6 +4,7 @@ import {ElementHelper, ProtractorBrowser} from './browser';
 import {IError} from './exitCodes';
 import {Locator} from './locators';
 import {Logger} from './logger';
+import {falseIfMissing} from './util';
 
 let clientSideScripts = require('./clientsidescripts');
 
@@ -17,6 +18,8 @@ let WEB_ELEMENT_FUNCTIONS = [
   'getLocation', 'isEnabled', 'isSelected', 'submit', 'clear', 'isDisplayed', 'getId', 'serialize',
   'takeScreenshot'
 ] as (keyof WebdriverWebElement)[];
+let FALSE_IF_MISSING_WEB_ELEMENT_FUNCTIONS =
+    ['isEnabled', 'isSelected', 'isDisplayed'] as (keyof WebdriverWebElement)[];
 
 /**
  * ElementArrayFinder is used for operations on an array of elements (as opposed
@@ -82,7 +85,8 @@ export class ElementArrayFinder extends WebdriverWebElement {
   constructor(
       public browser_: ProtractorBrowser,
       public getWebElements: () => wdpromise.Promise<WebElement[]> = null, public locator_?: any,
-      public actionResults_: wdpromise.Promise<any> = null) {
+      public actionResults_: wdpromise.Promise<any> = null,
+      public falseIfMissing_: boolean = false) {
     super();
 
     // TODO(juliemr): might it be easier to combine this with our docs and just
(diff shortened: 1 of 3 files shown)


lib/expectedConditions.ts
```diff
@@ -210,7 +210,7 @@ export class ProtractorExpectedConditions {
         // MSEdge does not properly remove newlines, which causes false
         // negatives
         return actualText.replace(/\r?\n|\r/g, '').indexOf(text) > -1;
-      });
+      }, () => false);
```
mgiambalvo: Should we be more specific about which errors we mask here?

lib/element.ts
```diff
@@ -801,7 +807,14 @@ export class ElementFinder extends WebdriverWebElement {
       // Access the underlying actionResult of ElementFinder.
       this.then =
           (fn: (value: any) => any | wdpromise.IThenable<any>, errorFn?: (error: any) => any) => {
-            return this.elementArrayFinder_.then((actionResults: any) => {
+            return this.elementArrayFinder_.then(null, (error) => {
+              if (elementArrayFinder_.falseIfMissing_ &&
+                  (error instanceof wderror.NoSuchElementError)) {
```
mgiambalvo: Do we also need to handle StaleElement exceptions?

mgiambalvo:
This has also come up in https://github.com/angular/protractor/issues/3578#issuecomment-272435016, and there's an npm package out there to patch around the issue: https://github.com/tilmanpotthof/protractor-save-expected-conditions

In that 'protractor-save-expected-conditions' package, he masks NoSuchElement and StaleElement in `textToBePresent`, which is a little confusing to me since it looks like we already handle those, but maybe the way our `isPresent()` works still allows StaleElement errors to get through.

mgiambalvo:
Also we could probably simplify `ElementFinder.isPresent()` as part of this.

sjelin:
@mgiambalvo all comments addressed

sjelin:
ack, I forgot to run tests.  Give me a second to debug this

tilmanpotthof:
@mgiambalvo Hi, the problem in the expected conditions is not the `isPresent` check. It happens, when an element is removed exactly in the moment after the `isPresent` check and before `isDisplayed` in `visibilityOf`. Although node is single threaded, the browser thread is not synchronised with the webdriver checks.

The possibility of these race conditions is very low for a single test, but I had it in almost every build with >600 test in 2-10 random test. This started after I replaced all `browser.sleep` usages to make my tests more reliable :-) With my temporary npm module they are really more reliable, but it would be great to see these changes in 5.0.1

tilmanpotthof:
@sjelin I submitted a PR focusing on the expected conditions and ignoring the problems in the ElementFinder functions.

https://github.com/angular/protractor/pull/4006

mgiambalvo:
Closing this as superceded by #4006

PR #3977, opened 2017-01-15
chore(types): make ElementArrayFinder more generic

jnorkus:
Add some generics to ElementArrayFinder methods.


jnorkus:
I signed it!

Issue #3894, opened 2016-12-29
Add a method to the framework adapters interface to run async code before each test


juliemr:
This will be used for things such as restarting the browser between tests. To avoid relying on control flow, we need to be able to run asynchronous things (like waiting for the new browser instance to start up). 

This will look something like:

```js
JasmineAdapter {
  addRunBeforeFunction(fn) {
    jasmine.beforeEach((done) => {
      fn().then(done);
    });
  }
}
```

Issue #3908, opened 2016-12-30
Support `runner.afterEach` in jasmine adapter


sjelin:
See https://github.com/angular/protractor/pull/3893 and https://github.com/angular/protractor/issues/3894

PR #3988, opened 2017-01-19
feat(frameworks): Support `runner.afterEach` in jasmine and mocha ada…

sjelin:
…pter

Closes https://github.com/angular/protractor/issues/3894,
https://github.com/angular/protractor/issues/3908, and
https://github.com/angular/protractor/issues/3909

### DIFF
--- a/scripts/test.js
+++ b/scripts/test.js
@@ -76,7 +76,7 @@ executor.addCommandlineTest('node built/cli.js spec/errorTest/timeoutConf.js')
       message: 'Timeout - Async callback was not invoked within timeout ' +
           'specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.'
     })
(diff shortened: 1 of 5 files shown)


scripts/test.js
```diff
@@ -76,7 +76,7 @@ executor.addCommandlineTest('node built/cli.js spec/errorTest/timeoutConf.js')
       message: 'Timeout - Async callback was not invoked within timeout ' +
           'specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.'
     })
-    .expectTestDuration(0, 100);
+    .expectTestDuration(0, 1000);
```
sjelin: After a lot of bench-marking, I have concluded that this change only matters for extremely short tests like this one, and shouldn't matter to normal users.

PR #3991, opened 2017-01-20
fix(driverProviders): Check config in the right place.

mgiambalvo:
Fixes CI, that's what I get for not being patient.


lib/driverProviders/local.ts
```diff
@@ -115,6 +111,9 @@ export class Local extends DriverProvider {
     if (serverConf.port === undefined) {
       serverConf.port = this.config_.seleniumPort;
     }
+    if (!Array.isArray(serverConf.jvmArgs)) {
```
sjelin: Should probably be in an `else` block back on line 110

mgiambalvo: Good call, done.

Issue #3902, opened 2016-12-30
Don't rewrap WebDriver promise into `q` promise in `DriverProvider.prototype.quitDriver`


sjelin:
Necessary for #3899 so that we return a WebDriver promise and don't mix promise types.

We did this because `quitDriver`'s promise gets passed to `teardownEnv` and `shutdown_`, each of which use `q` promises.  But we'll just have to move the rewrapping one level up I guess.

Issue #3899, opened 2016-12-30
`browser.restart` should return a promise which resolves when the new browser is ready


sjelin:
Proposed implementation:

```ts
browser_.restart = () => {
  let quitPromise = this.driverprovider_.quitDriver(browser_.driver);
  function forkAndOverwrite() { ... }
  if (wdpromise.USE_PROMISE_MANAGER !== false) {
    forkAndOverwrite();
  } else {
    quitPromise = quitPromise.then(forkAndOverwrite);
  }
  return quitPromise.then(() => {
    return browser_.ready();
  });
};
```

I know this is messy but:
 * Should the control flow be on, the fork/overwrite needs to be synchronous to support code like:

   ```ts
   browser.restart();
   browser.get('http://www.google.com'); // Must go to new instance
   ```

 * Should the control flow be off, we shouldn't do the fork/overwrite until after we've quit, or else we'll briefly have two browser instances floating around.  Also, it'd just be weird and unexpected for the global browser instance to change synchronously when you're used to nothing happening until you `await` it.

sjelin:
Blocked on https://github.com/angular/protractor/issues/3902

sjelin:
Better solution to the original problem:

```ts
browser_.restart = () => {
  const driverPromise: wdpromise.Promise<WebDriver> = browser_.ready().then(() => {
    return this.driverprovider_.quitDriver(browser_.driver);
  }).then<WebDriver>(() => {
    return browser_.forkNewUnwrappedDriverInstance(false, true);
  });

  const flow = browser.driver.controlFlow();
  const executor = browser.driver.getExecutor();
  const session = driverPromise.then(driver => driver.getSession());
  const onQuit = extractOnQuit(driverPromise);
  browser_.setDriver_(new WebDriver(session, executor, flow, onQuit));
  return browser_.ready();
};
```

Unfortunately, the only `extractOnQuit` implementation I can think of is horrible:
```ts
function extractOnQuit(driverPromise: wdpromise.Promise<WebDriver>) {
  return function() {
    const args = arguments;
    return driverPromise.then((driver) => {
      if ((driver as any).onQuit_) {
        return (driver as any).onQuit_.apply(this, args);
      } else {
        const originalSchedule = driver.schedule;
        driver.schedule = () => driver.flow.promise<void>(resolve => resolve());
        const result = driver.quit().catch(function() {}); // catch any errors from double quit
        driver.schedule = originalSchedule;
      }
    });
  };
}
```
[...]

PR #3992, opened 2017-01-20
chore(quitDriver): have quitDriver return a webdriver promise directly

sjelin:
Wrapping it in a `q` promise is blocking https://github.com/angular/protractor/issues/3899

Closes https://github.com/angular/protractor/issues/3902

Custom frameworks might not make this change but it'll be fine.  It'll only be a
problem in edge cases and they probably weren't returning the right promise
before anyway.


juliemr:
LGTM once Travis is happy.

Issue #3978, opened 2017-01-16
Protractor CLI rejects `cucumberOpts` as invalid


jan-molak:
Hello there!

It looks like the whitelist defined in the [`cli.ts`](https://github.com/angular/protractor/blob/cd9182618c768ccd2375e187fd85396853bd10de/lib/cli.ts#L35) rejects the `cucumberOpts` key which the [`protractor-cucumber-framework` uses](https://github.com/mattfritz/protractor-cucumber-framework).

- Node Version: `6.9.4`
- Protractor Version: `5.0.0`

Running protractor with the below config file results in this output:

```
Error: Found extra flags: cucumberOpts
```

```javascript
exports.config = {
    seleniumServerJar: path.resolve(node_modules, 'protractor/node_modules/webdriver-manager/selenium/selenium-server-standalone-2.53.1.jar'),
        
    framework: 'custom',
    frameworkPath: require.resolve('protractor-cucumber-framework'),
    specs: [ 'features/**/*.feature' ],
    cucumberOpts: {
        require:    [
            'features/**/*.ts'
        ],
        format:     'pretty',
        compiler:   'ts:ts-node/register',
    }
};
```

I guess there are several ways to resolve this issue:

1. run the protractor/cucumber combo with a `--disableChecks` switch (that's more of a workaround, though)
2. change the whitelist to include `cucumberOpts`, which allows people to run their cucumber tests same as they used to, but requires the Protractor Team to amend the whitelist whenever a new framework comes along or changes the way it's configured (which breaks "half of SOLID" ;-))
3. change the CLI to only perform the checks on keys recognised and supported by Protractor, ignoring any additional keys, which to me seems the most sensible approach.

Looking forward to hearing your thoughts!
Jan

mgiambalvo:
The goal of the flag checks is to help catch when people misspell a flag name, so ignoring additional keys would defeat the purpose. However, there should be a way for protractor-cucumber-framework to add to the whitelist, we can work on that.

jan-molak:
I guess there could be an API to allow plugin authors to register additional commands perhaps?

igniteram:
@jan-molak This is strange you should not face this issue. Are you sure you are not passing `cucumberOpts` flag through command line?

[...]

PR #3994, opened 2017-01-21
fix(cli): Allow frameworks to specify extra command line flags

mgiambalvo:
Fix for #3978.

Our initial plan to allow setting --disableChecks with an environment variable is insufficient, since the custom framework isn't even require()'d until after the config is parsed. This moves the unknown flag check into the runner, and gives frameworks a way to specify extra flags they accept.

### DIFF
--- a/lib/runner.ts
+++ b/lib/runner.ts
@@ -6,6 +6,7 @@ import * as util from 'util';
 import {ProtractorBrowser} from './browser';
 import {Config} from './config';
 import {buildDriverProvider, DriverProvider} from './driverProviders';
+import {ConfigError} from './exitCodes';
 import {Logger} from './logger';
 import {Plugins} from './plugins';
 import {protractor} from './ptor';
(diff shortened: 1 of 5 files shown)


lib/runner.ts
```diff
@@ -79,10 +80,21 @@ export class Runner extends EventEmitter {
...
    * @public
+   * @param {string[]=} An optional list of command line arguments the framework will accept.
    * @return {q.Promise} A promise that will resolve when the test preparers
    *     are finished.
    */
-  runTestPreparer(): q.Promise<any> {
+  runTestPreparer(extraFlags?: string[]): q.Promise<any> {
+    let unknownFlags = this.config_.unknownFlags_ || [];
+    if (extraFlags) {
+      unknownFlags = unknownFlags.filter((f) => extraFlags.indexOf(f) != 0);
```
cnishina: `!== -1`

mgiambalvo: Done

lib/runner.ts
```diff
@@ -79,10 +80,21 @@ export class Runner extends EventEmitter {
...
    * @public
+   * @param {string[]=} An optional list of command line arguments the framework will accept.
    * @return {q.Promise} A promise that will resolve when the test preparers
    *     are finished.
    */
-  runTestPreparer(): q.Promise<any> {
+  runTestPreparer(extraFlags?: string[]): q.Promise<any> {
+    let unknownFlags = this.config_.unknownFlags_ || [];
+    if (extraFlags) {
+      unknownFlags = unknownFlags.filter((f) => extraFlags.indexOf(f) === -1);
```
cnishina: Yup I stand corrected. This should be `===`. Filter flags that do not exist in extra flags.

segrey:
Hi guys!
Unfortunately, it won't help IntelliJ protractor integration which passes `--intellijOriginalConfigFile=/path/to/protractor.conf`. Even it already passes `--disableChecks` option, previously released version don't have it. What do you think about disabling checking for unknown flags in 5.1 completely, and restore this functionality in the next version? That would give us time to release IDE version with `--disableChecks` option. Thanks in advance!

mgiambalvo:
Unknown flags will log a warning message in 5.1 instead of quitting with an error. We're hoping to get 5.1 out this Monday.

segrey:
Thanks for the prompt turnaround! Warning is a good response to unknown flags.

anatoliyarkhipov:
Can you please provide an example of using the `unknownFlags_` option? We use protractor `5.1.2` and in the config file I tried to define custom flags:

```
  unknownFlags_: [
    'logX',
    '--logX',
  ],
```

But I still receive the warning:

```
$ npm run spec -- --logX

> protractor ./specs/conf.js "--logX"

[13:47:28] W/runner - Ignoring unknown extra flags: logX. This will be an error in future versions, please use
 --disableChecks flag to disable the  Protractor CLI flag checks.
```

Did I miss something? 🤔

Issue #3947, opened 2017-01-09
Replace browser.rootEl with browser.setAngularRoot()


mgiambalvo:
Similar to replacing `browser.ignoreSynchronization` with `browser.waitForAngularEnabled()`, we should change setting the root element from a property to a method call.

mgiambalvo:
Also, make sure we do the right thing in this case:

```
browser.setAngularRoot('x');
browser.get('http://something.com').then(() => {
   element(by.css('button')).click();
});
browser.setAngularRoot('y');
element(by.css('other_button')).click();
```

sjelin:
Same with `useAllAngularTestabilities`

PR #3996, opened 2017-01-23
Replace rootEl with browser.setAngularRoot()

mgiambalvo:
Implements #3947. This is the other side of https://github.com/angular/blocking-proxy/pull/21

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -189,18 +189,13 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
    * 'body' but if your ng-app is on a subsection of the page it may be
    * a subelement.
    *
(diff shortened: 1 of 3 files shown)


lib/browser.ts
```diff
@@ -189,9 +189,39 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
...
+   * Set the css selector for an element on which to find Angular. This is usually
+   * 'body' but if your ng-app is on a subsection of the page it may be
+   * a subelement.
+   *
+   * The change will be made within WebDriver's control flow, so that commands after
+   * this method is called use the new app root.
+   *
+   * @param {string} The new selector.
+   */
+  setAngularAppRoot(value: string) {
```
mgiambalvo: Good point, done

PR #3998, opened 2017-01-23
chore(deps): update dependencies and related docs

sjelin:

PR #4000, opened 2017-01-24
chore(types): types for extended webdriver functions

sjelin:
This really belonged in https://github.com/angular/protractor/pull/3998 but I guess we all make mistakes


juliemr:
this LGTM. 

In a follow-up - this would be a GREAT place in the code to have a comment about why on earth we're doing what we're doing.

sjelin:
@juliemr will do that in another PR

PR #4006, opened 2017-01-24
Issue 3578 expected conditions

tilmanpotthof:
Add error handler `falseIfMissing` to all expected conditions that depend
on the presence of an element.

Expected conditions check the presence of an element before other checks,
but when an element is removed exactly in the moment after the `isPresent`
and before `isDisplayed` in `visibilityOf` the condition used to fail.

This solution does not handle missing elements in (`isEnable`, `isDisplayed`, `isSelected`) and focused only on expected conditions (see
#3972)

This problem was also referenced in
#3578.

## Tests

Add test cases to reproduce the missing element race conditions possible in
expected condition methods `visibilityOf`, `textToBePresentInElement`,
`textToBePresentInValue` and `elementToBeClickable`.

See failing tests before implementation: https://circleci.com/gh/tilmanpotthof/protractor/14

### DIFF
--- a/lib/element.ts
+++ b/lib/element.ts
@@ -4,6 +4,7 @@ import {ElementHelper, ProtractorBrowser} from './browser';
 import {IError} from './exitCodes';
 import {Locator} from './locators';
 import {Logger} from './logger';
+import {falseIfMissing} from './util';
 
 let clientSideScripts = require('./clientsidescripts');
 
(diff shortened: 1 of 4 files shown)


mgiambalvo:
Thanks so much for this! Especially for including test cases!

tilmanpotthof:
@mgiambalvo Thanks for merging!

Issue #3896, opened 2016-12-30
Restarting forked browsers is broken


sjelin:
Let's say you do the following:

```ts
let b2 = browser. forkNewDriverInstance()
b2.restart()
```

Because `restart` assumes it's being called against the global browser instance, the global `browser` instance will be overwritten with the restarted `b2` instance, and `b2` will still be pointing to original (now closed) forked instance.

I propose we have `restart` return the new instance.  We can still overwrite the global instance when appropriate too (just add an `if` statement [here](https://github.com/angular/protractor/blob/4.0.14/lib/runner.ts#L255)).  So now the code will be:

```ts
let b2 = browser. forkNewDriverInstance()
b2 = b2.restart()
```

Whatever we end up doing to `forkNewDriverInstance` in #3892 we'll have to also do to `restart` for `async`/`await` users.

sjelin:
Fixed by https://github.com/sjelin/protractor/commit/b77cb928301fbe4f77ffcdcace424a490581416e

PR #4008, opened 2017-01-25
feat(restart): `browser.restart` should return a promise

sjelin:
Also allows `browser.restart` to work when the control flow is disabled, and
fixes it for forked browsers.

Closes https://github.com/angular/protractor/issues/3899 and
https://github.com/angular/protractor/issues/3896

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -236,11 +236,16 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
   params: any;
 
   /**
+   * Resolved when the browser is ready for use.  Resolves to the browser, so
+   * you can do:
+   *
+   *   forkedBrowser = await browser.forkNewDriverInstance().ready;
+   *
    * Set by the runner.
    *
-   * @type {q.Promise} Done when the new browser is ready for use
+   * @type {webdriver.promise.Promise.<ProtractorBrowser>}
    */
-  ready: wdpromise.Promise<any>;
+  ready: wdpromise.Promise<ProtractorBrowser>;
 
   /*
    * Set by the runner.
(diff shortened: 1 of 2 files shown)


lib/browser.ts
```diff
@@ -418,7 +423,15 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
...
-   * Set by the runner.
+   * @example
+   * // Running with control flow enabled
+   * var fork = browser.forkNewDriverInstance();
+   * fork.get('page1'); // 'page1' gotten by forked browser
+   *
+   * @example
+   * // Running with control flow disabled
+   * var forked = await browser.forkNewDriverInstance().ready;
+   * await fork.get('page1'); // 'page1' gotten by forked browser
```
juliemr: This should be `forked.get`

lib/browser.ts
```diff
@@ -432,11 +445,84 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
...
+   * await fork.get('page1');
+   * fork = await fork.restart();
+   * await fork.get('page2'); // 'page2' gotten by restarted fork
+   *
+   * @example
+   * // Unexpected behavior can occur if you save references to the global `browser`
+   * var savedBrowser = browser;
+   * browser.restart().then(function() {
+   *   console.log(browser === savedBrowser); // false
+   * });
```
juliemr: I think this scenario is better captured by this example:

```
var savedBrowser = browser;
browser.get('foo').then(function() {
  console.log(browser === savedBrowser); // false
});
browser.restart();
```

lib/runner.ts
```diff
@@ -113,7 +113,7 @@ export class Runner extends EventEmitter {
     this.frameworkUsesAfterEach = true;
     if (this.config_.restartBrowserBetweenTests) {
       // TODO(sjelin): remove the `|| q()` once `restart()` returns a promise
-      this.restartPromise = this.restartPromise || protractor.browser.restart() || q();
+      this.restartPromise = this.restartPromise || q(protractor.browser.restart());
```
juliemr: Remove the TODO above.

sjelin:
@juliemr all comments addressed

PR #4012, opened 2017-01-25
chore(ci): Switch to Firefox stable now that 51 is out

heathkit:


spec/ciFullConf.js
```diff
@@ -31,7 +31,7 @@ exports.config = {
     'tunnel-identifier': process.env.TRAVIS_JOB_NUMBER,
     'build': process.env.TRAVIS_BUILD_NUMBER,
     'name': 'Protractor suite tests',
-    'version': '47',
```
juliemr: Shouldn't this be 51 now?

mgiambalvo: This is still a work in progress. Saucelabs doesn't have 51 yet, and the beta appears to be broken now. I'll check back tomorrow.

aperdomob: what happen with this CR @juliemr @mgiambalvo

mgiambalvo: Saucelabs doesn't have FF 51 deployed yet, and last I checked we had issues with 50. I'll give it another look soon.

heathkit:
haha, firefox. Oh well, let's try this again later.

Issue #3985, opened 2017-01-18
Fix "mochawd" to work without the control flow


sjelin:
The equivalent of `jasminewd` for mocha is in `lib/frameworks/mocha`.  It needs to support the control flow being turned off

sjelin:
The core issue is that in https://github.com/angular/protractor/pull/3781 we hard-code `selenium-webdriver`'s implementation into Protractor.  The reason why we do this is because we want to overwrite functions like `global.it`, but as of `3.0.0-beta-4` `selenium-webdriver`'s wrapped `it` queries for `global.it` at invocation time.  So by overwriting `global.it` with`selenium-webdriver`'s wrapped version, we create a circular reference.  We could do one of two things:

1) Fix/maintain the code we've brought over
2) Change our wrappers to restore the global function when `selenium-webdriver` queries for it
   * Implementation 1
     ```js
     // This implementation could cause problems if mocha queries for global functions
     // before this code is run (as they did in their old implementation), but that 
     // could be solved with slightly more complex code
     var unwrapCounters = {
       it: 0
     };
     global.it = function() {
       if (unwrapCounters.it > 0) {
         unwrapCounters.it--;
         return mocha.it.apply(this, arguments);
       } else {
         unwrapCounters.it++;
         seleniumMochaWrappedFns.it.apply(this, arguments);
       }
     }
     ```
   * Implementation 2
     ```js
     var unwrappedFns = {
       it: global.it
     }
     global.it = function() {
       if (stackTraceShowsSeleniumCalledUs()) {
         return unwrappedFns.it.apply(this, arguments;
       } else {
         seleniumMochaWrappedFns.it.apply(this, arguments);
       }
     }
     ```

mgiambalvo:
+1 for the first approach. Seems like the least maintenance burden, and if we don't currently wrap nested calls why add it?

juliemr:
I'm down with approach 1 too.

PR #4013, opened 2017-01-25
chore(mocha): refactor to use `selenium-webdriver`'s mocha adapters

sjelin:
Closes https://github.com/angular/protractor/issues/3985


juliemr:
Looks like circle is legitimately unhappy with
```
  1) protractor library with async tests "after all" hook: verify mocha waited:
     TypeError: Cannot read property 'call' of undefined
```

sjelin:
@juliemr sorry about that, circle is fixed now

Issue #3900, opened 2016-12-30
Enhance `browser.ready`


sjelin:
In a world without the WebDriver Control Flow, we're going to need `browser.ready` to be better:

* Initialize to `driver.getSession()`, since that will wait for the browser session to start
* Update with the result of the [`getCapabilities()` call](https://github.com/angular/protractor/blob/4.0.14/lib/browser.ts#L309) call in the `ProtractorBrowser` constructor.
* Update with the [`setScriptTimeout()` call](https://github.com/angular/protractor/blob/4.0.14/lib/runner.ts#L231) in `runner.createBrowser`, instead of overwriting
* Update with the [`browser_.driver.getCurrentUrl()` call](https://github.com/angular/protractor/blob/4.0.14/lib/runner.ts#L243) in `forkNewDriverInstance`, and also return the result of `newBrowser.get(url)` in the subsequent `then` block

sjelin:
The `getCapabilities()` stuff should probably be moved inside the runner

PR #4015, opened 2017-01-26
feat(browser.ready): make `browser.ready` wait for all async setup work

sjelin:
Closes https://github.com/angular/protractor/issues/3900

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -348,7 +348,6 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
     this.ignoreSynchronization = false;
     this.getPageTimeout = DEFAULT_GET_PAGE_TIMEOUT;
     this.params = {};
-    this.ready = null;
     this.plugins_ = new Plugins({});
     this.resetUrl = DEFAULT_RESET_URL;
     this.debugHelper = new DebugHelper(this);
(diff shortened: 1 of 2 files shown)


lib/browser.ts
```diff
@@ -365,17 +364,21 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
...
-      // reset URL for Protractor.
-      // Safari accepts data urls, but SafariDriver fails after one is used.
-      // PhantomJS produces a "Detected a page unload event" if we use data urls
-      let browserName = caps.get('browserName');
-      if (browserName === 'internet explorer' || browserName === 'safari' ||
-          browserName === 'phantomjs' || browserName === 'MicrosoftEdge') {
-        this.resetUrl = 'about:blank';
-      }
-    });
+    this.ready = this.driver.controlFlow()
```
juliemr: Why do we need the wrapping in `execute` here?

lib/runner.ts
```diff
@@ -258,9 +260,13 @@ export class Runner extends EventEmitter {
         newBrowser.mockModules_ = browser_.mockModules_;
       }
       if (opt_useSameUrl) {
-        browser_.driver.getCurrentUrl().then((url: string) => {
-          newBrowser.get(url);
-        });
+        newBrowser.ready = wdpromise.all([newBrowser.ready, browser_.driver.getCurrentUrl()])
```
juliemr: This seems simpler as:

```ts
newBrowser.ready = newBrowser.ready().then(() => {
  return browser_.driver.getCurrentUrl();
}).then((url) => {
  return newBrowser.get(url);
}).then(() => {
  return newBrowser;
});
```
instead of messing with the promise.all.

lib/browser.ts
sjelin: `this.driver.getSession()` is not necessarily a promise which is on the control flow, so this is needed to ensure that the other stuff we stick onto `browser.ready` is on the control flow

lib/runner.ts
sjelin: done

PR #4016, opened 2017-01-26
chore(types): Inherit from webdriver.WebDriver types

sjelin:
I decided to address this comment:

https://github.com/angular/protractor/pull/4000#issuecomment-275281836

While doing do I decided to take on this TODO:

https://github.com/angular/protractor/blob/ccf02ab5f1070f0d7124318dc0099252f3c747e2/lib/browser.ts#L38

One possible issue here is that `ProtractorBrowser` only copies over methods, so it doesn't actually
implement `WebDriver`'s interface.  This isn't a problem right now, since `WebDriver`'s interface
only has functions on it.  But it could be a problem in the future.

This is also weird in that `browser instanceof webdriver.WebDriver` is false but TypeScript probably thinks otherwise.  An alternative option would be to do: 

```ts
export class AbstractWebDriver {
  actions: typeof WebDriver.prototype.actions;
  call: typeof WebDriver.prototype.call;
  ...
}
```

Which would accomplish basically what the TODO suggested, but without ever actually inheriting from anyone

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -1,5 +1,5 @@
 import {BPClient} from 'blocking-proxy';
-import {ActionSequence, By, Capabilities, Command as WdCommand, FileDetector, ICommandName, Options, promise as wdpromise, Session, TargetLocator, TouchSequence, until, WebDriver, WebElement} from 'selenium-webdriver';
+import {ActionSequence, By, Capabilities, Command as WdCommand, FileDetector, ICommandName, Options, promise as wdpromise, Session, TargetLocator, TouchSequence, until, WebDriver, WebElement, WebElementPromise} from 'selenium-webdriver';
 import * as url from 'url';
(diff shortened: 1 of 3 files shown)


lib/browser.ts
```diff
@@ -811,7 +738,7 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
    * @returns {!webdriver.promise.Promise} A promise that will be resolved to
    *      the located {@link webdriver.WebElement}.
    */
-  findElement(locator: Locator): WebElement {
+  findElement(locator: Locator): WebElementPromise {
```
sjelin: This should have been `WebElementPromise` the whole time

cnishina: We should fix the JSDoc `@returns` to reflect the same return type.

sjelin: Done

PR #4017, opened 2017-01-26
feat(browser): chain promises in `browser.get`

sjelin:
Part of angular#3904

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -890,131 +890,139 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
   get(destination: string, timeout = this.getPageTimeout) {
     destination = this.baseUrl.indexOf('file://') === 0 ? this.baseUrl + destination :
                                                           url.resolve(this.baseUrl, destination);
+
+    if (this.ignoreSynchronization) {
+      return this.driver.get(destination)
+          .then(() => this.driver.controlFlow().execute(() => this.plugins_.onPageLoad()))
+          .then(() => null);
+    }
+
     let msg = (str: string) => {
       return 'Protractor.get(' + destination + ') - ' + str;
     };
 
-    if (this.bpClient) {
-      this.driver.controlFlow().execute(() => {
-        return this.bpClient.setSynchronization(false);
-      });
-    }
-
-    if (this.ignoreSynchronization) {
-      this.driver.get(destination);
-      return this.driver.controlFlow().execute(() => this.plugins_.onPageLoad()).then(() => {});
-    }
-
-    let deferred = wdpromise.defer<void>();
-
-    this.driver.get(this.resetUrl).then(null, deferred.reject);
-    this.executeScriptWithDescription(
-            'window.name = "' + DEFER_LABEL + '" + window.name;' +
-                'window.location.replace("' + destination + '");',
-            msg('reset url'))
-        .then(null, deferred.reject);
-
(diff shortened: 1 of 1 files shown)


lib/browser.ts
```diff
@@ -837,79 +848,85 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
       return 'Protractor.get(' + destination + ') - ' + str;
     };
 
+    let retPromise: wdpromise.Promise<any>;
+    let then = (fn: () => wdpromise.Promise<any>) => {
```
mgiambalvo: This is clever, but naming it `then` is a little confusing for me, since it makes it hard to distinguish chaining from calling `.then` on a promise at a glance. Maybe just `chainThenables`, `andThen`, or something like that?

Also, why do we need this? Why wouldn't regular promise chaining work?

mgiambalvo:
Looking at the log for the test failure, it's not the alert timeout issue. It looks like disabling synchronization in get() when using blocking proxy is broken.

lib/browser.ts
```diff
@@ -837,79 +848,85 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
...
+      if (retPromise) {
+        return (retPromise = retPromise.then(fn));
+      } else {
+        return (retPromise = fn());
+      }
+    };
+
     if (this.bpClient) {
```
mgiambalvo: Move this below the this.ignoreSynchronization check, I think that'll fix the test on Circle.

lib/browser.ts
```diff
@@ -837,79 +848,85 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
...
         return this.bpClient.setSynchronization(false);
-      });
+      }));
     }
 
     if (this.ignoreSynchronization) {
-      this.driver.get(destination);
-      return this.driver.controlFlow().execute(() => this.plugins_.onPageLoad()).then(() => {});
+      then(() => this.driver.get(destination));
+      return then(() => this.driver.controlFlow().execute(() => this.plugins_.onPageLoad()))
```
mgiambalvo: Move the bpClient.setSynchronization call after this test. I'm not sure, but I think Circle is failing because waitForAngular gets turned off in BP but never gets restored if get() returns from ehre.

lib/browser.ts
sjelin: Regular promise chain would work, but it's tricky, since the first promise could be one of three different ones, and you want to prematurely exit if `ignoreSynchronization` is on.  In the light of day though, if I move the `ignoreSynchronization` case up, and start the synchronized case with a blank execute, I should be able to make promise chaining work fine.

lib/browser.ts
sjelin: Based on my debugging last night, I don't think that's the problem.  But I'll move it anyway

sjelin:
Since it appears to be only the `browser.get` chaining that's causing problems, I've split everything else into a separate PR: https://github.com/angular/protractor/pull/4021

mgiambalvo:
Looking through the tcpdump from the earlier failure, and it looks like bp.setSynchronization(true) is getting called before addBaseMockModules. BP then runs the wait script, which fails because angular isn't on the page. Not totally sure what's going on, but it is looking like there was some kind of ordering issue earlier.

sjelin:
@mgiambalvo @juliemr it appears the problem was solved by rebasing, whatever it was

PR #4028, opened 2017-01-27
fix(cli): Make unknown flag check a warning instead of an error

mgiambalvo:


lib/runner.ts
```diff
@@ -93,10 +93,7 @@ export class Runner extends EventEmitter {
       unknownFlags = unknownFlags.filter((f) => extraFlags.indexOf(f) === -1);
     }
     if (unknownFlags.length > 0 && !this.config_.disableChecks) {
-      throw new ConfigError(
-          logger,
-          'Found extra flags: ' + unknownFlags.join(', ') +
-              ', please use --disableChecks flag to disable the Protractor CLI flag checks. ');
+      logger.warn('Ignoring unknown extra flags: ' + unknownFlags.join(', '));
```
sjelin: But a TODO and make an issue about reverting this change in 6.0

mgiambalvo: Done

PR #4029, opened 2017-01-28
fix(element chaining): make element chaining work when the control fl…

sjelin:
…ow is disabled

Also added some tests to `spec/ts/noCF/smoke_spec.ts` double checking that the control flow is off

@juliemr I know we talked offline and said it'd be best not to mess with `getWebElement` and have  chaining only work on value.  But given functions like `isElementPresent` and `equals` and `count`, it seems like there are going to be a lot of edge cases on that route and no clear dividing line on which functions should work and which shouldn't.  This implementation is relatively clean, and should be very rigorously tested by our suite.  So let me know.

### DIFF
--- a/lib/element.ts
+++ b/lib/element.ts
@@ -482,18 +482,30 @@ export class ElementArrayFinder extends WebdriverWebElement {
     let callerError = new Error();
     let actionResults = this.getWebElements()
                             .then((arr: any) => wdpromise.all(arr.map(actionFn)))
-                            .then(null, (e: IError | string) => {
-                              let noSuchErr: any;
-                              if (e instanceof Error) {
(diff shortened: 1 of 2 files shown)


juliemr:
This looks good to me and the tests are happy, so I'm good. Can you remind me which test was flaky without this? Is there a way we can force that test to always fail w/o this change, to improve testing?

sjelin:
@juliemr The old flaky test was `"should chain element actions"`.  The new test `"should run chained element actions in sequence"` should always fail without this change.

sjelin:
(you may have missed the new test because clang apparently reformatted the whole file)

juliemr:
k, LGTM

PR #4036, opened 2017-01-31
fix(plugins): do not force ManagedPromise in plugins.ts

sjelin:


juliemr:
Approved, with requisite wishing we actually had a test to demonstrate the problem.

sjelin:
@juliemr test added, will merge when CI is done

sjelin:
hahahaha that was three hours ago never decide to do a bit of debugging before you go to sleep 😂 (also fun fact: node doesn't warn you about circular references it just randomly gives your empty objects and lets you figure out what's gone wrong)

PR #4040, opened 2017-01-31
chore(doc-gen): fix the doc gen script

sjelin:
I never ran into the annotation problem @juliemr talked about, so please test this on your local machines

### DIFF
--- a/scripts/generate-docs.sh
+++ b/scripts/generate-docs.sh
@@ -1,8 +1,5 @@
 #!/bin/sh
 cd "$( dirname "${BASH_SOURCE[0]}" )/.."
-cd testapp && npm install
-
-cd ../website
 
 # Check that directory is clean
 if [ $(git status --porcelain | wc -l) != "0" ]; then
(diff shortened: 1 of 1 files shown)


juliemr:
I get `sh: gulp: command not found`. Confused as to where that's coming from - I'm looking into it...

scripts/generate-docs.sh
```diff
@@ -27,8 +24,56 @@ if [ $? -ne 0 ]; then
...
+  echo -e "\033[0;31m" 1>&2 # Red
+  echo "Couldn't install testapp"
+  echo -e "\033[0m" 1>&2 # Normal Color
+  exit 1
+fi
 
-# Generate files
+echo "Building the website..."
+cd website
 npm run build
```
juliemr: This should also run `npm install` here.

juliemr:
I also get a lot of error messages but they don't seem to actually stop docs from being generated. 

Notably:

```
(node) v8::ObjectTemplate::Set() with non-primitive values is deprecated
(node) and will stop working in the next major release.
```

I think this is due to old versions of some sub dependencies, probably from dgeni.

scripts/generate-docs.sh
sjelin: Sorry about that.  Looks like my clean from a couple lines up wasn't as thorough as I though, so I didn't catch this.  Fixed.

sjelin:
@juliemr those definitely don't sound like a problem with the script itself.

juliemr:
This looks pretty good to me and runs without completely falling over.

PR #4042, opened 2017-01-31
deps(update): update webdriver-manager to ^12.0.1

cnishina:


juliemr:
We're going to need a commit message here that will be useful for users who might be confused about the new defaults. Something like:

Running `webdriver-manager update` will now by default grab the latest versions of all binaries (standalone, chromedriver, etc). You can continue to pin to specific versions using `blah blah`. As of this release, the latest versions are `blah blah` and that causes these changes with firefox.

juliemr:
Can you expand in the commit message about what works for firefox? e.g. note that 3.0.1 does NOT?

PR #4048, opened 2017-02-01
chore(docgen): fix docgen for 5.1

sjelin:
Also added the ability to run `generate-docs.sh` against a custom
branch/tag/commit by doing `./scripts/generate-docs.sh branch_name`.  Still
defaults to the version from package.json though.

Also updated release.md to tell people to check that doc generation/the website
is working BEFORE doing a release.  Needing a change in the codebase is more
likely now that we're compiling down to es5 only this one time, and ideally
any changes to the codebase would happen before release.

Also including a minor change in the codebase where we wrap a promise from
blocking proxy in a webdriver promise.  This probably should have been done the
whole time, but is unlikely to matter either way at runtime, since blocking
proxy wouldn't be returning a webdriver promise regardless.  Mostly did I did
this to fix typechecking.

### DIFF
--- a/scripts/generate-docs.sh
+++ b/scripts/generate-docs.sh
@@ -1,6 +1,12 @@
 #!/bin/sh
 cd "$( dirname "${BASH_SOURCE[0]}" )/.."
 
+# Check number of parameters
+if [ "$#" -gt 1 ]; then
+  echo "Usage: ./scripts/generate-docs.sh [commit_ref]"
+  exit 1
+fi
+
 # Check that directory is clean
 if [ $(git status --porcelain | wc -l) != "0" ]; then
   echo -e "\033[0;31m" 1>&2 # Red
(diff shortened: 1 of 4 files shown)


juliemr:
Happy with the protractor typing fix, but as a note for the future, I don't think that blocking proxy should return a webdriver promise, since it should't be tied to any particular webdriver bindings.

scripts/generate-docs.sh
```diff
@@ -12,14 +12,22 @@ if [ $(git status --porcelain | wc -l) != "0" ]; then
   exit 1
 fi
 
-echo "Switching to last release..."
-VERSION=$(node scripts/get-version.js)
 EXEC_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+if [ "$#" -eq 0 ]; then
```
juliemr: Because reading bash is hard, can we add a comment here? Maybe:

```
# If a commit ref is passed as a command line option, use that. Otherwise, default to the
# latest tagged version.
```

sjelin:
@juliemr I'll clarify that it would be ideal from Protractor's point of view, but wouldn't make sense for blocking proxy.

tsconfig.json
```diff
@@ -7,13 +7,7 @@
...
     "removeComments": false,
     "noImplicitAny": true,
-    "outDir": "built/",
-    "types": [
-      "jasmine", "jasminewd2", "node",
-      "chalk", "glob", "minimatch",
-      "minimist", "optimist", "q",
-      "selenium-webdriver"
-    ]
+    "outDir": "built/"
```
juliemr: So I assume ts just uses all available typings if there's not an array here?

sjelin: Essentially.  It defaults to everything in `node_modules/@types`

scripts/generate-docs.sh
sjelin: Done (paraphrased slightly to be technically correct).

Also added usage up top

PR #4065, opened 2017-02-04
fix(error message): fix error message `browser.get`

sjelin:


sjelin:
Just gonna merge this in...

Issue #1451, opened 2014-10-16
Protractor logging level


dime832:
Is there any way to change the protractor logging levels? For I would like suppress the following warning. While I realize the warning is there for a reason there are times when I would like to just see the jasmine log messages without the protractor log messages intertwined. Thank You

warning: more than one element found for locator By.className("schedule-box") - you may need to be more specific

mtyeh411:
not sure if this is related, but it'd be nice to suppress all the webdriver INFO messages (eg, 'Executing', 'Done' messages). while helpful, it can be a bit 'noisy'.  i'm sure this is a webdriver preferences settings, just not sure how to configure it via protractor (or at least not seeing where that is documented).

bes89:
:+1: 
actually there is no way to configure the log level. the child process is spawn with some predefined arguments: [see here](https://github.com/angular/protractor/blob/master/bin/webdriver-manager#L280)

DevVersion:
A temporary workaround could be

```js
// Because Protractor runs selenium instances concurrently in the CI and each instance
// will create a bloat of logs we don't want to show any info messages.
// Until protractor#1451 is solved, manually disable all info messages.
require('protractor/built/logger').Logger.prototype.info = () => {};
```

> This snippet will be placed inside of the protractor configuration file.

trevor92:
Is there an ETA on this?

PR #4068, opened 2017-02-05
feat(logger): Add log level configuration (#1451)

bcaudan:


bcaudan:
Is there anything I could do to help this PR to be merged?

lib/logger.ts
```diff
@@ -54,6 +54,8 @@ export class Logger {
   static set(config: Config): void {
     if (config.troubleshoot) {
       Logger.logLevel = LogLevel.DEBUG;
+    } else if (config.logLevel) {
+      Logger.logLevel = (<any>LogLevel)[config.logLevel];
```
NickTomlin: What is `(<any>LogLevel)` accomplishing here? I'm able to compile without it (that said, I'm not a typescript expert, so perhaps I am missing something).

NickTomlin: NVM this is an error in `typescript@2.2.1` but not `2.0.10` (which we have pinned)

bcaudan: I guess you figured out but it is used to convert string to enum, found on [SO](http://stackoverflow.com/questions/17380845/how-to-convert-string-to-enum-in-typescript). 

Would you prefer a comment to make it easier to read or just remove it for now?

bcaudan:
I have updated `cli.ts` accordingly

lib/logger.ts
NickTomlin: I think i'm okay leaving it as is; unless others have strong opinions.

NickTomlin:
I was playing around with the typing for LogLevel a little more (using TS 2.1) and I think i've found a simplified version that has the benefit of explicitly stating the values for `logLevel` in code https://github.com/NickTomlin/protractor/commit/3fe8606651fe17dc4b985adad0d798414b6385eb. This results in an exception in the logger if the value is not valid: 

```
./bin/protractor example/conf.js --logLevel=NOT_VALID
/Users/ntomlin/workspace/protractor/built/logger.js:130
                throw new Error('Log level invalid or undefined');
                ^

Error: Log level invalid or undefined
    at Logger.log_ (/Users/ntomlin/workspace/protractor/built/logger.js:130:23)
    at Logger.debug (/Users/ntomlin/workspace/protractor/built/logger.js:86:14)
    at Object.initFn [as init] (/Users/ntomlin/workspace/protractor/built/launcher.js:98:12)
```

lib/logger.ts
Xotabu4: We might try something like this for log levels :

```typescript
let logLevel: 'ERROR'|'WARN'|'INFO'|'DEBUG'
logLevel = 'OLOLO' //compilation error here
```

bcaudan:
I have updated the PR with @NickTomlin and @Xotabu4 suggestions

bcaudan:
How are we on this PR?

NickTomlin:
@seanmay mind taking a look at this when you have some free time?

igniteram:
can we merge this PR? It is a useful feature!

qiyigg:
Could you please add a test for this change?

alecmerdler:
We would also really like to see this merged.

igniteram:
@bcaudan If you get time can you please add some tests? This PR is pending from long time! 
@alecmerdler Meanwhile you could use [cliptor](https://github.com/igniteram/Cliptor.js) for configuring this log level, It basically wraps the same setting which is implemented in this PR.

alecmerdler:
I've been using [this hacky solution](https://github.com/angular/protractor/issues/1451#issuecomment-269806048) for now. 

Also, I would be open to writing tests for this PR (or opening another with tests) if that would be okay with @bcaudan.

qiyigg:
Sorry for the long delay.
I think this PR was pending for a long time and @alecmerdler or someone else, please feel free to create a new one for it. 
As long as the new one is created, I will close this one, Thanks!

alecmerdler:
@qiyigg This PR doesn't have any merge conflicts, why do we need to open a new one? I am glad to do it but would prefer to give @bcaudan the credit.

bcaudan:
@qiyigg I have added some unit tests for the `Logger`. 
Let me know if you had other tests in mind.

qiyigg:
@bcaudan I already triggered a rebuild, let's see what happens.
The circleCI should be already fixed since last release.
Thanks for adding these tests. My understanding is it would be better to have a integration test, say, have a separate config file which sets loglevel there, and check it during test. But since it is a simple change, I am OK with current test.
I can merge the change as long as the circleCI is passed.

alecmerdler:
@qiyigg Will there be a new release soon with these changes?

qiyigg:
@alecmerdler to be honest, I don't have enough bandwidth recently, I'll try to make it this week, but no guarantee.

PR #4073, opened 2017-02-07
chore(release): Version bump and changelog for release

mgiambalvo:


package.json
```diff
@@ -15,7 +15,7 @@
     "@types/node": "^6.0.46",
     "@types/q": "^0.0.32",
     "@types/selenium-webdriver": "~2.53.39",
-    "blocking-proxy": "0.0.5",
+    "blocking-proxy": "^0.0.5",
```
juliemr: I'm a bit worried about allowing `^` since this is pre version 1 and versioning is nebulous -can we leave it precisely pinned?

mgiambalvo: Yeah, I was hoping to be able to get out fixes without having to publish a new Protractor. But you're right, with things as they are it's probably better to just leave pinned.

wswebcreation:
@mgiambalvo,

Maybe a stupid question, but protractor uses semver and the commits are features, shouldn't protractor be updated to 5.2.0 in stead of a fix release?

Tnx in advance

mgiambalvo:
The main purpose of this release was to undo our accidental dependency on TS 2.1 (which was an unintentional breaking change) and fix support for TS 2.0. There were a couple tiny, low risk features that also made it in, so I left it at 5.1.1. Maybe you're right though, it could have easily been 5.2.0

wswebcreation:
Tnx for explaining, it's clear to me.

PR #4097, opened 2017-02-17
chore(docs): getCssValue takes param, returns RGB instead of hex

olore:
I am using v5.1.1, not sure what which point this changed

### DIFF
--- a/lib/selenium-webdriver/webdriver.js
+++ b/lib/selenium-webdriver/webdriver.js
@@ -523,18 +523,18 @@ webdriver.WebElement.prototype.getTagName = function() {};
 /**
  * Gets the computed style of an element. If the element inherits the named
  * style from its parent, the parent will be queried for its value. Where
- * possible, color values will be converted to their hex representation (e.g.
- * #00ff00 instead of rgb(0, 255, 0)).
+ * possible, color values will be converted to their rgb representation (e.g.
+ * rgba(247, 247, 247, 1) instead of #f7f7f7).
  *
  * _Warning:_ the value returned will be as the browser interprets it, so
  * it may be tricky to form a proper assertion.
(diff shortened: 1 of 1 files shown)


Xotabu4:
Firefox returns HEX i think, Chrome rgba (i might be wrong). Haven't tested this in Edge/IE. 

I think it is better to just add message, that returning value depends on browser

olore:
Thanks @Xotabu4, You are correct, I was testing in Chrome. I will try to confirm Firefox, but I don't think I can do Edge/IE.

[According to this](https://github.com/SeleniumHQ/selenium/blob/c181d392fb04c93f88be2549e2961a883221b6fb/java/client/src/org/openqa/selenium/htmlunit/HtmlUnitWebElement.java#L882
), it looks  like it **should** always return RGB.  This change was made in [June of 2015](https://github.com/SeleniumHQ/selenium/commit/c181d392fb04c93f88be2549e2961a883221b6fb#diff-f444f778bbe28cb2c8acc18f83265282R890)

olore:
Fun fact - Chrome and Firefox both return RGB, at least for `background-color`, but it isn't the same:

```
  Firefox 50.0.2: rgb(247, 247, 247)
  Chrome 56.0:    rgba(247, 247, 247, 1)
```

NickTomlin:
I'm in agreement with @Xotabu4 here; we should document this as differing between browsers but not actually change from Hex -> RGB. Perhaps just expanding the existing warning section to surface and document some of these issues?

olore:
@NickTomlin that sounds fair. My main motivation for this was to update `getCssValue` function to show that it takes a param. The example was more of a collateral change. I like the idea of providing more detail. I'll make an update this week.

Issue #3790, opened 2016-12-03
Replace q with native promises


thorn0:
`selenium-webdriver` depends on native promises anyway, so why keep using the q library?
Would you be interested in a PR that replaces `q` with native promises?

sjelin:
We compile to es5, so this would actually just become whatever typescript generates.  I'd be fine with it though

cnishina:
If someone decides to do this, they should use the `beta` branch:

- setup for ES6 support and tests in CI against node 6+
- has selenium-webdriver 3 dependency which uses native promises

thorn0:
Even selenium-webdriver 2.x relies on native promises. It uses them to shedule microtasks.

sjelin:
Good call.  And it appears I was wrong that typescript compiles out `Promise`.  You'll have to install `@types/es6-promise`, but that's no big deal.

NickTomlin:
Going to go ahead and close this since https://github.com/angular/protractor/commit/4e73d5ed7b0f09c5ebe376a33c7937fb876ee686 resolves it (and it is an internally facing issue so no need to wait until for a release). If anyone feels that that is wrong holler or re-open it.

PR #4098, opened 2017-02-17
Use ES6/TS native Promise in place of Q.

seanmay:


seanmay:
I signed it!

scripts/test/test_util.js
```diff
@@ -177,13 +177,14 @@ var CommandlineTest = function(command) {
             flushAndFail('expecting test max duration: ' +
                 self.expectedMaxTestDuration_ + ', actual: ' + duration);
       }
-    }).fin(function() {
```
NickTomlin: `fin` is equivalent to `Promise.finally`, can we use that instead of `catch`ing?

scripts/test/test_util.js
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 var child_process = require('child_process'),
-    q = require('q'),
+    // q = require('q'),
```
NickTomlin: We should remove this comment

scripts/interactive_tests/interactive_test_util.js
```diff
@@ -140,11 +161,7 @@ exports.InteractiveTest = function(interactiveServerStartCmd, port) {
           }
         };
         return verifyAll(0);
-      }).fin(function() {
```
NickTomlin: `fin` is equivalent to `Promise.finally`, can we use that instead of `catch`ing?

scripts/interactive_tests/interactive_test_util.js
```diff
@@ -79,22 +94,23 @@ var Client = function(port) {
...
     socket.on('data', ondata);
 
     var onerror = function(data) {
-      deferred.reject('Received error: ' + data);
+      deferredReject('Received error: ' + data);
     };
     socket.on('error', onerror);
 
     socket.write(cmd + '\r\n');
-    return deferred.promise.fin(function() {
```
NickTomlin: fin is equivalent to Promise.finally, can we use that instead of catching?

scripts/interactive_tests/interactive_test_util.js
```diff
@@ -62,15 +72,20 @@ var Client = function(port) {
   // Send a command to the server and wait for a response. Return response as a 
   // promise. 
   this.sendCommand = function(cmd) {
-    var deferred = q.defer();
+    let deferredResolve;
```
NickTomlin: How about wrapping everything within a `new Promise` here? It would remove the need to create our own deferred (which is a bit of an anti-pattern IHMO). 

If we must go the deferred route, we should create a helper for this pattern so we aren't repeating ourselves in multiple files.

scripts/interactive_tests/interactive_test_util.js
```diff
@@ -47,11 +52,16 @@ var Client = function(port) {
 
   // Connect to the server. 
   this.connect = function() {
-    var deferred = q.defer();
+    let deferredResolve;
```
NickTomlin: I think this could be simplified to

```
return new Promise((resolve, reject) => {
  socket = net.connect({port: port}, function() {
    resolve();
  })
})
```

scripts/interactive_tests/interactive_test_util.js
```diff
@@ -1,5 +1,5 @@
 var child_process = require('child_process'),
-    q = require('q'),
+    // q = require('q'),
```
NickTomlin: We should remove this comment

scripts/interactive_tests/interactive_test_util.js
```diff
@@ -8,7 +8,12 @@ var TIMEOUT = 10000;
 var Server = function(serverStartCmd, port) {
   // Start protractor and its debugger server as a child process.
   this.start = function() {
-    var deferred = q.defer();
+    let deferredResolve;
```
NickTomlin: Similar to my comments below, we should just return a promise instead of using a deferred here

lib/util.ts
```diff
@@ -33,10 +32,10 @@ export function filterStackTrace(text: string): string {
  * Internal helper for abstraction of polymorphic filenameOrFn properties.
  * @param {object} filenameOrFn The filename or function that we will execute.
  * @param {Array.<object>}} args The args to pass into filenameOrFn.
- * @return {q.Promise} A promise that will resolve when filenameOrFn completes.
+ * @return {Promise} A promise that will resolve when filenameOrFn completes.
  */
 export function runFilenameOrFn_(configDir: string, filenameOrFn: any, args?: any[]): Promise<any> {
-  return Promise((resolvePromise) => {
+  return new Promise((resolvePromise) => {
```
NickTomlin: nit: How about just `resolve` here instead of `resolvePromise` (I know that was the old style but I prefer simplifying it while we are in here)

lib/taskRunner.ts
```diff
@@ -65,7 +64,12 @@ export class TaskRunner extends EventEmitter {
     config.specs = this.task.specs;
 
     if (this.runInFork) {
-      let deferred = q.defer();
+      let deferredResolve: (x?: any) => void;
```
NickTomlin: Similar to comments below, how about just wrapping the block in a promise and returning it

lib/runner.ts
```diff
@@ -492,8 +492,13 @@ export class Runner extends EventEmitter {
           let exitCode = testPassed ? 0 : 1;
           return this.exit_(exitCode);
         })
-        .fin(() => {
```
NickTomlin: Similar to comments below, we should just be able to use `finally` here

lib/plugins.ts
```diff
@@ -508,7 +509,10 @@ export class Plugins {
               .map(
                   pluginObj =>
                       this.safeCallPluginFun(pluginObj, funName, args, promiseType, failReturnVal));
-      return promiseType == PromiseType.Q ? q.all(promises) : webdriver.promise.all(promises);
+      return promiseType == PromiseType.Q ?
```
NickTomlin: I'm a little ignorant here, do we need to differentiate between Q and native promises here? Can't `Promise.all` handle both types?

lib/launcher.ts
```diff
@@ -296,9 +300,13 @@ let initFn = function(configFile: string, additionalConfig: Config) {
               }
               return cleanUpAndExit(exitCode);
             })
-            .done();
+            .catch(err => setTimeout(() => {
```
NickTomlin: This is a tough one because there's no equivalent to `done` on the native side. I think we should either move these into a helper function or talk about returning this deferred so the exception can bubble back up the chain.

lib/launcher.ts
```diff
@@ -243,7 +241,13 @@ let initFn = function(configFile: string, additionalConfig: Config) {
           }
         }
 
-        let deferred = q.defer<any>();  // Resolved when all tasks are completed
+        let deferredResolve: () => void;
```
NickTomlin: Do we ever return this deferred? If not, we should just wrap the following code in a promise

lib/frameworks/mocha.js
```diff
@@ -14,7 +12,12 @@ exports.run = function(runner, specs) {
   // Add hooks for afterEach
   require('./setupAfterEach').setup(runner, specs);
 
-  var deferred = q.defer();
+  let deferredResolve;
```
NickTomlin: Similar to below, I'm 👍 on just returning a promise here and wrapping the below

lib/driverProviders/sauce.ts
```diff
@@ -27,33 +25,37 @@ export class Sauce extends DriverProvider {
...
    * @public
    * @param {Object} update
-   * @return {q.promise} A promise that will resolve when the update is complete.
+   * @return {Promise} A promise that will resolve when the update is complete.
    */
-  updateJob(update: any): q.Promise<any> {
+  updateJob(update: any): Promise<any> {
     let deferredArray = this.drivers_.map((driver: WebDriver) => {
-      let deferred = q.defer();
+      let deferredResolve: () => void;
```
NickTomlin: Similar to below, we should just return a promise that wraps this code instead of using a deferred

lib/driverProviders/local.ts
```diff
@@ -123,7 +122,12 @@ export class Local extends DriverProvider {
 
     this.server_ = new remote.SeleniumServer(this.config_.seleniumServerJar, serverConf);
 
-    let deferred = q.defer();
+    let deferredResolve: (x?: any) => void;
```
NickTomlin: Similar to below, we should just return a promise that wraps this code instead of using a deferred

lib/driverProviders/driverProvider.ts
```diff
@@ -89,42 +88,38 @@ export abstract class DriverProvider {
...
-        .then(
-            () => {
-              deferred.resolve();
-            },
-            () => {
-              deferred.resolve();
-            });
-    return deferred.promise;
+  static quitDrivers(provider: DriverProvider, drivers: WebDriver[]): Promise<void> {
+    let deferredResolve: (x?: any) => void;
```
NickTomlin: Similar to below, we should just return a promise that wraps this code instead of using a deferred

lib/driverProviders/driverProvider.ts
```diff
@@ -89,42 +88,38 @@ export abstract class DriverProvider {
...
+    let deferredReject: (x?: any) => void;
+    let deferred = new Promise<void>((resolve, reject) => {
+      deferredResolve = resolve;
+      deferredReject = reject;
+    });
+    wdpromise.all(drivers.map((driver: WebDriver) => provider.quitDriver(driver)))
+        .then(() => deferredResolve(), () => deferredReject());
```
NickTomlin: Could this just be

```
return wdpromise.all(drivers.map((driver: WebDriver) => provider.quitDriver(driver)))
```

Since it will `then` or `catch` further up the promise chain?

If we have to return a native promise, how about just 

```
return new Promise((resolve, reject) => {
  wdpromise.all(drivers.map((driver: WebDriver) => provider.quitDriver(driver)))
    .then(resolve, reject)
})
```

lib/driverProviders/browserStack.ts
```diff
@@ -25,11 +24,16 @@ export class BrowserStack extends DriverProvider {
...
    * @public
    * @param {Object} update
-   * @return {q.promise} A promise that will resolve when the update is complete.
+   * @return {Promise} A promise that will resolve when the update is complete.
    */
-  updateJob(update: any): q.Promise<any> {
+  updateJob(update: any): Promise<any> {
     let deferredArray = this.drivers_.map((driver: WebDriver) => {
-      let deferred = q.defer();
+      let deferredResolve: (x?: any) => void;
```
NickTomlin: Similar to the comments below, how about:

```typescript
let promises = map((driver: WebDriver) => {
  return new Promise((resolve, reject) => {
    // function body
  })
})

return Promise.all(promises)
```

scripts/test/test_util.js
seanmay: Perhaps you are thinking about Bluebird promises? `finally`, from all I can tell does not appear to be a method in the ES6 spec, as a static nor instance method.

scripts/interactive_tests/interactive_test_util.js
seanmay: I felt horrible writing this pattern; it really makes me cringe every time.
The reasoning is twofold:

1. the current expectation in any of these places where `q.defer()` is used is such that if an exception is thrown, that exception will unwind the stack in the traditional format; in a promise which wraps the whole method, any exception thrown would be propagated through the promise chain. My first take was to do exactly what you suggest, until I started considering how exception throwing is actually used for application flow control in certain parts of Protractor, and how all of that would hang in perpetuity, if something went missing.

2. The reason I had left all of the ugliness at the call-sites rather than in a helper is expressly because it's an awful anti-pattern, actively harmful to the asynchronous data-flow through the app, but in order to make it idiomatic, it requires assessing how all of the pieces fit together, such that a refactor into better concurrency doesn't break the intended functionality. I didn't want to create my own version of `q.defer` to replace `q.defer` because the pattern itself is the problem. I did intend, however, to comment the site of each inside-out promise, along with a few of the infinite-hang sites which result from neglecting to call `deferred.reject()` in any error handling (unrelated to these changes).

lib/driverProviders/driverProvider.ts
seanmay: The WebDriver promises have different characteristics, I believe, and are tied into ControlFlow, which I have yet to deeply familiarize myself with.

As such, propagating them through the system might magically improve everything, or might be a complete non-starter, when considering the concurrency stories.
The additional anonymous functions wrapping the resolve and reject were to guarantee that nothing was passed on, in the instance that the method signature of `quitDriver` never changes, or to equally maintain the behaviour of ignoring the error thrown.

My intent is to look deeper into these cases, but due to the nature of how these things are set up, changing them may lead to rather large refactorings (which I'm not opposed to; I just didn't want to juggle it, while also worrying about changing libraries).

scripts/interactive_tests/interactive_test_util.js
seanmay: Absolutely. Thanks for the catch.

seanmay:
@NickTomlin Thanks for the great review.
I've found, at least locally, that the test suites have been running faster. That, however, might be related to some other factor related to webdriver-manager or the like.

As for types, I have yet to go out of my way to vary the types, tightening or loosening them. The regular flow of the application has been updated to work with native promises. The plug-in system was changed to expect native promises (and any internal code plugged in was changed to expect ES6 promises). Leaving support for Promise.Q (in addition to extending out to native) was due to expectations other people may have about the system. Q has many methods and behaviours which are not supported by the Promises/A+ spec.

lib/util.ts
seanmay: Absolutely. Thanks for the catch.

scripts/test/test_util.js
NickTomlin: Ah I think you are right! I am getting confused with other implementations 😄

lib/plugins.ts
seanmay: It can, but the promise that Promise.all will return is an ES6 promise. If there are no expectations in the future, about retrieving a promise which has support for the large array of Q methods, that sounds like a bigger conversation than an internal tweak. That said, I have no qualms with stripping the vestiges of Q from anything that might be externally-facing, if so desired.

lib/launcher.ts
seanmay: There is also a consideration of whether an exception is to be swallowed or thrown out of the method.

lib/launcher.ts
seanmay: There are a few spots with 

```javascript
      // ...
    }).done();
  }).done();
```
I think that there are a few talks that would be beneficial, in terms of scope of remapping some of the flow of the application.

I will, though, move my `tap` and the like, out to the `util` file as higher-order functions that can be configured in the scope they expect to operate within.

scripts/interactive_tests/interactive_test_util.js
NickTomlin: Thanks for the thorough and thoughtful response!

1. This is totally fair. Agreed that it is sad. I'm fine with moving forward a half-step by implementing things natively and then trying to transition control flow over (if possible) in the future.

2. Interesting; I think I still feel that we would be better served by moving this into a helper because we creating the same thing as `q.defer` regardless. If you feel strongly that the "weight" of having to copy the pattern is discourages others from repeating it then I'm fine with keeping things the way they are. Either way, I like your idea of calling out sites where we may run into issues with control flow.

lib/plugins.ts
NickTomlin: We only use this designation internally, and the expectations for method support would all be internal though right? As long as the plugin author returns a `then`able we should be able to resolve it with `Promise.all` in Runner/Browser. I could have totally missed something here though, so let me know if I haven't gotten it right 😄 

If there's no contract with `Q` i'd love to just do away with this branching and make a WD promise the only special case.

lib/launcher.ts
NickTomlin: 👍 makes sense. I'd love to have a chat about alterations we can make to the flow.

lib/driverProviders/driverProvider.ts
NickTomlin: 👍 I this is fine. I think we should just revisit this later when we talk about changes in the flow

vikerman:
Fixes #3790

juliemr:
Yay! Thanks folks!

Issue #4107, opened 2017-02-23
Add option to create folders to json reports


igniteram:
Enhancement(Feature Request)

* Currently users can create protractor json reports in the current directory but would get an error if they want to create folders.
* There has to be way to generate folders & reports implicitly with protractor.

PR #4109, opened 2017-02-23
Enhancement(Add option to create folders to json reports)

igniteram:
Addresses issue https://github.com/angular/protractor/issues/4107.

TO DO: write tests for it, may be in next PR!


igniteram:
I have seen the same CI errors in other PR's as well, not sure what is causing this!

Issue #3590, opened 2016-09-21
Protractor doesn't wait for beforeLaunch to be resolved when using multiple capabilities


remcohaszing:
Bug report
---

When specifying multiple capabilities in the configuration file, protractor doesn't wait for the promise returned by `beforeLaunch` to be resolved.

- Node Version: `6.6.0`
- Protractor Version: `4.0.8`
- Browser(s): `Chrome / Firefox`
- Operating System and Version `Ubuntu 16.04`
- Your protractor configuration file

```js
'use strict';

/* eslint-env jasmine */

const path = require('path');

require('ts-node/register');
const reporters = require('jasmine-reporters');
const webpack = require('webpack');
const WebpackDevServer = require('webpack-dev-server');


const config = {
  framework: 'jasmine',
  specs: ['app/**/*.e2e-spec.ts'],

  multiCapabilities: [
    {
      browserName: 'chrome'
    },
    {
      browserName: 'firefox'
    }
  ],

  /**
   * These settings remove the need for Selenium.
   */
  chromeDriver: path.resolve(__dirname, 'node_modules/.bin/chromedriver'),
  directConnect: true,

  /**
   * Start a webpack dev server on a random port.
   *
   * The protractor `baseUrl` is set to this port on localhost.
   */
  beforeLaunch: () => {
    return new Promise(resolve => {
      setTimeout(() => {
        const compiler = webpack(require('./webpack.config'));
        const server = new WebpackDevServer(compiler, {
          stats: 'errors-only'
        });
        server.listen(0, 'localhost', () => {
          // `server.listeningApp` will be returned by `server.listen()`
          // in `webpack-dev-server@~2.0.0`
          const address = server.listeningApp.address();
          config.baseUrl = `http://localhost:${address.port}`;
          resolve();
        });
      }, 5000);
    });
  }
};


module.exports = {
  config
};
```

- Steps to reproduce the bug
1. Create a minimal configuration file implementing a `beforeLaunch` function returning a promise, e.g. using a timeout.
2. Add one of:
  - capabilities,
  - multiCapabilities,
  - getMultiCapabilities
3. Notice that when a single capabilities object is added (either as `capabilities` or the only item of `multiCapabilities` or `getMultiCapabilities`), the tests won't run until the promise is resolved.
4. Add another capability
5. Notice that the tests are run before the promise is resolved.

NickTomlin:
[...]

PR #4131, opened 2017-03-03
feat(config): Allow users to modify config in beforeLaunch

NickTomlin:
This is stab at resolving some of the issues in https://github.com/angular/protractor/issues/3590

Essentially, there's no great way for users to modify the config from `beforeLaunch` when running forked through multiCapabilities.

The approach I took:

1. Removes the config file parsing from `taskRunner` and `runnerCli` and instead uses parsed configuration passed in by `launcher`
   - I don't deeply understand either of those classes, so if this is a big faux paux let me know. It seems much cleaner to me to parse once but there may be gotchas.
2. Allow users to merge additional options onto `config` by resolving the promise returned from `beforeLaunch` with an object
     - right now I'm doing a very simple truthy check, I don't know of a way of expressing `if (configFromBeforeLaunch typeof Config)` at runtime :|
        - I'd be happy to more explicitly check for an object here

The following specs were failing for me locally:

- `spec/onPrepare/onPrepare_spec.js`
  - this also fails on `master` locally but could very well be me
- `spec/basic/polling_spec.js`
  - seems environmental as well

### DIFF
--- a/lib/launcher.ts
+++ b/lib/launcher.ts
@@ -108,7 +108,10 @@ let initFn = function(configFile: string, additionalConfig: Config) {
 
   // Run beforeLaunch
   helper.runFilenameOrFn_(config.configDir, config.beforeLaunch)
-      .then(() => {
+      .then((configFromBeforeLaunch) => {
+        if (configFromBeforeLaunch) {
+          configParser.addConfig(configFromBeforeLaunch);
+        }
 
         return new Promise<any>((resolve: Function, reject: Function) => {
                  // 1) If getMultiCapabilities is set, resolve that as
@@ -251,7 +254,7 @@ let initFn = function(configFile: string, additionalConfig: Config) {
         let createNextTaskRunner = () => {
           let task = scheduler.nextTask();
           if (task) {
-            let taskRunner = new TaskRunner(configFile, additionalConfig, task, forkProcess);
+            let taskRunner = new TaskRunner(config, task, forkProcess);
             taskRunner.run()
                 .then((result) => {
                   if (result.exitCode && !result.failedCount) {
--- a/lib/runnerCli.ts
+++ b/lib/runnerCli.ts
@@ -3,7 +3,6 @@
  * requested by the launcher.
  */
 
-import {ConfigParser} from './configParser';
 import {Logger} from './logger';
 import {Runner} from './runner';
 
@@ -15,15 +14,7 @@ process.on('message', (m: any) => {
       if (!m.capabilities) {
         throw new Error('Run message missing capabilities');
       }
-      // Merge in config file options.
-      let configParser = new ConfigParser();
-      if (m.configFile) {
-        configParser.addFileConfig(m.configFile);
-      }
-      if (m.additionalConfig) {
-        configParser.addConfig(m.additionalConfig);
-      }
-      let config = configParser.getConfig();
+      let config = m.config;
       Logger.set(config);
 
       // Grab capabilities to run from launcher.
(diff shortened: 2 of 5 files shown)


lib/launcher.ts
```diff
@@ -251,7 +254,7 @@ let initFn = function(configFile: string, additionalConfig: Config) {
         let createNextTaskRunner = () => {
           let task = scheduler.nextTask();
           if (task) {
-            let taskRunner = new TaskRunner(configFile, additionalConfig, task, forkProcess);
```
mgiambalvo: Does this break setting flags on the command line? It looks like cli.ts passes in `argv` as `additionalConfig`.

I think the reason additionalConfig is passed into the task runner is so that it can be the last thing merged into the config, after the multicapabilities logic and any other files.

I like moving config parsing out of the TaskRunner, though. Maybe add the additionalConfig right before this?

NickTomlin: Great point! I don't think this will prevent options in`argv` from flowing through to the runners : `additionalConfig` is already [merged into config](https://github.com/angular/protractor/blob/master/lib/launcher.ts#L100-L102) and isn't referenced elsewhere in launcher. `multiCapabilities` related mutations happen directly to the `config` object.  Is the use case you are suggesting that `additionalConfig` may have options that we want to override modifications to to config that happen in between the initial parsing and (e.g. our detection of [`capabilities` or `multiCapabilities`](https://github.com/angular/protractor/blob/master/lib/launcher.ts#L143-L156))?

I wasn't able to find a test that verified that contract though :|. Do we have a pattern for testing cli and config related interactions?

NickTomlin:
@mgiambalvo it is tricky. I don't _think_ this would be a breaking change but the complexities make it hard to say. My initial read is that the config is parsed several times and resolves to the same values, but I could totally be missing some environmental quirks that necessitate this.

I'd definitely like to expand some of the test coverage here if possible to make sure. Any thoughts on test cases that could reveal and ensure the interactions we expect?

lib/launcher.ts
mgiambalvo: Right, I missed that. `additionalConfig` might as well be renamed to make it something more clear.

mgiambalvo:
I think the reason config ends up getting parsed multiple times in this weird way is so that when the [TaskScheduler](https://github.com/angular/protractor/blob/master/lib/taskScheduler.ts) creates a task for each entry in multicapabilities, the config object in each child process is built up the same way it would if we were running in a single process. 

We could really use some unit tests around config, particularly how configs get built with sharding and multicapabilities. Unfortunately, I don't think we have any tests that make sure overriding config with command line flags works as expected.

That said, this does look ok to me, and I think it's a step in the right direction. But there's enough potential for unintended consequences that I think @seanmay should take a look at it as well. Thanks for the PR!

lib/launcher.ts
NickTomlin: Totally! I'd be all for either `argv` or `commandLineFlags`|`commandLineConfig`.

seanmay:
Yeah. I'm hitting fun walls in this exact area, in other aspects of the program. Another consideration (and reason for multiple parsing of the config) is that it is possible to add configuration out of sync with the bootstrapping of the application.

I found this out via the e2e spec, rather than via intuiting it while trying to learn the internals... so there is that.

NickTomlin:
@seanmay yeah it's tough; would you be up for doing a hangout to pair on this? I'd love to talk through the issues/options.

deveshsinghal22:
I am trying to update spec using promise, but this is not working.
I am missing anything?

NickTomlin:
Doing some personal PR housekeeping. Closing this.

Issue #4140, opened 2017-03-07
Browser.refresh function does not keep Referer and other header parameters.


mmikirtumov:
In some cases of our business logic we need to refresh the page and keep in request header Referer value and other values. But actually when we use method browser.refresh it does not send all these information in header. In protractor when we do not set ignoreSynchronization=true, protractor using browser.get(href).
And in documentation there is the following 
# Makes a full reload of the current page and loads mock modules before

We need to have the method to refresh the page and get Referer value from request header without using ignoreSynchronization=true.

NickTomlin:
You should be able to use a combination of ignoreSynchronization and `waitForAngular` to preserve the referrer. 

Because of its usage and debugging nature, your question is better suited for StackOverflow or Gitter. Please [ask a question there with the 'protractor' tag](http://stackoverflow.com/questions/ask?tags=protractor) or post in the [Gitter Channel](https://gitter.im/angular/protractor) to get help. 

If you are not able to find a solution and believe this is an issue with Protractor core, please submit an issue and follow the [issue template](https://github.com/angular/protractor/blob/master/.github/ISSUE_TEMPLATE).

Thanks! 😄

mmikirtumov:
Hi Nick,

Thank you very much, but after when I use waitForAngular, it does not fix my problem, because from any action .click(), .isDisplayed() etc., our team should use then logic. And it is hard to maintain code. So every time we should use 
```
element.click().then(function(){
})
```
I have created on pull request to solve this problem.

https://github.com/angular/protractor/pull/4142

PR #4142, opened 2017-03-09
Add reload function which will send Referer value when refresh browser

mmikirtumov:
feat(lib\browser.ts): Add reload function wich will send Referer
Ticket number for mentioned feature is:
Browser.refresh function does not keep Referer and other header
parameters. #4140
https://github.com/angular/protractor/issues/4140

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -1015,6 +1015,23 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
   }
 
   /**
+   * @see webdriver.WebDriver.refresh
+   *
+   * Makes a reload of the current page.
+   *
+   * Method calls location.reload function of browsers
+   */
+  reload() {
+    if (this.ignoreSynchronization) {
+      return this.driver.navigate().refresh();
+    }
+    
+    return this
+        .executeScriptWithDescription(
+            'return location.reload()', 'Protractor reload page');
+  }
+
+  /**
    * Mixin navigation methods back into the navigation object so that
    * they are invoked as before, i.e. driver.navigate().refresh()
    */


mmikirtumov:
I signed it!

NickTomlin:
I'd hesitate to add another method here. Can your issue in https://github.com/angular/protractor/issues/4140#issuecomment-285274253 just be solved by a helper that does this for your specific use case?

```
browser.executeScriptWithDescription(
              'return location.reload()', 'Protractor reload page');
```

mmikirtumov:
We can create wrapper to add this method. We have multiple protractor projects and also we can create npm package to easily add into project. But I am just wondering if it is only browser refresh functionality and it is not working in a proper way, why do you hesitate to add this method. Because when you refresh your browser manually, press the key F5 or just click on refresh button it send Referer and other header values. And also Selenium driver also works in this way, in protractor we have the following code:
`refresh(opt_timeout?: number) {
      return this.driver.navigate().refresh();
    }

    return this
        .executeScriptWithDescription(
            'return window.location.href', 'Protractor.refresh() - getUrl')
        .then((href: string) => {
          return this.get(href, opt_timeout);
        });
  }`

this.get(href,opt_timeout) means that we just say browser to open the existing link, but there are many difference between refresh and open the current link.
I think this new method will solve this issue and in future release there will be no need to create wrapper.
Which is the main reason that you hesitate to add this method? 

Thanks a lot!

musclor:
Hi,

I have tried : 
`browser.executeScriptWithDescription(
              'return location.reload()', 'Protractor reload page');`
but it does not work for my use case.

The browser is receiving a cookie from my server after user signs in but it is lost when I ask Protractor to reload the page.

I have also tried : 
`browser.driver.navigate().refresh();`

but I still have the same issue.

I am using : 

- Protractor v.5.1.1 
- Webdriver-manager  v12.0.4

mmikirtumov:
I have tried many times and looks like protractor do not remove cookie.

Could you please run 
`location.reaload()`
 in your browser console and see does the cookie exists?
And also when you click on reload button or F5 does this cookie exists?

Also you can try the following
location.refresh(true)
According to [link](https://www.w3schools.com/jsref/met_loc_reload.asp)
Parameter for this function means:
Optional. Specifies the type of reloading:
false - Default. Reloads the current page from the cache.
true - Reloads the current page from the server.

May also you can try location.refresh(true).

musclor:
Hi @mmikirtumov,

Thank you for your confirmation and advices.
You are right, Protractor does not remove the cookies.

I finally managed to make my test work but I had to add a sleep() instruction before the reload() instruction...

Hereafter my test case : 
```
username.sendKeys('test');
password.sendKeys('ok');
connexion.click(); // submit the form to the server and once authenticated Angular redirects to homepage.
        
browser.sleep(1000);
        
browser.executeScriptWithDescription(
        'return location.reload()', 'Protractor reload page');          
```

Actually, with the sleep() I have tried using : 
`browser.driver.navigate().refresh();`
and even
`browser.refresh();`
instead of
```
browser.executeScriptWithDescription(
         'return location.reload()', 'Protractor reload page');  
```
and in all the cases now Protractor refreshes the homepage correctly and the cookies are kept.

I do not know why the sleep() is needed knowing I have left the Control Flow enabled.

mmikirtumov:
Hi @musclor ,

I would suggest you to use ExpectedConditions.urlIs logic.
The documentation you can find [link](http://www.protractortest.org/#/api?view=ProtractorExpectedConditions.prototype.urlIs)
Maybe this approach will solve your problem.

Thanks

musclor:
Thank you @mmikirtumov for your suggestion.

It works like a charm and it makes my test more understandable for the readers.

Issue #841, opened 2014-05-22
IE stability issue - occasional Javascript errors


juliemr:
Example: 

```
     UnknownError: JavaScript error (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 129 milliseconds
Build info: version: '2.41.0', revision: '3192d8a', time: '2014-03-27 17:17:32'
System info: host: 'WIN-8LIIH1BGM1D', ip: '172.20.42.188', os.name: 'Windows Server 2008 R2', os.arch: 'x86', os.version: '6.1', java.version: '1.6.0_38'
Session ID: ce12158e-4456-4529-8cc9-b8171739c296
Driver info: org.openqa.selenium.ie.InternetExplorerDriver
Capabilities [{platform=WINDOWS, elementScrollBehavior=0, javascriptEnabled=true, enablePersistentHover=true, ignoreZoomSetting=false, browserName=internet explorer, enableElementCacheCleanup=true, unexpectedAlertBehaviour=dismiss, version=11, cssSelectorsEnabled=true, ignoreProtectedModeSettings=false, requireWindowFocus=false, allowAsynchronousJavaScript=false, initialBrowserUrl=, handlesAlerts=true, nativeEvents=true, takesScreenshot=true}]
   Stacktrace:
     UnknownError: JavaScript error (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 129 milliseconds
Build info: version: '2.41.0', revision: '3192d8a', time: '2014-03-27 17:17:32'
System info: host: 'WIN-8LIIH1BGM1D', ip: '172.20.42.188', os.name: 'Windows Server 2008 R2', os.arch: 'x86', os.version: '6.1', java.version: '1.6.0_38'
Session ID: ce12158e-4456-4529-8cc9-b8171739c296
Driver info: org.openqa.selenium.ie.InternetExplorerDriver
Capabilities [{platform=WINDOWS, elementScrollBehavior=0, javascriptEnabled=true, enablePersistentHover=true, ignoreZoomSetting=false, browserName=internet explorer, enableElementCacheCleanup=true, unexpectedAlertBehaviour=dismiss, version=11, cssSelectorsEnabled=true, ignoreProtectedModeSettings=false, requireWindowFocus=false, allowAsynchronousJavaScript=false, initialBrowserUrl=, handlesAlerts=true, nativeEvents=true, takesScreenshot=true}]
==== async task ====
WebDriver.executeScript()
==== async task ====
[...]

Issue #4143, opened 2017-03-09
Protractor 5.1.1 + IE11, test cases fail randomly


aozolin:
Hi,

I have a bunch of test cases to run against IE11. With protractor 4.0.14 + selenium server 2.53.1 it works fine. But with the latest protractor 5.1.1 + selenium server 3.3.0 it doesn't. Some runs passed success, but some doesn't. It fails unexpected with random timeout errors.

Please investigate and fix the issue.

- Node Version: `v6.9.1`
- Protractor Version: `5.1.1`
- Angular Version: `2.4.7`
- Browser(s): `IE11 (11.0.9600.18538)`
- Operating System and Version `Windows 8.1`

```
require('ts-node/register');
var helpers = require('./webpack/helpers');
var Jasmine2HtmlReporter = require('protractor-jasmine2-html-reporter');
var XMLReporter = require('ruru-protractor-junit-reporter');

exports.config = {
  seleniumServerJar: 'node_modules/protractor/node_modules/webdriver-manager/selenium/selenium-server-standalone-3.3.0.jar',
  localSeleniumStandaloneOpts: {
      jvmArgs: ['-Dwebdriver.ie.driver=node_modules/protractor/node_modules/webdriver-manager/selenium/IEDriverServer_Win32_3.3.0.exe']
  },

  resultJsonOutputFile: 'my-protractor-e2e-results.json',
  exclude: [],

  suites: {
    mySuite: [
      helpers.root('test/e2e/**/client-server/my.test.e2e.ts')
    ]
  },

  framework: 'jasmine2',

  allScriptsTimeout: 110000,
  getPageTimeout: 110000,

  plugins: [{
    path: 'node_modules/protractor-istanbul-plugin'
  }],
  jasmineNodeOpts: {
    showTiming: true,
    showColors: true,
    isVerbose: false,
    includeStackTrace: false,
    defaultTimeoutInterval: 600000
  },
  directConnect: false,

  multiCapabilities: [
    {
      'browserName': 'internet explorer',
      'platform': 'ANY',
      'version': '11',
      'nativeEvents': false,
      'unexpectedAlertBehaviour': 'accept',
      'ignoreProtectedModeSettings': true,
      'enablePersistentHover': true,
      'disable-popup-blocking': true,
      'ignoreZoomSetting': true,
      shardTestFiles: true,
      maxInstances: 3
    }
  ],

  params: {
    url: '/test-app/'
  },

  onPrepare: function () {
[...]

PR #4149, opened 2017-03-15
fix(navigation): ignore unknown JS errors in IE

thorn0:
The `err` object doesn't have the `code` property any more (Selenium Server Standalone 3.3.1 + IEDriver win32 3.3.0), so we need a new way to detect those errors. See #841


thorn0:
This might be related to #4143.

thorn0:
Can we have 5.1.2 or at least a patch release with just this fix?
In my observation, IE got flakier after upgrading to Protractor 5, and this fix seems to repair it.

heathkit:
Thanks for the fix! I also need this, I'll ping the caretakers to get a new release.

Issue #2253, opened 2015-06-16
Where is default timeout for browser.wait?


PaddyMann:
I'm running tests with Jasmine2, and have config set up with:

    allScriptsTimeout: 11000,
    getPageTimeout: 10000,
    jasmineNodeOpts: {
      defaultTimeoutInterval: 90000
    }

I used to specify a timeout with each `browser.wait` call. e.g.:

    browser.wait(function () {
       ...
    }, 3000);

But when getting cross-browser tests to run, I discovered I didn't need to specify the 3000 manually and was better off letting it run until it hit the configured limits.

But... now I'm running the tests locally again and it seems to ignore the configuration. I've checked that it is set up in the same way (using `browser.getProcessedConfig` and reviewing the output), but all my `browser.wait`s are failing too quickly. e.g. `Failed: Wait timed out after 110ms`.

Am I missing something obvious?

sjelin:
Could you give a small example?

PaddyMann:
From a page object:

    lblSuccessMessage: {
      get: function() {
        return $('.edit-team-member_success-message');
      }
    },

    ...
    
    goSubmitAndClose: {
      value: function () {
        var page = this;
        page.btnAddTeamMember.click();
        browser.wait(function () {
          return page.lblSuccessMessage.isPresent();
        });
        page.btnCloseSuccess.click();
      }
    }

The browser.wait gives the error:

    Message:
        Failed: Wait timed out after 65ms
      Stack:
        Error: Wait timed out after 65ms
            at Array.forEach (native)
        From: Task: <anonymous wait>
            at Fn.module.exports.Page.create.goSubmitAndClose.value (  ...  /addTeamMember.page.js:111:15)
            at Object.<anonymous> (  ...  /addTeamMember.spec.js:67:12)
        From: Task: Run fit("should allow user to add a team member with limited access") in control flow
            at Array.forEach (native)
        From asynchronous test: 
        Error
            at Suite.<anonymous> (  ...  /addTeamMember.spec.js:62:5)
            at Suite.<anonymous> (  ...  /addTeamMember.spec.js:35:3)
            at Object.<anonymous> (  ...  /addTeamMember.spec.js:3:1)

juliemr:
[...]

PR #4154, opened 2017-03-16
docs(browser.wait): add default browser.wait default

wesww:
There's a bit of confusion about the default behavior of the second parameter for the `browser.wait` function (e.g. https://github.com/angular/protractor/issues/2253). Adding this info to the docs might clear that up

```
browser.wait(EC.textToBePresentInElement($('#foo'), 'Example words'), 5000); // 5 seconds
browser.wait(EC.textToBePresentInElement($('#foo'), 'Example words')); // 30 seconds
```
default time comes from an optional param in user's `protractor.conf.js` file
```
jasmineNodeOpts: {
  defaultTimeoutInterval: 30000
}
```

which falls back to the default set in  `lib/configParser.ts` if not set:
```
jasmineNodeOpts: {showColors: true, defaultTimeoutInterval: (30 * 1000)},
```
https://github.com/angular/protractor/blob/bc583321a233453fc2b89472013b2ec3e1d6b6f9/lib/configParser.ts#L37

### DIFF
--- a/lib/selenium-webdriver/webdriver.js
+++ b/lib/selenium-webdriver/webdriver.js
@@ -277,7 +277,7 @@ webdriver.WebDriver.prototype.call = function(fn, opt_scope, var_args) {};
  *           function(!webdriver.WebDriver): T)} condition The condition to
  *     wait on, defined as a promise, condition object, or  a function to
  *     evaluate as a condition.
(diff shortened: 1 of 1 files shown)


wesww:
I signed it!

Issue #4148, opened 2017-03-13
Protractor 5.1.1 selenium-webdriver @types version inconsitency


CAL-RJ:
Bug report
---

- Node Version: `6.9.4`
- Protractor Version: `5.1.1`
- Angular Version: `2.4.8`
- Browser(s): `n/a`
- Operating System and Version `Windows 7`

I've recently upgraded to Protractor 5.1.1 and am facing some issues when setting cookies via browser.manage().addCookie()

The API has changed between versions 2 and 3 of Selenium-webdriver to expect an object rather than the previous 2..6 arguments. When I make the changes to my code to use the object, the typescript compiler complains saying that it expects 2..6 arguments.

old api:

`browser.manage().addCookie('cookieName',  'cookieVal');`

new api:

`browser.manage().addCookie({name:'cookieName', value: 'cookieVal'});`

I think this is because the @types/selenium-webdriver in the package.json of protractor v5.1.1 is pointing at version 2.53.39. The version of the actual selenium-webdriver the same package.json is referencing is 3.0.1.

I think they should both be version 3.*

[related stack overflow question](http://stackoverflow.com/questions/42722950/protractor-5-1-1-selenium-webdriver-version-inconsitency)

NickTomlin:
Thanks for raising this! This is something that we will need to take care of internally. 

The current workaround in Craig's [SO answer](http://stackoverflow.com/a/42791857/1048479):

```
(browser.manage() as any).addCookie({name:'cookieName', value: 'cookieVal'});
```

I can pick this up later in the week unless someone has time to work on it before then.

cnishina:
This should be updated. @NickTomlin do a search for `// TODO`'s. We have a few places that we hacked in the `as any` to get Protractor's lib to transpile.

NickTomlin:
@cnishina it looks like `webdriver-extender` is the underlying issue here since the types point to selenium `2.x`. 

I was able to resolve some todos in #4158 but there's still a mismatch for `addCookie`. I'll do some digging in web driver extender to try and update there as well but the code is pretty opaque to me at the moment 😄 Any tips would be very much appreciated

soad666p:
any update on this,  as i still see a problem  with adding the cookie and get cookie

whyboris:
November 27th, problem persists with `"protractor": "5.2.0"`

monkpit:
[...]

PR #4158, opened 2017-03-16
fix(typings): Update selenium typings and fix tsc errors

NickTomlin:
This is a first crack at solving the typing issues raised in https://github.com/angular/protractor/issues/4148. I picked up some of the low-hanging fruit in regards to typing fixes

**Disclaimer**: I'm still getting used to the TS ecosystem, so if I missed something obvious here please educate me 😄 

Also, on a side note, who would we feel about bumping TS to `2.1.x`? We are still on `2.x`; I know there's hesitancy about moving to `2.2` or `2.3` so we could also wait until those are more accepted in the angular/angular tooling community.

### DIFF
--- a/lib/driverProviders/direct.ts
+++ b/lib/driverProviders/direct.ts
@@ -7,7 +7,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import {Capabilities, WebDriver} from 'selenium-webdriver';
 import {Driver as ChromeDriver, ServiceBuilder as ChromeServiceBuilder} from 'selenium-webdriver/chrome';
-import {Driver as FirefoxDriver} from 'selenium-webdriver/firefox';
+import {Driver as FirefoxDriver, ServiceBuilder as FirefoxServiceBuilder} from 'selenium-webdriver/firefox';
 
 import {Config} from '../config';
 import {BrowserError} from '../exitCodes';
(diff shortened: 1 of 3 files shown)


seanmay:
I am on board with the jump to TS2.1, short of anything I might be missing in terms of the prior team's reluctance.

While my favourite 2.1 feature is Readonly, I feel like keyof, Partial and Pick would be helpful in keeping type information sound, in various corners of Protractor.

seanmay:
The code looks good, but I want to hold off on the merge, momentarily, due to discussions that Vikram and I have had around a move forward strategy, regarding bug-fixes and features which will break typing, externally, versus those that won't.

The intent, I believe, is to move the breaking changes and the like (ES6 promises inclusive) to a major branch, and complete the minor changes in master. Once this is done, I'll be back here, but I think this code is good to go.

NickTomlin:
Hmm, after some more digging this may not actually resolve the issue in #4148 so holding off on a merge is good. I still think updating the typings and removing some of the todos is good though. 

Restructuring releases sounds good; let me know what we settle on and if there's any way I can help.

mgiambalvo:
FYI, Protractor has to use the same TypeScript version that Angular is on. We sort of unintentionally upgraded to 2.1 while Angular was still on 2.0 and created a mess, so we had to downgrade.

Upgrade TypeScript as often as you can, but don't get ahead of Angular!

heathkit:
Sorry, I'm going to close this as it's pretty outdated and I have no idea what's going on. We do need to update our selenium typings, though. It's also important to remember that we need to be on the same version of typescript as Angular.

Issue #4110, opened 2017-02-23
Feature Request/ Implementation question| hooking driver creation methid


daniel4x:
Feature Request

I'm looking for a guidelines or help implementing retry mechanism for driver creation. 
It's really useful and important for testing remote real devices while retrieve one from a pool of devices. 

- **Reasons for adopting new feature**:
Remote testing could fail sometimes when trying to retrieve a device because the device already in use or unavailable. 
I would like to prevent this by retry in X sec later or by getting a different device. 
  
- **Is this a breaking change? (How will this affect existing functionality)**
Not sure.

NickTomlin:
Could you flesh out what this might look like with some pseudocode? I'm not familiar with remote device testing

wswebcreation:
Hi @NickTomlin,

What @daniel4x means is the following. Some cloudproviders don't have automatic queueing like Sauce Labs has. For example, you request 10 instances of Chrome, you only have 5, so 5 are queued.

Cloudproviders like for example Perfecto (who has physical devices and browsers) can't queue. The will give back a "error"-message saying that the driver could not be created because they are (all) busy.

In protractor we can't "hook" onto the instantiation of the driver with (get)multi-capabilities. It would be nice that the "error"/"info"-message could be processed with a custom script with some kind of hook or something, like this:

```
// calling with
multicapabilities:[{
        browserName:chrome
    }, {
        browserName: firefox
}],

// when error message of driver
onDriverError: (message){
    // process the message and trigger the driver start again for x times, this can be done by the script
    // user is responsible of handling the amount of retries and so on
}
```

Hope this makes is more clear.

Grtz

Wim

wswebcreation:
Hi @NickTomlin

Did you had time to check this?

NickTomlin:
I think this seems like a valid use case; I personally won't be able to get to it but i'd welcome a PR. 

I'm trying to think of the best way to handle this, do you have an example integration with perfecto? It seems like it might be something that best fits into an additional method for a driver provider that could negotiate this on behalf of the user (I'm open to exposing this via the root `config` if necessary as well).

wswebcreation:
Tnx @NickTomlin

@daniel4x will look at it and create a PR

NickTomlin:
@wswebcreation sure! Feel free to use this thread as a place to discuss options. Thanks!

daniel4x:
Hi @NickTomlin,
@wswebcreation and I came up with an idea for implementation of this feature. 
[...]

PR #4165, opened 2017-03-19
Async driver creation & retry mechanism for driver creation

daniel4x:
Please see: 
https://github.com/angular/protractor/issues/4110

Added a retry mechanism for driver creation.
For users who uses cloud providers such as Perfecto, Sauce labs, etc ... or for local usage.

There's 2 new configurations:
1. createBrowserAsync  - true / false, this option enable/disable async driver creation.
2. retryMechanismConf - enable/disable retry mechanism, set the number of retries and delay between each retry.

Using this 2 configuration the user will be able to make a retry in case driver creation failed. 

Passed local tests (npm test)

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -474,6 +474,22 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
   }
 
   /**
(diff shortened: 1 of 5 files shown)


daniel4x:
I signed it!

PR #4166, opened 2017-03-19
Async driver creation & retry mechanism for driver creation

daniel4x:
Please see: 
https://github.com/angular/protractor/issues/4110

Added a retry mechanism for driver creation.
For users who uses cloud providers such as Perfecto, Sauce labs, etc ... or for local usage.

There's 2 new configurations:
1. createBrowserAsync  - true / false, this option enable/disable async driver creation.
2. retryMechanismConf - enable/disable retry mechanism, set the number of retries and delay between each retry.

Using this 2 configuration the user will be able to make a retry in case driver creation failed. 

Passed local tests (npm test)

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -474,6 +474,22 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
   }
 
   /**
+   * Same as forkNewDriverInstance just asynchronously
+   * 
+   * @param {boolean=} useSameUrl Whether to navigate to current url on creation
+   * @param {boolean=} copyMockModules Whether to apply same mock modules on creation
+   * @param {boolean=} copyConfigUpdates Whether to copy over changes to `baseUrl` and similar
+   *   properties initialized to values in the the config.  Defaults to `true`
+   *
+   * @returns {Promise<ProtractorBrowser>} A browser instance.
+   */
+  forkNewDriverInstanceAsync(
+      useSameUrl?: boolean, copyMockModules?: boolean,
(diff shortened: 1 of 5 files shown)


daniel4x:
Hey @NickTomlin , 
I made a pull request as you offered [here](https://github.com/angular/protractor/issues/4110).

I've passed travis-ci tests but circleci build failed.
Yet, when i run npm test locally (as i see, this is the test circleci run) i pass all the tests.

Can you help with this & code review, change code/doc if necessary?

Thanks!

daniel4x:
+ @wswebcreation

NickTomlin:
We are re-doing the merging/pr process internally here so I may not be able to get to this for awhile. Thanks for the PR though!

heathkit:
Thank you so much for the effort you put into this. Unfortunately, the logic around `createBrowser` is frustratingly complex, and duplicating it with an async version is the opposite direction we'd like to take. We'd really like to refactor that specific part of the project, especially in light of the issues we've seen around forking and restarting browsers.

I think I see what you're trying to achieve, and it would be valuable. But we'd need a solution that involves fewer changes, especially to such a confusing, sensitive area as createBrowser. Is there a way to achieve what you need by adding a hook to the plugin api, or maybe changing runner?

Sorry, I'm going to close this. But we'd welcome an implementation of this feature (providing the ability to configure retries for driver creation) that introduced less complexity and duplication. Simply making createBrowser async instead of having two versions would be an acceptable direction.

wswebcreation:
Tnx for the feedback @heathkit ,

We'll rethink this part.

PR #4173, opened 2017-03-20
Update home.html

zsong:


zsong:
Why it failed?

NickTomlin:
I don't think the partial is the right place for this documentation. This should probably be either in the web driver manager docs or in the [faq](https://github.com/angular/protractor/blob/master/docs/faq.md)

PR #4205, opened 2017-04-06
chore(docs): use phantomjs-prebuilt

Shyam-Chen:


heathkit:
Given that [PhantomJS is not supported](https://groups.google.com/forum/#!topic/phantomjs/9aI5d-LDuNE) we might just remove that bit from the doc someday.

PR #4219, opened 2017-04-11
docs(await): Update example to show debugging with chrome devtools

mgiambalvo:
Change the async/await example to use ts-node. Also, when using
SELENIUM_PROMISE_MANAGER=false, you can debug tests using
`node --inspect` and the chrome debugger.

Also, update the docs around timeouts to include Angular vs AngularJS


mgiambalvo:
Had to move the example config from js to ts so that you can run the example using ts-node without compiling. This is the pattern angular-cli uses. Also, I have some extra fixes for the CI - this is intended to be merged without squashing.

CircleCI is still failing, due to the known issue with Firefox and directConnect.

PR #4227, opened 2017-04-14
Export Runner, not just its type.

LucasSloan:


vikerman:
This was for a very specific use where we had to set the spec name in the browser capabilities and hence we wanted to override the implementation of createBrowser.

sjelin:
I do not like this.  The runner is not a public API and we shouldn't be encouraging people to hack it like this.  If people are determined to do this kind of hack, they can use `require('protractor/built/runner.js').Runner;`.

The long term solution to @vikerman's problem is to have a plugin hook in createBrowser that allows people to modify the capabilities (https://github.com/angular/protractor/issues/4230).

heathkit:
@sjelin agreed, I think it would be better to keep runner private.

NickTomlin:
I think I misunderstood the benefits here; my bad :(. I believe @sjelin and @heathkit are right on this. I love the idea of exposing more "building blocks" to external users but it would be best to do that with a crafted external API and not by just exposing the internals (as they stand) to the outside world.

@vikerman can we revert this and go with a `require` based reach into the internals as a short term solution while we eventually move towards the solution in #4230 ?

PR #4228, opened 2017-04-14
Use http when relaying selenium commands through sauce connect tunnel

ndmanvar:


ndmanvar:
@cnishina @heathkit Could you please review and merge in?

lib/driverProviders/sauce.ts
```diff
@@ -63,10 +63,11 @@ export class Sauce extends DriverProvider {
...
     this.config_.capabilities['accessKey'] = this.config_.sauceKey;
     this.config_.capabilities['build'] = this.config_.sauceBuild;
-    let auth = 'https://' + this.config_.sauceUser + ':' + this.config_.sauceKey + '@';
-    this.config_.seleniumAddress = auth +
-        (this.config_.sauceSeleniumAddress ? this.config_.sauceSeleniumAddress :
-                                             'ondemand.saucelabs.com:443/wd/hub');
+
+    let auth = this.config_.sauceUser + ':' + this.config_.sauceKey + '@';
+    this.config_.seleniumAddress = this.config_.sauceSeleniumAddress ?
+        'http://' + auth + this.config_.sauceSeleniumAddress :
```
seanmay: Is there a reason that this would never, ever be secure?

ndmanvar: @seanmay This would not be secure (http, not https) when relaying Selenium commands through Sauce Connect Tunnel. (i.e. when pointing at Sauce connect's selenium listener/relay, rather than sending selenium command over the internet to Sauce's https endpoint).

seanmay: I suppose where I was going with that is that is that we need to support `http`, but in this solution, we're now guaranteeing that https is just as impossible as http is, now, if there is a path that has been unconsidered, thus far.

Would a config flag work for you?

heathkit:
Sorry, closing in favor of #4236, which lets you use either protocol.

PR #4236, opened 2017-04-18
feat(saucelabs): Add Sauce Labs protocol customization support

xaka:
Sauce Connect Proxy listens on 4445 port by default for Selenium communications and always uses HTTP protocol instead of HTTPS. It's not rare to run Sauce Connect Proxy locally or inside private network meaning we're forced to use HTTP and `sauceSeleniumAddressProtocol` is the way to customize which protocol to use.

P.S. In case there are no objections, it'd be nice to have a patch release otherwise we're stuck with dirty looking monkey-patching.

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -137,6 +137,13 @@ export interface Config {
    */
   sauceBuild?: string;
   /**
+   * If true, Protractor will use http:// protocol instead of https:// to
+   * connect to Sauce Labs defined by sauceSeleniumAddress.
+   *
+   * default: false
+   */
+  sauceSeleniumUseHttp?: boolean;
+  /**
    * Use sauceSeleniumAddress if you need to customize the URL Protractor
    * uses to connect to sauce labs (for example, if you are tunneling selenium
    * traffic through a sauce connect tunnel). Default is
--- a/lib/driverProviders/sauce.ts
+++ b/lib/driverProviders/sauce.ts
@@ -63,7 +63,8 @@ export class Sauce extends DriverProvider {
     this.config_.capabilities['username'] = this.config_.sauceUser;
     this.config_.capabilities['accessKey'] = this.config_.sauceKey;
     this.config_.capabilities['build'] = this.config_.sauceBuild;
-    let auth = 'https://' + this.config_.sauceUser + ':' + this.config_.sauceKey + '@';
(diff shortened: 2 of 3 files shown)


xaka:
Turned out there is another PR https://github.com/angular/protractor/pull/4228 fixing the same problem with slightly different approach. I'll leave it up to maintainers to decide which one they prefer to keep.

lib/config.ts
```diff
@@ -137,6 +137,12 @@ export interface Config {
    */
   sauceBuild?: string;
   /**
+   * Use sauceSeleniumAddressProtocol if you need to customize the protocol Protractor
+   * uses to connect to sauce labs URL defined by sauceSeleniumAddress. Default is
+   * https://
+   */
+  sauceSeleniumAddressProtocol?: string;
```
heathkit: Could this just be a boolean that defaults to false? `sauceSeleniumUseHttp`

lib/driverProviders/sauce.ts
```diff
@@ -63,7 +63,8 @@ export class Sauce extends DriverProvider {
     this.config_.capabilities['username'] = this.config_.sauceUser;
     this.config_.capabilities['accessKey'] = this.config_.sauceKey;
     this.config_.capabilities['build'] = this.config_.sauceBuild;
-    let auth = 'https://' + this.config_.sauceUser + ':' + this.config_.sauceKey + '@';
+    let auth = (this.config_.sauceSeleniumAddressProtocol || 'https://') + this.config_.sauceUser +
```
heathkit: I'd suggest something like

```
let protocol = this.config_.sauceSeleniumUseHttp ? 'http://' : 'https://'
let auth = protocol + this.config_.sauceUser + ':' + this.config_.sauceKey + '@';
```

Also, if you wanted to build the address with the `url` module or template strings or something, that'd also be nifty.

lib/config.ts
xaka: changed `sauceSeleniumAddressProtocol` to `sauceSeleniumUseHttp`

lib/driverProviders/sauce.ts
xaka: - adjusted auth part construction to use `sauceSeleniumUseHttp`
- decided not to switch to url/templates to keep changes to bare minimum

xaka:
@heathkit addressed the feedback; please take another look when you've got time; thanks!

PR #4257, opened 2017-05-03
fix(elementExplorer): Properly construct paths for ElementExplorer

### LINKED ISSUE #4011: Element Explorer totally unresponsive
- Node Version: `7.4.0`
- Protractor Version: `5.0.0`
- Browser: `Chrome`
- OS: `Windows 7 64bit`

I'm following the guide [here](https://github.com/angular/protractor/blob/master/docs/debugging.md) for setting up a REPL, which says to run protractor thusly: `protractor --elementExplorer`.  I do this, it launches a Chrome window, shows output in the console that suggests everything worked (e.g. "server started...", "debugger listening.."), and leaves me at a prompt (`>`).  

Nothing I type in this prompt produces any effect/response at all. I tried the examples given, such as `browser.get('http://www.angularjs.org')`, but pressing ENTER simple adds a carriage return to the screen and nothing else. I can keep hitting ENTER and the cursor keeps dropping down a line, but there is no response from protractor, nothing happens in the browser, etc. The only thing that produces a response is CTRL+C or CTRL+D, which serve to let me know the process is still running but that's about it. 

protractor-element-explorer

Note, if I keep trying to enter commands (they just pile up with no response), I eventually get this:

`(node:3012) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 break listeners added. Use emitter.setMaxListeners() to increase limit`

heathkit:
Fixes #4011

### DIFF
--- a/lib/debugger/debuggerCommons.js
+++ b/lib/debugger/debuggerCommons.js
@@ -1,4 +1,5 @@
 var baseDebugger = require('_debugger');
+var path = require('path');
 
 /**
  * Create a debugger client and attach to a running protractor process.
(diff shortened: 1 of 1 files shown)


heathkit:
Doesn't actually fix.

Issue #4253, opened 2017-05-01
directConnect not working with FF 52+ and Protractor 5.1.1 (unable to parse new session response)


heathkit:
Protractor 5.1.1 is using version 3.0.1 of the Selenium JS bindings. When using `directConnect: true` Protractor will control geckodriver directly. Unfortunately, this version of the JS bindings is no longer compatible with the latest geckodriver (See https://github.com/mozilla/geckodriver/issues/555 https://github.com/mozilla/geckodriver/issues/529 and https://github.com/SeleniumHQ/selenium/issues/3607).

The solution is to use selenium standalone when testing with Firefox. The latest version of the selenium standalone server is compatible with Protractor's selenium JS bindings and with geckodriver. You can update and launch the standalone server with

```
webdriver-manager update
webdriver-manager start
```

and set `seleniumAddress: http://localhost:4444`. For an example using selenium standalone, see [protractor cookbook](https://github.com/angular/protractor-cookbook/tree/master/protractor-javascript).

In the meantime, `directConnect: true` will not work with Firefox until the external issues are fixed and Protractor's selenium-webdriver version is updated.

heathkit:
There are various issues talking about Firefox that are somewhat out of date or confusing, so I thought I'd collect the latest info in a new issue.

Protractor 5.1.1 requires the latest Selenium 3 and only works with Firefox 52 or higher (you might be able to get Firefox < 47 working with `marionette: false`, I'm not sure). You need to use a selenium standalone server, `directConnect: true` will not work due to the above external issues. This is the case for geckodriver 0.15. Firefox 52, and Protractor 5.1.1, but may change with future versions.

Geckodriver is maturing and getting closer to the spec, so I'm betting directConnect will work again in future Protractor releases with updated selenium-webdriver JS bindings.

NickTomlin:
Thanks for collecting this; super helpful when dealing with issues that may have different symptoms but all reflect problems with FF

mleu:
Thanks also for collecting this. As far as I (naively) see there seem to be no relevant external bugs open anymore, so the "only" thing missing would Protractor updating it's Selenium JS Bindings to match the [requirements](https://github.com/mozilla/geckodriver#supported-clients) of geckodriver...?

heathkit:
[...]

PR #4264, opened 2017-05-05
chore(ci): Disable FF directConnect test temporarily to unblock CI

heathkit:
Disable Firefox directConnect test until #4253 is fixed.

### DIFF
--- a/spec/directConnectConf.js
+++ b/spec/directConnectConf.js
@@ -14,7 +14,9 @@ exports.config = {
     'browserName': 'firefox',
   }],
   */
-  browserName: 'chrome',
+  capabilities: {
+    browserName: 'chrome',
+  },
 
   baseUrl: env.baseUrl + '/ng1/',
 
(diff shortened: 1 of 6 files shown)


spec/directConnectConf.js
```diff
@@ -5,11 +5,16 @@ exports.config = {
...
+
+  // Disabled until https://github.com/angular/protractor/issues/4253 is resolved
+  /*
   multiCapabilities: [{
     'browserName': 'chrome'
   }, {
     'browserName': 'firefox',
   }],
+  */
+  browserName: 'chrome',
```
cnishina: `browserName` should be set in `capabilities`. This should not matter since chrome is the default value if no capabilities are set.
```
capabilities: {
  browserName: 'chrome'
}
```

heathkit: Hah, that shows you how often I write protractor configs...

PR #4285, opened 2017-05-13
feat(saucelabs): Removed unnecessary username and accesskey from cap…

sirwalt:
…abilities and fixed build in capabilities to be only overwritten under condition

There was no actual reason to set username and accesskey as capabilities on the selenium driver. Sourcelabs doesn't require those.

Overwriting the build capability from the sauceBuild config parameter seams to be logical since it couldn't differ between multiple capabilities within a single build, but it is unexpected. Without any check if the value of souceBuild is defined, there is no chance to set it within capabilities as someone is used to do.


sirwalt:
I signed it!

sirwalt:
I signed it!

PR #4326, opened 2017-06-15
docs(browser-support): Fixed incorrect example

marklagendijk:
Fixed incorrect example of setting window size with headless Chrome. When `800x600` is used it is ignored. The correct syntax is `800,600`.


marklagendijk:
I signed it!

Issue #4307, opened 2017-06-01
Element explorer doesn't work on Node 8


monkpit:
Bug report
---

- Node Version: `8.0.0`
- Protractor Version: `5.1.2`
- Angular Version: `n/a`
- Browser(s): `Chrome / chromedriver 2.29.0`
- Operating System and Version `Mac Sierra 10.12.5`
- Your protractor configuration file `n/a`

After installing node v8.0.0 and npm v5.0.0, reinstalling protractor globally and running `webdriver-manager update`, I cannot run `protractor --elementExplorer` because I receive the following error:

```
protractor --elementExplorer
(node:76684) [DEP0022] DeprecationWarning: os.tmpDir() is deprecated. Use os.tmpdir() instead.
[11:04:10] I/hosted - Using the selenium server at http://localhost:4444/wd/hub
[11:04:11] I/protractor -
[11:04:11] I/protractor - ------- Element Explorer -------
[11:04:11] I/protractor - Starting WebDriver debugger in a child process. Element Explorer is still beta, please report issues at github.com/angular/protractor
[11:04:11] I/protractor -
[11:04:11] I/protractor - Type <tab> to see a list of locator strategies.
[11:04:11] I/protractor - Use the `list` helper function to find elements by strategy:
[11:04:11] I/protractor -   e.g., list(by.binding('')) gets all bindings.
[11:04:11] I/protractor -
module.js:487
    throw err;
    ^

Error: Cannot find module '_debugger'
    at Function.Module._resolveFilename (module.js:485:15)
    at Function.Module._load (module.js:437:25)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/usr/local/lib/node_modules/protractor/built/debugger/debuggerCommons.js:1:82)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
```

If I revert back to node 7.10.0 I don't get this error.

NickTomlin:
I don't think we are currently testing against node 8 so it makes sense that this may be broken. Thanks for bringing this up!

NickTomlin:
I'll try to dig into this in the next few days but a PR to fix this would be very welcome!

monkpit:
[...]

PR #4345, opened 2017-06-20
Incompatibility with node 8

zerdos:
Fixing the element explorer with node8 won't be trivial - until then it worth to mention in the readme that protractor not working with the newest node version well.

https://github.com/angular/protractor/issues/4307

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Protractor [![Build Status](https://travis-ci.org/angular/protractor.svg?branch=
 Compatibility
 -------------
 
-Protractor 5 is compatible with nodejs v6 and newer.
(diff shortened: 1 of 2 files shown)


wswebcreation:
@heathkit and @NickTomlin 

What do you think about this PR for the current NodeJS 8 support?

qiyigg:
Do we also have other compatibility issues besides protractor debugger/explorer?

qiyigg:
We only have compatibility issues with debugger and we don't want to use that debugger after node 8. We already had doc to show how to use chrome debugger and also explain the previous debugger was deprecated in node 8; therefore close this PR.

Issue #4385, opened 2017-07-15
CircleCI is broken due to drag and drop issue


heathkit:
Our CircleCI build [is failing](https://circleci.com/gh/angular/protractor/705). The test that's failing is "using an ActionSequence should drag and drop" from basicConf.js. I'm guessing this is due to a change in selenium webdriver around the actions API. It would be good to get this fixed so we can get a passing CI run again. If it's a big unfeasible problem, marking the test pending and referencing this issue would also be fine, as long as we investigate what's happening.

wswebcreation:
Will dive into it the coming days and create a PR for it

wswebcreation:
Hi @heathkit ,

I dove into this and I think it's related to https://github.com/angular/protractor/issues/583. What I see in the testapp of Protractor is that the html 5 slider is working, but it will only go to value 5. It doesn't mater if you try to slide it to 1000px to the right, it will still stick on 5.

We now have a few options:

1. change expected value 10 to 5. We know the drag and drop work, only not with HTML5 sliders due to a Chromedriver issue (maybe we can also close issue 583). Also refer to https://github.com/angular/protractor/issues/583 in a comment in the test why we adjusted it to 5.
2. accept that is wont work and skip this test

I'm more in favour of 1 because the dag and drop does work (set some sleeps into it to see if it's working and it's working), only not completely due to https://github.com/angular/protractor/issues/583. 

Like to hear what you think

BassemDev:
@heathkit do you face a problem with getCurrentUrl in circleCi with chrome, i can't get the new url after login to the home page it's always sending the old url which mean the /signIn page ?
I don't know the reason but with firefox on circleCi and locally (on my mac) everything works fine.
could you maybe share an example where you use getCurrentUrl on circleCi after signIn. meanWhile this is an example of my code:

```
import { browser, Config, element, by } from 'protractor';
import { assert } from 'chai';

describe('test sample', function() {

    function urlChanged(url) {
            return function () {
                return browser.getCurrentUrl().then(function(actualUrl) {
                    return url != actualUrl;
                });
            };
    };

    beforeEach(function() {
        browser.ignoreSynchronization = true;
    });

    it('should login', function() {
        browser.get('http://localhost:8080/#/signin');
        const email = element(by.name('email'));
        const password = element(by.name('password'));
        const button = element(by.css('[type="submit"]'));
        email.sendKeys('qa-test@fashion.cloud');
        password.sendKeys('fc123456789');
        button.click();
[...]

PR #4413, opened 2017-07-30
fix: fix all locator examples

wswebcreation:
replaced all `browser.findElement(..)` for `element(..)` to make the API documentation more consistent

### DIFF
--- a/lib/selenium-webdriver/locators.js
+++ b/lib/selenium-webdriver/locators.js
@@ -25,7 +25,7 @@ webdriver.By = {};
  *
  * @example
  * // Returns the web element for dog
- * var dog = browser.findElement(by.className('dog'));
+ * var dog = element(by.className('dog'));
  * expect(dog.getText()).toBe('Dog');
  *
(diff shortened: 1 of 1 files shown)


wswebcreation:
When looking @ this PR, can you also please check https://github.com/angular/protractor/issues/4385#issuecomment-315630653 for the broken circleCI?

cnishina:
This makes sense. We are trying to tell our users that these are inherited by the Protractor locator `by` from selenium-webdriver's By. Also with this change we are telling our users that this can be used by the `element` API.

I like it. 😄

wswebcreation:
Tnx, I did a webinar last week and saw this during my demo. It made people confuse, because they now saw different implementations. I think this would be more consistent.

wswebcreation:
What about the circleCI? Shall I also change that with a PR!

cnishina:
Hmmm... I don't think your changes should cause errors in the test suite. I plan to rerun circle then merge it

Issue #4151, opened 2017-03-15
Forked driver instance ignores waitForAngularEnabled


artur-charcenko:
Bug report
---

- Node Version: `6.10.0`
- Protractor Version: `5.11`
- Angular Version: `n/a`
- Browser(s): `Chrome`
- Operating System and Version ` OS X 10.11.6`

After upgrading from v4.0.14 to v5.1.1 all tests using `browser.forkNewDriverInstance()` started to fail. I tried using `waitForAngularEnabled` instead of `ignoreSynchronization` but got the same result.

Example test:
```
it('', () => {
    const otherBrowser = browser.forkNewDriverInstance();
    otherBrowser.ignoreSynchronization = true;
    return otherBrowser.get('https://google.com')
      .then(() => otherBrowser.$('body').isDisplayed())
      .then(isDisplayed => expect(isDisplayed).to.be.true);
  });
```

Error:
```
Error: Angular could not be found on the page https://google.com/.If this is not an Angular application, you may need to turn off waiting for Angular.
```

mtheoryx:
I've noticed this happening with jasmine as well, FWIW

johnlister:
forkNewDriverInstance is also called on browser.restart, which means after every test the browser properties including waitForAngularEnabled) are reset to their defaults. I suspect forkNewDriverInstance should be passing in the current browser to createBrowser as the parent such that the properties are copied - does this have any unexpected consequences?

annu13aquar:
I am facing the exact issue with Node, Protractor and Chrome driver upgrades to the above mentioned versions. I am using Jasmine framework. Copying the parent browser properties to the new browser instances doesn't fix the issue.
Are there any more updates on the issue ?

josepmc:
This bug also causes that after restarting a browser the user profile may not be the same. Therefore, it makes impossible to perform tests that should ensure that values have been saved before the browser is closed if these are kept in localStorage or cookies.

AshokKumarMelarkot:
I,m facing the same issue and tried with the fix provided by slotik and no luck :(
Anticipating the fix for the same ASAP.

polar-lukas:
In my case, the following worked:
```
browser2 = browser.forkNewDriverInstance();
browser2.waitForAngularEnabled(false).then(() => {
  browser2.get("yourUrlHere");
}
```
(and don't forget to use ```browser2.element(by.whatever...)```, that's another showstopper...)

rafalf:
annoying... not working

[...]

PR #4418, opened 2017-08-01
Potential fix for issue 4151: Forked driver ignores waitForAngularEnabled

slotik:
Very minimalistic candidate to fix https://github.com/angular/protractor/issues/4151.

My hope is that it won't break other stuff.

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -417,8 +417,8 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
    * Call waitForAngularEnabled() without passing a value to read the current
    * state without changing it.
    */
-  waitForAngularEnabled(enabled: boolean|wdpromise.Promise<boolean> = null):
-      wdpromise.Promise<boolean> {
+  waitForAngularEnabled(enabled: boolean|wdpromise.Promise<boolean> = null): boolean
+      |wdpromise.Promise<boolean> {
     if (enabled != null) {
       const ret = this.driver.controlFlow().execute(() => {
         return wdpromise.when(enabled).then((enabled: boolean) => {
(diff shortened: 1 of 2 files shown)


slotik:
@cnishina Do you think you might find the time to look at this within the following week? It would be super helpful if a fix for the issue got released in the near future.

slotik:
@wojtek1150 Could we merge this in? Is that something I can do?

rafalf:
when will this be available?

rafalf:
@cnishina any chance for a review and release soon?

lib/browser.ts
```diff
@@ -417,8 +417,8 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
    * Call waitForAngularEnabled() without passing a value to read the current
    * state without changing it.
    */
-  waitForAngularEnabled(enabled: boolean|wdpromise.Promise<boolean> = null):
-      wdpromise.Promise<boolean> {
+  waitForAngularEnabled(enabled: boolean|wdpromise.Promise<boolean> = null): boolean
```
qiyigg: What's the purpose of changing the return type? IMO, changing the return type will make user confusing and they don't know whether the function is async or not. If you want to give a clear/accurate type inference, you should use function overloading.

lib/runner.ts
```diff
@@ -290,7 +290,7 @@ export class Runner extends EventEmitter {
 
     browser_.forkNewDriverInstance =
         (useSameUrl: boolean, copyMockModules: boolean, copyConfigUpdates = true) => {
-          let newBrowser = this.createBrowser(plugins);
+          let newBrowser = this.createBrowser(plugins, browser_);
```
qiyigg: IMO, using the default config or the config of current browser for new browser, both of them make sense.
Creating browser based on parent browser just copy some init data from parent browser, I don't know why setting "waitForAngularEnabled" after creating the browser doesn't work in your case. Perhaps "waitForAngularEnabled" is async, you need wait for it? I am a little hesitate about this since I don't know whether some tests rely on the previous behavior and this change will break their tests.

PR #3868, opened 2016-12-22
fix(sauce): **breaking change** Remove redundant sauceAgent setting.

mgiambalvo:
Fixes #3694

PR #4419, opened 2017-08-01
fix(sauce): bring back sauceProxy as a configuration option

### LINKED ISSUE #4405: Proposal: Bring Back sauceProxy configuration
Bug report
---

- Node Version: `6.9.1`
- Protractor Version: `5.1.2`
- Angular Version: `^4.0.0`
- Browser(s): `All`
- Operating System and Version `Windows 7`
- Your protractor configuration file
```
const conf = require('./e2e/config/conf.e2e.json');
const { SpecReporter } = require('jasmine-spec-reporter');
const jasmineReporters = require('jasmine-reporters');

const prerun = { "background": true, "executable": "sauce-storage:auth_sauce.exe" };
const shartTestFiles = false;
const maxInstances = 1;

console.log("proxy is " + process.env.HTTP_PROXY + '--------------------------------------');

exports.config = {
  SELENIUM_PROMISE_MANAGER: 0,
  sauceBuild: process.env.JENKINS_BUILD_NUMBER || 'localTestBuildNumber__00',
  sauceUser: process.env.SAUCE_USERNAME || 'test',
  sauceKey: process.env.SAUCE_ACCESS_KEY || 'test',
  sauceSeleniumUseHttp: true,
  sauceSeleniumAddress: 'localhost:4445/wd/hub',
  webDriverProxy: process.env.HTTP_PROXY,
  multiCapabilities: [

    // {
    //   build,
    //   buildName,
    //   "name": "Windows 7 IE 11",
    //   "username": conf.sauceUser,
    //   "accessKey": conf.sauceKey,
    //   "tunnel-identifier": "shepherd-jenkins",
    //   "platform": "Windows 7",
    //   "browserName": "internet explorer",
    //   "version": "11",
    //   "avoid-proxy": true,
    //   "unexpectedAlertBehaviour": "ignore",
    //   maxInstances,
    //  shartTestFiles,
    //   prerun
    // },
    {
      count: 1,
      "name": "Windows 7 chrome 57",
      "tunnel-identifier": "shepherd-jenkins",
      "platform": "Windows 7",
      "browserName": "chrome",
      chromeOptions: {
        prefs: {
          'credentials_enable_service': false,
          'profile': {
            'password_manager_enabled': false
          }
        },
        args: [
          '--disable-cache',
          '--disable-application-cache',
          '--disable-offline-load-stale-cache',
          '--disk-cache-size=0',
          '--v8-cache-options=off'
        ]
      },
      "version": "57",
      shartTestFiles,
      prerun
    },
  ],
  allScriptsTimeout: 11000,
  seleniumArgs: ['--ignore_ssl'],
  specs: [
    './e2e/**/*.e2e-spec.ts',
  ],

  directConnect: true,
  baseUrl: 'my.url.com',
  framework: 'jasmine2',
  jasmineNodeOpts: {
    showColors: true,
    defaultTimeoutInterval: 30000,
    print: function () {
    }
  },

  onPrepare() {
    let capabilities
    require('ts-node').register({
      project: 'e2e/tsconfig.e2e.json'
    });

    jasmine.getEnv().addReporter(new SpecReporter({
      spec: { displayStacktrace: true }
    }));

    return browser.getProcessedConfig().then((c) => {
      capabilities = c.capabilities;
      // used to change the namespace for the data.
      process.env.CURRENT_BROWSER = c.capabilities.browserName.replace(new RegExp(' ', 'g'), '-');
      jasmine.getEnv().addReporter(
        new jasmineReporters.JUnitXmlReporter({
          filePrefix: c.capabilities.browserName,
          consolidateAll: true,
          modifySuiteName: function (generatedSuiteName, suite) {
            // this will produce distinct suite names for each capability,
            // e.g. 'firefox.login tests' and 'chrome.login tests'
            return c.capabilities.browserName + '.' + generatedSuiteName;
          }
        }));

      return browser.driver.get(c.baseUrl).then(function () {
        browser.sleep(2000);
      });
    });
  },

  onComplete() {
    return browser.getProcessedConfig().then(function (c) {
      return browser.getSession().then(function (session) {
        // required to be here so saucelabs picks up reports to put in jenkins
        console.log('SauceOnDemandSessionID=' + session.getId() + ' job-name=' + c.capabilities.name);
      });
    });
  }
};

```

- A relevant example test
Setup Saucelabs in a jenkins build using node version 6.2.1 (i had 6.9.1 on my local)

- Output from running the test
Basically what it boils down to is I need my proxy to make the SauceLabs call to update my tests however I'm running my tests in Jenkins.  When I run my tests, I get this problem

```
03:10:21 <div id="content">
03:10:21 <p>The following error was encountered while trying to retrieve the URL: <a href="http://myusername@localhost:4445/wd/hub/session">http://myusername@localhost:4445/wd/hub/session</a></p>
03:10:21 
03:10:21 <blockquote id="error">
03:10:21 <p><b>Access Denied.</b></p>
03:10:21 </blockquote>
```

Now I'm not sure if it was because the Jenkins job was running on NODE 6.2.1 or not but I do know it was usingthe webdriver proxy for my connection to the Sauce Selenium Relay but oddly was not including the saucekey.  Whenever i got access denied it always only showed myusername only.  I'm able using seleniumAddress and setting username / accesskey in each of my capabilities without setting webdriverproxy to run the tests.

Feature Request
---
My request would be to allow SauceLabs plugin to be configurable separate from the main testing connection like it used to be.  Right now there is no way for me to tell Protractor this beucase webDriverProxy is used for the driverProvider config + SauceLabs config.

It would be a breaking revert to what it previously was but it wouldn't be that big.  I'd be willing to looking into doing the PR when I have time.  I just don't know why sauceProxy was taken out because they are 2 separate usages of the proxy.  At the very least have some way that I could maybe slip in my own SauceLabs plugin as we do the sauceAgent.

As a part of a relavant thing it would be also nice if that sauce driverProvider covered console.log(session / job); as you see in my configuration so it would be a more complete integration with saucelabs. (specifically reporting for the jenkins plugin but it doesn't hurt anything anywhere else)

cnishina:
Reverts the change to PR #3868. Protractor should connect to Sauce Labs
via the sauceProxy. This is different than the webDriverProxy. The
webDriverProxy is tied to the browser session where we define where the
traffic will go through.

closes #4405

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -97,9 +97,14 @@ export interface Config {
    */
   seleniumSessionId?: string;
   /**
-   * The address of a proxy server to use for the connection to the
-   * Selenium Server. If not specified no proxy is configured. Looks like
-   * webDriverProxy: 'http://localhost:3128'
+   * The address of a proxy server to use for communicating to Sauce Labs rest APIs via the
+   * saucelabs node module. For example, the Sauce Labs Proxy can be setup with: sauceProxy:
+   * 'http://localhost:3128'
+   */
+  sauceProxy?: string;
+
+  /**
+   * The proxy address that browser traffic will go through which is tied to the browser session.
    */
   webDriverProxy?: string;
 
--- a/lib/driverProviders/sauce.ts
+++ b/lib/driverProviders/sauce.ts
@@ -58,7 +58,7 @@ export class Sauce extends DriverProvider {
       username: this.config_.sauceUser,
       password: this.config_.sauceKey,
       agent: this.config_.sauceAgent,
-      proxy: this.config_.webDriverProxy
+      proxy: this.config_.sauceProxy
     });
     this.config_.capabilities['username'] = this.config_.sauceUser;
     this.config_.capabilities['accessKey'] = this.config_.sauceKey;


cnishina:
@kenisteward could you review my comments and PR?

kenisteward:
The change looks good. In the comments I'd specify the proxy is used
specifically with the saucelabs library to make rest API calls for saucelabs

kenisteward:
Tha KS for this revision

PR #4428, opened 2017-08-05
deps(typescript): upgrade to 2.4.1

cnishina:


cnishina:
Try things and watch them fail!

PR #4462, opened 2017-08-28
added SELENIUM_PROMISE_MANAGER to the acceptable cli arguments

chrisgodsey:
This is to enable refactoring tests on a spec-file basis to utilize SELENIUM_PROMISE_MANAGER without having to duplicate the protractor.conf file.


chrisgodsey:
Updated...

chrisgodsey:
I think it's resolved now maybe...

PR #4463, opened 2017-08-29
Corrected compatibility

Itrulia:
Updated compatibility to reflect removal of `_debugger` which will break protractor for node v8 when using `browser.pause()`


Itrulia:
Duplocate of https://github.com/angular/protractor/pull/4345

PR #4467, opened 2017-08-29
added SELENIUM_PROMISE_MANAGER to the acceptable cli arguments

chrisgodsey:


chrisgodsey:
Added cla account

chrisgodsey:
Just to make it clear what value this adds, the intent is that it could be toggled on and off at the command line instead of in the protractor.conf so that it could be toggled off for some tests and not others - allowing an incremental refactor of tests that are currently dependent on the promise manager before it becomes fully phased out.

kylecordes:
How would this change work, though? The code one writes inside of a test is quite different with the promise manager turned off. Is there a use case for running the exact same test code, but then deciding at run time to flip this switch?

chrisgodsey:
So let me start by saying that I may not be 100% on the context that led me down this path as I opened the issue almost a year ago - but that being said the use case that led me to this, I believe, was:

We had a giant repository of hundreds (thousands?) of protractor tests that were all being governed by one configuration file, and we used command line arguments to dictate a certain subset of files (which we referred to as suites).

At some point we wanted to begin the refactor of converting these tests to no longer use selenium promise manager but didn't want to all-or-nothing the switch but instead selectively refactor tests over time to be compatible with the lack-of the selenium promise manager.

Trying to resolve this led me, if I recall correctly, down one of two paths - one was managing which configuration file was executed against - or flagging suites as having been refactored and turning the promise manager off if a flagged suite was encountered.

The first one would've involved a much bigger architecture effort on our part due to our CI process than the second which I got working pretty quickly.

PR #4488, opened 2017-09-14
add Perfecto lab as a data provider to Protractor 

UziEilon:
we have large number of customers that are using Protractor with our cloud.
Adding Perfecto as a data provider will simplify the execution workflow.
After adding these change we would like to add better interation with our reporting system (phase II)


wswebcreation:
Hi @UziEilon 

PR looks great, a nice addition. @cnishina can you take a look at this?

Issue #3952, opened 2017-01-10
Using --grep will call every defined beforeAll() in your suite/spec


wburgess-datical:
When using `--grep` to delineate individual tests across a suite of multiple specs, each defined `beforeAll`() will be called, even those belonging to specs that do not contain a captured test. This is troublesome as it adds significant setup time to a test run that is unnecessary, and can lead to workflow corruption. 

Note that superfluous `beforeEach()`s are not called (just the appropriate ones). Also note that this is directly related to issues experienced by users on https://github.com/angular/protractor/issues/164, but that has been closed without response.

Ideally for each test that is captured via `--grep`, it will look upwards and run the corresponding `beforeAll()` and `beforeEach()`, run the captured test, then run `afterEach()` and `afterAll()`, then move to the next captured test and repeat. This will avoid running "extra" `beforeAll()`s.

I really love the `--grep` functionality and would love to see it better supported! The `iit/xit/fit` paradigm requires manually editing potentially hundreds of tests across a ton of files, which is tedious and does not fit well into fully automated CI. `--grep` (plus a suite of official tags) is my current approach and typically works very well, except for this one inconvenience. 

Example:
``` javascript

spec01.js:

describe('Spec 1.1', function() {

    beforeAll(function() {
        console.log('1.1 beforeAll executed');
    });

    beforeEach(function() {
        console.log('1.1 beforeEach executed');
    });

    it('Test case 1.1.1', function() {
        console.log('1.1.1 test case executed');
    });

    it('Test case 1.1.2', function() {
        console.log('1.1.2 test case executed');
        expect(1).toEqual(1);
    });
});

describe('Spec 1.2', function() {

    beforeAll(function() {
        console.log('1.2 beforeAll executed');
    });

    it('Test case 1.2.1', function() {
        console.log('1.2.1 test case executed');
    });
});

-------------------------------------------------
spec02.js:


describe('Spec 2.1', function() {

    beforeAll(function() {
        console.log('2.1 beforeAll executed');
    });

    beforeEach(function() {
        console.log('2.1 beforeEach executed');
    });

    it('Test case 2.1.1 #RunThisTestOnly', function() {
        console.log('2.1.1 test case executed');
    });
});

```

Let's assume these files are captured in the "all" suite, and then run with: `protractor conf.js --suite=all --grep="#RunThisTestOnly"`

Output:
```
Spec started
1.1 beforeAll executed

1.2 beforeAll executed
2.1 beforeAll executed
[...]

PR #4506, opened 2017-09-26
fix(frameworks) specFilter should return false if spec not in filter

dsteinbach:
We override jasmine.env.specFilter function in protractor but it will always returns true while jasmine's implementation returns true or false:

```
this.env.specFilter = function(spec) {
  return specFilter.matches(spec.getFullName());
};
```

This is to resolve:
https://github.com/angular/protractor/issues/3952


dsteinbach:
I signed it!

vikerman:
Hi - Can you add a test for this?

PR #4515, opened 2017-10-01
Add support for regex in cssContainingText.

joeheyming:
In order to get this working, I had to serialize the regex before
we send it over the wire to the browser.

Since there is no standard way to do this, I took guidance from
a stackoverflow answer, where they call toString on the regex.
Then, on the browser, you use a regex to extract out the text in between
/someregex/

The hard part is to also extract out the modifiers, like i for ignore case.

Anyhow, this approach seems to work, and I have tests to prove it.

### DIFF
--- a/scripts/test.js
+++ b/scripts/test.js
@@ -5,7 +5,7 @@ var Executor = require('./test/test_util').Executor;
 
 var passingTests = [
   'node built/cli.js spec/basicConf.js',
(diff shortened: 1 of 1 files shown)


joeheyming:
I signed it!

joeheyming:
Link to Stackoverflow answer that I used: 
https://stackoverflow.com/a/33416684/1222743

joeheyming:
Why does it still say I have not signed the agreement?

PR #4528, opened 2017-10-05
chore(release): update selenium-webdriverjs and bump version number

vikerman:


qiyigg:
Need to investigate why "node node_modules/jasmine/bin/jasmine.js JASMINE_CONFIG_PATH=scripts/unit_test.json" is failed. (Not due to flaky, it will fail every time and won't fail before the change)

qiyigg:
Use #4538 instead.

Issue #1360, opened 2014-09-22
Support for regular expressions with cssContainingText


Droogans:
Right now I have some [pretty slow code](https://github.com/rackerlabs/encore-ui/blob/024e0297084583ea7d52303278465960b7f8f57f/src/rxBreadcrumbs/rxBreadcrumbs.page.js#L70-L82) to work around having multiple matches coming back for `cssContainingText`.

I'd really like to be able to something like this instead.

``` js
element(by.cssContainingText(/$Components^/));
```

jafd:
@omer727, this solution doesn't take into account that XPath expressions need escaping. And as of XPath 1.0, which is what most browsers use, this is a sad story (if your text contains both single and double quotes).

joeheyming:
I have an open pull request for this: https://github.com/angular/protractor/pull/4515

joeheyming:
Ping... It's been 29 days...

wswebcreation:
@joeheyming 

...pong..., I've added reviewers to the PR, sorry

PR #4532, opened 2017-10-08
feat(locators): Add support for regex in cssContainingText

joeheyming:
In order to get this working, I had to serialize the regex before
we send it over the wire to the browser.

Since there is no standard way to do this, I took guidance from
a stackoverflow answer, where they call toString on the regex.
Then, on the browser, you use a regex to extract out the text in between
/someregex/

The hard part is to also extract out the modifiers, like i for ignore case.

Anyhow, this approach seems to work, and I have tests to prove it.

This is related to #1360

### DIFF
--- a/lib/clientsidescripts.js
+++ b/lib/clientsidescripts.js
@@ -676,7 +676,7 @@ functions.findByPartialButtonText = function(searchText, using) {
  * Find elements by css selector and textual content.
  *
  * @param {string} cssSelector The css selector to match.
- * @param {string} searchText The exact text to match.  Can be a serialized regex.
+ * @param {string} searchText The exact text to match or a serialized regex.
  * @param {Element} using The scope of the search.
  *
  * @return {Array.<Element>} An array of matching elements.
--- a/lib/clientsidescripts.js
+++ b/lib/clientsidescripts.js
@@ -693,9 +693,11 @@ functions.findByCssContainingText = function(cssSelector, searchText, using) {
   for (var i = 0; i < elements.length; ++i) {
     var element = elements[i];
     var elementText = element.textContent || element.innerText || '';
+    var elementMatches = searchText instanceof RegExp ?
+        searchText.test(elementText) :
+        elementText.indexOf(searchText) > -1;
     
-    if ((searchText instanceof RegExp && searchText.test(elementText)) ||
-        elementText.indexOf(searchText) > -1) {
+    if (elementMatches) {
       matches.push(element);
     }
   }
--- a/lib/clientsidescripts.js
+++ b/lib/clientsidescripts.js
@@ -676,7 +676,7 @@ functions.findByPartialButtonText = function(searchText, using) {
  * Find elements by css selector and textual content.
  *
  * @param {string} cssSelector The css selector to match.
- * @param {string} searchText The exact text to match.
+ * @param {string} searchText The exact text to match.  Can be a serialized regex.
  * @param {Element} using The scope of the search.
  *
  * @return {Array.<Element>} An array of matching elements.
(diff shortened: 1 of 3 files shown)


joeheyming:
I looked at the continuous test failure, and the failures are unrelated to my changes.
I touched cssContainingText.
The test that failed, protractor/spec/ts/basic/element_spec.ts:177:3)
has nothing to do with cssContaining text.  It is trying to fill out a form

joeheyming:
Not sure what to do with this:

running: node scripts/interactive_tests/with_base_url.js
.
/home/ubuntu/protractor/node_modules/q/q.js:155
                throw e;
                ^
Did not start interactive server in 10s. Server startup output: [19:50:49] I/local - Starting selenium standalone server...
[19:50:51] I/local - Selenium standalone server started at http://172.17.197.126:44518/wd/hub


>>> fail: Error: expecting exit code: 0, actual: 1
running: node node_modules/jasmine/bin/jasmine.js JASMINE_CONFIG_PATH=scripts/unit_test.json

joeheyming:
Hey so... Ping.... echo... echo... echo... echoo........

wswebcreation:
Hi @joeheyming 

Sorry for not reacting. I'll add someone from the core team to review this PR.

lib/clientsidescripts.js
```diff
@@ -676,20 +676,26 @@ functions.findByPartialButtonText = function(searchText, using) {
...
+ * @param {string} searchText The exact text to match.  Can be a serialized regex.
  * @param {Element} using The scope of the search.
  *
  * @return {Array.<Element>} An array of matching elements.
  */
 functions.findByCssContainingText = function(cssSelector, searchText, using) {
   using = using || document;
 
+  if (searchText.indexOf('__REGEXP__') === 0) {
+    var match = searchText.split('__REGEXP__')[1].match(/\/(.*)\/(.*)?/);
```
qiyigg: I am not sure enough whether this match method works well here

Would you mind adding another test that using a regex with no regex flags, we can be more convinced of this change

lib/clientsidescripts.js
```diff
@@ -676,20 +676,26 @@ functions.findByPartialButtonText = function(searchText, using) {
  * Find elements by css selector and textual content.
  *
  * @param {string} cssSelector The css selector to match.
- * @param {string} searchText The exact text to match.
+ * @param {string} searchText The exact text to match.  Can be a serialized regex.
```
qiyigg: Would it be better if we say: searchText The exact text to match or a serialized regex.

lib/clientsidescripts.js
```diff
@@ -676,20 +676,26 @@ functions.findByPartialButtonText = function(searchText, using) {
...
+  }
   var elements = using.querySelectorAll(cssSelector);
   var matches = [];
   for (var i = 0; i < elements.length; ++i) {
     var element = elements[i];
     var elementText = element.textContent || element.innerText || '';
-    if (elementText.indexOf(searchText) > -1) {
+    
+    if ((searchText instanceof RegExp && searchText.test(elementText)) ||
+        elementText.indexOf(searchText) > -1) {
```
qiyigg: My understanding is "elementText.indexOf(searchText) > -1" should only be applied to "searchText is not a RegExp"
For now, if elementText doesn't satisfy searchText regex, "elementText.indexOf(searchText) > -1" will still be called

lib/clientsidescripts.js
joeheyming: sure

lib/clientsidescripts.js
joeheyming: There are no side-effects to calling indexOf.  The way || works is that javascript will take the first true value.
So I don't think indexOf will be called.

Here is an example I tried out:
x = () => { console.log('here'); return true; }
y = () { console.log('there'); return false; }

z = x() || y()
here
true

lib/clientsidescripts.js
joeheyming: ok, will do

joeheyming:
I did a force push to my branch on my fork.  Does github do well with that?

lib/clientsidescripts.js
joeheyming: Ok, I tried to address this issue.  Let me know if you like the updated code.

joeheyming:
The last CircleCi failure was a pip install failure.  Nothing to see here

PR #4538, opened 2017-10-11
chore(release): update selenium-webdriverjs and bump version number

qiyigg:
1.Update selenium-webdriverjs and bump version number
2.Solve compatibility problem: new webdriver has a more strict method when setting script timeout. If no script timeout was specified, set timeout to 0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "optimist": "~0.6.0",
     "q": "1.4.1",
     "saucelabs": "~1.3.0",
-    "selenium-webdriver": "3.0.1",
+    "selenium-webdriver": "3.6.0",
     "source-map-support": "~0.4.0",
     "webdriver-js-extender": "^1.0.0",
     "webdriver-manager": "^12.0.6"
@@ -81,5 +81,5 @@
   "engines": {
     "node": ">=6.9.x"
   },
-  "version": "5.1.2"
+  "version": "5.2.0"
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "optimist": "~0.6.0",
     "q": "1.4.1",
     "saucelabs": "~1.3.0",
-    "selenium-webdriver": "3.0.1",
+    "selenium-webdriver": "3.5.0",
```
wswebcreation: Hi @vikerman @qiyigg 

Don't we want to go to 3.6.0 that's released a few weeks ago?

qiyigg: It there any new features in 3.6.0 that you rely on?

wswebcreation: I was trying to search for the post of Simon Stewart but I couldn't find it. He said they didn't update the changelog (https://github.com/SeleniumHQ/selenium/blob/selenium-3.6.0/javascript/node/selenium-webdriver/CHANGES.md#vnext) but it was wise to update to 3.6.0 (on of the last versions before they are going to release version 4.0)

So if it's not to much effort I would advice to go to version 3.6.0

qiyigg: OK, I will try it right now. If it is easy to upgrade to 3.6, I will upgrade protractor to 3.6 directly; otherwise, I 'll do it in a separate PR later on.

Issue #4543, opened 2017-10-18
waitForAngular with Angular2 sometimes causes "Error while waiting for Protractor to sync with the page: true"


andrewpmontgomery:
Node Version: 6.11.1
Protractor Version: 5.2.0
Angular Version: 4.4.5
Browser: Chrome 60
OS: Windows 7

Following on from the comment by @laurelnaiad here: https://github.com/angular/protractor/issues/3611#issuecomment-256192364

This is a very simple issue so I'll try to create a pull request myself.

Here is file `clientsidescripts.js` line 181:
`return testability.whenStable(testCallback);`

The Angular method `testability.whenStable` will call `testCallback` with a boolean argument indicating if any work was done to stabilize the system (true), or if the system was already stable (false).
Most of the time, the page we're testing is already stable, so `whenStable` simply calls `testCallback(false)`, which works fine.
But sometimes the page isn't stable - for example if we're waiting for a setTimeout - so it calls `testCallback(true)`.

In `browser.ts` lines 650-657, the callback is handled like this:
```
    return runWaitForAngularScript()
        .then((browserErr: Function) => {
            if (browserErr) {
                throw new Error(
                    'Error while waiting for Protractor to ' +
                    'sync with the page: ' + JSON.stringify(browserErr));
            }
        })
```

Because browserErr is being passed as true, we get this failure message:
_Error while waiting for Protractor to sync with the page: true_

I'll create a pull request to replace clientsidescripts.js line 181 with this:
`return testability.whenStable(function() { testCallback(); });`

qiyigg:
PR merged

PR #4544, opened 2017-10-18
Update clientsidescripts.js to fix waitForAngular where whenStable calls back true

andrewpmontgomery:
Fix for issue #4543


andrewpmontgomery:
I signed it!

PR #4549, opened 2017-10-19
fix(element): correct typing of method map()

### LINKED ISSUE #4548: ElementArrayFinder.map typing shows element as possibly undefined
Bug report
---

- TypeScript Version: `2.5.2`

In the current typing of method ElementArrayFinder.map(), the `elementFinder` parameter of the mapping function is [marked as possible undefined](https://github.com/angular/protractor/blob/5.2.0/lib/element.ts#L653). However, following the code upwards, each element of the array being mapped is produced with [`ElementFinder.fromWebElement_`](https://github.com/angular/protractor/blob/5.2.0/lib/element.ts#L520) which can't produce undefined values.

Using these types, client code has to assert that the value to map can't be undefined. E.g.:

```ts
element.all('p').map(paragraph => new Paragraph(paragraph!));
```

This exclamation mark is unnecessary if the types are corrected.

RomkeVdMeulen:
The elementFinder and index arguments of the mapping function passed to ElementArrayFinder.map() will never be undefined.

Fixes #4548

### DIFF
--- a/lib/element.ts
+++ b/lib/element.ts
@@ -650,10 +650,10 @@ export class ElementArrayFinder extends WebdriverWebElement {
    * @returns {!webdriver.promise.Promise} A promise that resolves to an array
    *     of values returned by the map function.
    */
-  map<T>(mapFn: (elementFinder?: ElementFinder, index?: number) => T | any):
+  map<T>(mapFn: (elementFinder: ElementFinder, index: number) => T | any):
       wdpromise.Promise<T[]> {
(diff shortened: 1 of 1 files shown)


RomkeVdMeulen:
I signed it!

JacobPatenaude-AI:
I'm not sure if this PR is still active, but should [each](https://github.com/angular/protractor/blob/d7cddc56662af8a0665b1f298e65d22c92359f3a/lib/element.ts#L528) also have its typings changed?

Issue #2973, opened 2016-02-25
getVal for input


philn5d:
from the FAQ.md

"The result of  getText  from an input element is always empty

This is a webdriver quirk.  <input>  and  <textarea>  elements always have empty  getText  values. Instead, try  element.getAttribute('value') ."

Why not make a getValue() or getVal()? Or map getText() if the input type is input or textarea?

I know it seems trivial to do so, but there are maybe some reasons?

Suggest modeling getVal() after [JQuery val()](http://api.jquery.com/val/), maybe I'm not the first to make this suggestion, but it didn't seem to come up in an issues search.

macroking:
After going through the Protractor & selenium-webdriver libraries, found that  we can add getVal() function in  webdriver.js and update the function name in WEB_ELEMENT_FUNCTIONS array in element.js file.

Not sure whether I have to write test for that. I'll work on this and submit a PR.

juliemr:
Seems reasonable, we'd look at a PR for this.

PR #4552, opened 2017-10-20
feat(element): getText returns text for input fields

viktorpergjoka:
feat(element): getText returns text for input fields

(#2973) element.getText() returns now also the text for input fields

### DIFF
--- a/lib/element.ts
+++ b/lib/element.ts
@@ -14,9 +14,8 @@ export class WebdriverWebElement {}
 export interface WebdriverWebElement extends WebElement { [key: string]: any; }
 
 let WEB_ELEMENT_FUNCTIONS = [
-  'click', 'sendKeys', 'getTagName', 'getCssValue', 'getAttribute', 'getText', 'getSize',
-  'getLocation', 'isEnabled', 'isSelected', 'submit', 'clear', 'isDisplayed', 'getId',
-  'takeScreenshot'
+  'click', 'sendKeys', 'getTagName', 'getCssValue', 'getAttribute', 'getSize', 'getLocation',
+  'isEnabled', 'isSelected', 'submit', 'clear', 'isDisplayed', 'getId', 'takeScreenshot'
 ];
 
 /**
@@ -434,7 +433,7 @@ export class ElementArrayFinder extends WebdriverWebElement {
   /**
    * Returns true if there are any elements present that match the finder.
    *
-   * @alias element.all(locator).isPresent()
+   * @alias element.all(locator).isnpm run formatPresent()
    *
    * @example
    * expect($('.item').isPresent()).toBeTruthy();
@@ -444,6 +443,14 @@ export class ElementArrayFinder extends WebdriverWebElement {
   isPresent(): wdpromise.Promise<boolean> {
     return this.count().then((count) => {
       return count > 0;
+    });
+  }
+
+  getText(): wdpromise.Promise<any> {
+    return this.asElementFinders_().then((parentWebElements: WebElement[]) => {
+      let childList =
+          parentWebElements.map((parentWebElement: WebElement) => parentWebElement.getText());
+      return wdpromise.all(childList);
     });
   }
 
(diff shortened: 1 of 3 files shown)


lib/element.ts
```diff
@@ -447,6 +446,14 @@ export class ElementArrayFinder extends WebdriverWebElement {
     });
   }
 
+  getText(): wdpromise.Promise<any> {
```
MrViktor: I had to return here "any" because in the interface it's defined as string, but string[] would be needed here.
It is also not documented that ElementArrayFinder getText() is available. But there are tests where getText is used on ElementArrayFinder.

lib/element.ts
```diff
@@ -434,7 +433,7 @@ export class ElementArrayFinder extends WebdriverWebElement {
   /**
    * Returns true if there are any elements present that match the finder.
    *
-   * @alias element.all(locator).isPresent()
+   * @alias element.all(locator).isnpm run formatPresent()
```
qiyigg: typo?

lib/element.ts
qiyigg: It doesn't quite make sense for me to have getText() for ElementArrayFinder
As you said, there are tests where getText is used on ElementArrayFinder, what are the tests that you mentioned?

wswebcreation: @qiyigg 

I must say that I use `getText()` a lot on `ElementArrayFinder`, it is possible, so a fix on the typings would be great.
What @viktorpergjoka means is also the amount of examples in this files that `getText()` is used on `ElementArrayFinder`'s

lib/element.ts
```diff
@@ -1166,6 +1173,34 @@ export class ElementFinder extends WebdriverWebElement {
...
+   * expect(foo.getText()).toEqual('Inner text');
+   *
+   * @returns {!webdriver.promise.Promise.<string>} A promise that will be
+   *     resolved with the element's visible text.
+   */
+  getText(): wdpromise.Promise<string> {
+    let webElem = this.getWebElement();
+    return webElem.getText().then((text) => {
+      if (text) {
+        return text.trim();
```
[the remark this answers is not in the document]
viktorpergjoka: yes I thought that this is a little dangerous. I created this here because of IE or Edge who sometimes appends whitespaces on the text (my previous commit failed because of this).

viktorpergjoka:
yes that is true. getText() behaves correctly regarding the specs. I just saw that everyone who is starting to use getText for the first time will wonder why it doesn't return the text.
Maybe I suggest then not to merge this PR, because as you said getAttribute('value') can easily be called and is not painful at all, so that element.ts doesn't get blown up with functions.

lib/element.ts
viktorpergjoka: Regarding the tests for example in elements_spec.js:
describe('ElementArrayFinder', function() {

  it('action should act on all elements', function() {
    browser.get('index.html#/conflict');

    var multiElement = element.all(by.binding('item.reusedBinding'));
    expect(multiElement.getText()).toEqual(['Outer: outer', 'Inner: inner']);
  });

lib/element.ts
viktorpergjoka: oh damn yes sorry

lib/element.ts
qiyigg: oh, I see. The function name is a little misleading.

lib/element.ts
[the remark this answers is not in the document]
qiyigg: yes, I think we should try not to change the original behavior.
For now, we can "correct" our smoke test by trimming the actual result in test, and create an issue/bug to change it back after Microsoft fixed that bug.

qiyigg:
From html point of view, getText() return empty for input/textarea makes sense
For user who is writing testcases may not be familiar with html, it might surprise and block them for a while.

But, yes, I think we should not change the original behavior unless it is necessary or the original one is problematic. Users may already rely on it. e.g. In user test, they may already assume input.getText() is empty and use as an expect value when checking whether a group of texts are correct or not. expect(elementArray.getText()).equalTo(["jane", "123" ,"" -> input tag,  "google"]) 

But I think we should still think a better way to guide user to use getAttribute('value')  to handle input/textarea

viktorpergjoka:
Its already stated in the FAQ under the section
"The result of getText from an input element is always empty"

qiyigg:
Yes, that's great, but you know, we have many docs and user tends to not read docs... 
I checked our tutorial, maybe we can add an example to use getAttribute()(ideally, use getAttribute("value") to handle input/textarea) in there and add some comments to explain this.

viktorpergjoka:
I reverted my changes an added a new PR where I added an example in the tutorial. See #4566

PR #4554, opened 2017-10-23
fix(direct): Use config's geckoDriver when specified

shaungrady:
This change makes the `firefox` capability more closely match `chrome`'s. The `firefox` capability was not looking for `config_.geckoDriver` like `chrome` was, making the `geckoDriver` configuration useless for directConnect and always resulting in an error if the driver wasn't installed in the default location.

### DIFF
--- a/lib/driverProviders/direct.ts
+++ b/lib/driverProviders/direct.ts
@@ -91,15 +91,27 @@ export class Direct extends DriverProvider {
         break;
       case 'firefox':
         let geckoDriverFile: string;
-        try {
-          let updateJson = path.resolve(SeleniumConfig.getSeleniumDir(), 'update-config.json');
-          let updateConfig = JSON.parse(fs.readFileSync(updateJson).toString());
-          geckoDriverFile = updateConfig.gecko.last;
-        } catch (e) {
-          throw new BrowserError(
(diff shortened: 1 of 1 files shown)


shaungrady:
I signed it!

talolierwork:
Hitting that problem, waiting for that fix...


BTW, we wrote a small script which copies the gecko driver from the protractor node modeules.
We run it in the "pree2e": "./update_webdriver.sh" (in the package.json)
Here is the script:
# Download selenium drivers
./node_modules/protractor/node_modules/webdriver-manager/bin/webdriver-manager update --versions.standalone 3.6.0

# Copy the Gecko driver 
driver_directory=./node_modules/protractor/node_modules/webdriver-manager/selenium
echo Looking for Gecko driver in: $driver_directory
filename=$(ls $driver_directory | grep '.*geckodriver.*[0-9]$' | sort | tail -1)
echo Lastest gecko driver found: $filename
cp $driver_directory/$filename ./geckodriver
echo Driver copied.

shaungrady:
Yeah we did the same thing to work around it.

shaungrady:
@juliemr This is a real simple fix. Can you—or someone else I should ping—take a look at this please?

PR #4555, opened 2017-10-24
Add docs for debugging with the control flow disabled

theRobinator:
When the control flow is disabled, `browser.pause()`, `browser.debugger()`, and `browser.enterRepl()` all crash. I couldn't find this documented anywhere other than bug reports, so I've added it to the docs on debugging.


theRobinator:
CLA signed.

wswebcreation:
@qiyigg / @vikerman 

Can we merge this low impact PR

PR #4563, opened 2017-10-27
Update selenium-webdriver typings version

shoom3301:
@types/selenium-webdriver@2.53.39 is outdated.
For example:
With old typings `browser.manage().addCookie('key', 'value')`
In real `browser.manage().addCookie({name: 'foo', value: 'bar'})`


qiyigg:
This was fixed by #4882

PR #4566, opened 2017-10-28
doc(tutorial): added example for element.getAttribute('value') to read text from an input

viktorpergjoka:
doc(tutorial): added example for element.getAttribute('value') to read text from an input

Added in tutorial an example how to read the text of an input field with using element.getAttribute('value').

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -153,6 +153,7 @@ describe('Protractor Demo App', function() {
 ```
 
 Here, we've pulled the navigation out into a `beforeEach` function which is run before every `it` block. We've also stored the ElementFinders for the first and second input in nice variables that can be reused. Fill out the second test using those variables, and run the tests again to ensure they pass.
+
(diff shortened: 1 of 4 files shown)


docs/tutorial.md
```diff
@@ -144,11 +144,18 @@ describe('Protractor Demo App', function() {
...
+  it('should read the value from an input', function() {
+    firstNumber.sendKeys(1);
+    expect(firstNumber.getAttribute('value')).toEqual('1');
+  });
 });
 ```
 
 Here, we've pulled the navigation out into a `beforeEach` function which is run before every `it` block. We've also stored the ElementFinders for the first and second input in nice variables that can be reused. Fill out the second test using those variables, and run the tests again to ensure they pass.
 
+In the last assertion we set a value in the input field `firstNumber` and then read it and compare the value. Input fields does not have text, instead they have values. To read a value from an input, we cannot therefore use `firstNumber.getText()`. Instead we can obtain the value with `firstNumber.getAttribute('value')`.
```
qiyigg: The example is great, just a little problem, could you fix the grammar? 
Input fields does(do) not have text(s), instead they have values.

qiyigg:
Thanks for adding the new example. 
The example looks good to me, just a minor grammar problem and it would be better to squash the comments into one comment and remove irrelevant contents

viktorpergjoka:
Changed. I hope now its fine

PR #4594, opened 2017-11-20
Add space after full stop in error message.

iphydf:
Linkifiers interpret the ".If" at the end of the URL as part of the URL.


qiyigg:
Could you please update the commit message according to the guideline:
https://github.com/angular/protractor/blob/master/CONTRIBUTING.md
Thanks!

PR #4608, opened 2017-12-01
fix(Jasmine): Update Jasmine to support Node8 async/await

qiyigg:
fix(jasmine): Update Jasmine to support Node8 async/await
Breaking change for TypeScript:

JasmineWD doesn't know anything about async/await, turns off JasmineWD if control flow was disabled.

It will affect TypeScript tests that are using async/await and
1. miss some await keyword in the test.(Previously, this might cause the
test failed silently and be reported as pass), or
2. use Promise in jasmine expect function

Before
```ts
await expect(getPromise()).toEqual(42);
```

After
```ts
expect(await getPromise()).toEqual(42);
```

### DIFF
--- a/lib/frameworks/jasmine.js
+++ b/lib/frameworks/jasmine.js
@@ -63,7 +63,11 @@ exports.run = function(runner, specs) {
   var jrunner = new JasmineRunner();
   /* global jasmine */
 
-  require('jasminewd2').init(webdriver.promise.controlFlow(), webdriver);
+  // Don't even require JasmineWD if the control
+  // flow is turned off.
+  if (webdriver.promise.USE_PROMISE_MANAGER) {
+    require('jasminewd2').init(webdriver.promise.controlFlow(), webdriver);
+  }
 
   var jasmineNodeOpts = runner.getConfig().jasmineNodeOpts;
 
--- a/spec/ts/basic/element_spec.ts
+++ b/spec/ts/basic/element_spec.ts
@@ -8,8 +8,8 @@ describe('ElementFinder', function() {
     await browser.get('index.html#/form');
     const nameByElement = element(by.binding('username'));
 
-    await expect(nameByElement.getText())
-        .toEqual(browser.findElement(by.binding('username')).getText());
+    expect(await nameByElement.getText())
+        .toEqual(await browser.findElement(by.binding('username')).getText());
   });
 
   it('should wait to grab the WebElement until a method is called', async function() {
@@ -19,11 +19,11 @@ describe('ElementFinder', function() {
 
     await browser.get('index.html#/form');
 
-    await expect(name.getText()).toEqual('Anon');
+    expect(await name.getText()).toEqual('Anon');
 
     await usernameInput.clear();
     await usernameInput.sendKeys('Jane');
-    await expect(name.getText()).toEqual('Jane');
+    expect(await name.getText()).toEqual('Jane');
   });
 
   it('should chain element actions', async function() {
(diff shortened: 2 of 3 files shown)


juliemr:
Code LGTM! I'd suggest updating the commit message to have a clear 'before/after' for breaking changes, like this: https://github.com/angular/protractor/commit/5034c89242794dd14aba294ba3468937e06a7b69

qiyigg:
Sure, will squash and rewrite the comment after fixing all breaks.

rdbaron:
Just to clarify:

"It will affect TypeScript tests that are using async/await and"

It will affect ANY tests that turn off the control flow, not just TypeScript tests, right?

Out of curiosity, what was broken?  I recently migrated to async/await (I don't use TypeScript though) and my expect statements have been working as expected using:

`await expect(getPromise()).toEqual(42)`

qiyigg:
It will affect ANY tests that turn off the control flow AND using async/await
Only TypeScript supports async/await before this change (you cannot use javascript async/await before this change); therefore it will only affect/break TypeScript

For JavaScript, if you write test like this : await expect(getPromise()).toEqual(42). it should throw an ugly error.
If you don't get that error, maybe
1. you forgot to turn off control flow
2. your getPromise() doesn't really return a promise.

rdbaron:
I'm sure I'm missing something obvious here, but I seem to have my 2000+ line repo running without issue using Protractor 5.2.0 (prior to this commit).

I am using `SELENIUM_PROMISE_MANAGER: false`

Here is a trivial example that *seems* to show everything working as I would expect:

```
let angularBanner = element(by.css('.hero')).$('h2')
let todoList = element.all(by.repeater('todo in todoList.todos'))

describe('async/await should', () => {
  it('work when Promises resolve', async () => {
    await browser.get('https://angularjs.org')
    await expect(angularBanner.getText()).toEqual('AngularJS')
    return expect(todoList.count()).toBe(2)
  })

  it('fail when expectations are incorrect', async () => {
    await expect(angularBanner.getText()).toEqual('SomeJunk')
    return expect(todoList.count()).toBe(0)
  })
})
```

renehamburger:
@qiyigg, would you mind explaining briefly which issue this commit fixes? We've also been making use of the async expect as in `await expect(getPromise()).toEqual(42);` throughout, as it made our tests more readable.

qiyigg:
@rdbaron hmm, that's also quite surprising for me. I double checked the JasmineWD github and found jasmineWD has already supported async/await in the new release, but we didn't have that change inside google and therefore we didn't notice async/await can be used directly externally.

@renehamburger the original idea was to disable jasmineWD when disable control flow since previously jasmineWD didn't support  async/await and we thought jasmineWD shouldn't be existed if we don't use control flow. JasmineWD change the original behavior of Jasmine expect, so that it can accept promise.

"await expect(getPromise()).toEqual(42);" looks more readable: My personally understanding is when we make an assertion using "expect(a).toEqual(b)", a and b should be comparable. Only the process returns promise should be awaited. e.g. we don't actually wait for "button = element(by.css('button'))" (the locator is lazy loaded), but wait for button.getText() / button.click().

But anyway, this change was unnecessary at this point.
For now, we just revert the change and release a new version.

Sorry for the inconvenience.

renehamburger:
Thanks for the quick fix, @qiyigg!

rdbaron:
Thanks for the reply, @qiyigg.

Can you just clarify your previous comment for me though.  

> Only TypeScript supports async/await before this change (you cannot use javascript async/await before this change)

I am not using TypeScript and things seem to work.  Is there something specific I should be aware of?

PR #4613, opened 2017-12-06
fix(doc): update doc for testing with node 8 async/await and chrome inspector.

qiyigg:
inspector.


wswebcreation:
Tnx !!

vsravuri:
@qiyigg @wswebcreation 

Appreciate if an example is provided for Page Objects with Node 8 async/await

https://github.com/angular/protractor/blob/master/docs/page-objects.md

qiyigg:
Hi, 
For me, there's no difference in writing async/await test when using or not using Page Objects. Page Object is just a user wrapper to make your test more clear and stable.
But yes, it might have some problems if user doesn't understand async/await when they write tests using Page Objects.
Would you mind creating an issue, describe the requirement, and see if someone would like to write this? I don't think I could do it recently.
Thanks!

vsravuri:
Sure, i will create a an issue.

vsravuri:
@qiyigg 

I have opened a new issue for documentation of Page Objects.

https://github.com/angular/protractor/issues/4637

qiyigg:
Thanks!

Issue #4634, opened 2017-12-18
Wrong typings for `each()` method


inad9300:
In Protractor 5.2.2 (and in previous versions too), the typings for the `each()` method of the `ElementArrayFinder` class are defined like so:

```typescript
each(fn: (elementFinder?: ElementFinder, index?: number) => any): wdpromise.Promise<any>;
```

This is a misunderstanding of TypeScript `?` modifier. What you are saying with such a type for `fn` is that `fn` may receive 0 arguments, 1 argument or 2 arguments depending on the case. What you want to represent isn't this – `each()` will every single time receive both `elementFinder` and `index`.

This kind of errors are now becoming evident thanks to TypeScript's new [`--strictFunctionTypes` flag](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-6.html#strict-function-types), causing a lot of compilation problems.

Many other functions (e.g. `map()`) contain the same mistake.

wswebcreation:
@inad9300 , care for a PR ;-)?

inad9300:
@wswebcreation Done ;)

PR #4671, opened 2018-01-20
Fix typings for several ElementArrayFinder methods

inad9300:
Some of the callback functions had an incorrect `?` for their arguments.

Done for [#4634|https://github.com/angular/protractor/issues/4634].

### DIFF
--- a/lib/element.ts
+++ b/lib/element.ts
@@ -230,7 +230,7 @@ export class ElementArrayFinder extends WebdriverWebElement {
    *     of element that satisfy the filter function.
    */
   filter(
-      filterFn: (element: ElementFinder, index?: number) => boolean |
+      filterFn: (element: ElementFinder, index: number) => boolean |
           wdpromise.Promise<boolean>): ElementArrayFinder {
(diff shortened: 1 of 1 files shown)


inad9300:
I've signed the agreement thing... Am I now an official non-paid Google employee?

lib/element.ts
```diff
@@ -230,7 +230,7 @@ export class ElementArrayFinder extends WebdriverWebElement {
    *     of element that satisfy the filter function.
    */
   filter(
-      filterFn: (element: ElementFinder, index?: number) => boolean |
+      filterFn: (element: ElementFinder, index: number) => boolean |
```
tymfear: Why does the `index` parameter should be required?

inad9300: Removing the question mark does not mean that you are forced to accept `index` as a parameter in your callback. This is the whole point of https://github.com/angular/protractor/issues/4634.

tymfear: Gotcha! Thanks a lot!

PR #4682, opened 2018-01-29
Fix typo about Angular 2 compatibility

monkpit:
A recent change in the Readme deleted wording that said `Angular 2`. It makes it sound like no version of Angular will support the `binding` and `model` locators, which is not true. We should re-add the "2" back to make this more clear.

Original PR that introduced the issue: https://github.com/angular/protractor/commit/af6afa646cacf5dd2f023561f84429c05f508b77#diff-04c6e90faac2675aa89e2176d2eec7d8

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Compatibility
 
 Protractor 5 is compatible with nodejs v6 and newer.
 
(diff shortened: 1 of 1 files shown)


monkpit:
Bleh - author was set to my work email. I'll recreate this PR

PR #4683, opened 2018-01-29
Changes "Angular" to "Angular 2"

monkpit:
A recent change in the Readme deleted wording that said `Angular 2`. It makes it sound like no version of Angular will support the `binding` and `model` locators, which is not true. We should re-add the "2" back to make this more clear.

Diff that introduced the issue: https://github.com/angular/protractor/commit/af6afa646cacf5dd2f023561f84429c05f508b77#diff-04c6e90faac2675aa89e2176d2eec7d8

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Compatibility
 
 Protractor 5 is compatible with nodejs v6 and newer.
 
-Protractor works with AngularJS versions greater than 1.0.6/1.1.4, and is compatible with Angular applications. Note that for Angular apps, the `binding` and `model` locators are not supported. We recommend using `by.css`.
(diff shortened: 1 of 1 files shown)


michelled66:
Does this mean that Angular 4 DOES support the binding and model locators?
Because my understanding is that there's AngularJS (which refers to Angular 1) and Angular (which refers to Angular 2+) which would have been why the ReadMe would have the distinction of AngularJS v. Angular and why the "2" was removed.

qiyigg:
Oh, my bad, I didn't take a look at the whole file. It has already mentioned "angular" in the first section. We should make it consistent in the whole page. 
Do you agree with that @monkpit ?

monkpit:
I was not aware of the distinction between Angular and AngularJS. I had just assumed the removal of the 2 was accidental.

qiyigg:
ok. so I will rollback this change. 
Thanks for pointing out! @michelled66

michelled66:
Happy to help!

PR #4716, opened 2018-02-23
Updates selenium webdriver in package.json

ns-cwagenaar:
Updates the version of Selenium Webdriver given in package.json to be consistent with that in config.json


ns-cwagenaar:
Meant to make this on NarrativeScience/protractor, sorry!

PR #4732, opened 2018-03-14
docs(page_objects): Remove unnecessary "return await"

### LINKED ISSUE #4730: Return await statement is unnecessary in protractor examples.
Followed by ESLint rule 

> Inside an async function, return await is useless. Since the return value of an async function is always wrapped in Promise.resolve, return await doesn’t actually do anything except add extra time before the overarching Promise resolves or rejects. This pattern is almost certainly due to programmer ignorance of the return semantics of async functions

I found it in page-objects.md code 
```
async function() {
    return await greeting.getText();
  };
```

jonhermansen:
Fixes #4730 

More info here: https://eslint.org/docs/rules/no-return-await

### DIFF
--- a/docs/page-objects.md
+++ b/docs/page-objects.md
@@ -61,7 +61,7 @@ var AngularHomepage = function() {
   };
 
   this.getGreetingText = async function() {
-    return await greeting.getText();
+    return greeting.getText();
   };
 
   // Not async, returns the element
--- a/docs/page-objects.md
+++ b/docs/page-objects.md
@@ -60,7 +60,7 @@ var AngularHomepage = function() {
     await nameInput.sendKeys(name);
   };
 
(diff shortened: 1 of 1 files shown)


jonhermansen:
I signed it!

qiyigg:
@marcincharezinski agreed, maybe we can remove the async modifier as well. @joeheyming what do you think?

jonhermansen:
@qiyigg I just pushed a commit to make getGreetingText synchronous as well. Makes sense to me. The async/await version of the code prefixes the function call with `await`, and Jasmine's expect() is able to resolve the promise anyway in the non async/await version. I agree that the function doesn't need to be async now +1

joeheyming:
works for me

PR #4755, opened 2018-03-31
- added addLocation function usage example with config file(for using with mocha framework)

IgorSasovets:
I had a problem with using of addLocator function. Tests were written using JavaScript. Example from [Protractor API](https://www.protractortest.org/#/api?view=ProtractorBy.prototype.addLocator) not worked for me. So, I decided to add small example(including config file) to help other engineers and save their time.

### DIFF
--- a/example/addLocatorExample/addLocator_usage_example_spec.js
+++ b/example/addLocatorExample/addLocator_usage_example_spec.js
@@ -1,0 +1,9 @@
+describe('angular material checkboxes page', function() {
+  it('should click on checkbox', function() {
+    browser.get('https://material.angular.io/components/checkbox/overview');
+
+    element(by.checkboxText('Check me!')).click();
+
+    expect($('.mat-checkbox-input').getAttribute('aria-checked')).toEqual('true');
+  });
+});
--- a/example/addLocatorExample/conf.js
+++ b/example/addLocatorExample/conf.js
@@ -1,0 +1,27 @@
+// An example configuration file for addLocation function usage with mocha framework.
+exports.config = {
+  directConnect: true,
+
+  // Capabilities to be passed to the webdriver instance.
+  capabilities: {
+    'browserName': 'chrome'
+  },
+
+  specs: ['addLocator_usage_example_spec.js'],
+
+  onPrepare: function() {
+    // Create custom locator to get checkbox component by text
+    By.addLocator('checkboxText', function(checkboxText, opt_parentElement) {
+      // The first argument is checkbox text and second one opt_parentElement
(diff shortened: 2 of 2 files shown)


IgorSasovets:
I signed it!

qiyigg:
Could you explain why the original example doesn't work for you? Probably the original doc also need to be updated if doesn't work.

example/addLocatorExample/addLocator_usage_example_spec.js
```diff
@@ -0,0 +1,9 @@
+describe('angular material checkboxes page', function() {
+  it('should click on checkbox', function() {
+    browser.get('https://material.angular.io/components/checkbox/overview');
+
+    element(by.checkboxText('Check me!')).click();
+
+    expect($('.mat-checkbox-input').getAttribute('aria-checked')).toEqual('true');
```
glepretre: We faced this question when testing checkboxes with material: is `aria-checked` the proper way to test a checkbox in e2e?

IgorSasovets: Hi, @glepretre ! Actually, you can also use isSelected() function. In this case you will have:
```
expect($('.mat-checkbox-input').isSelected()).toBe(true);
```
But, I think that it depends on component structure. Here we have material component and as for me both methods works good.

IgorSasovets:
@qiyigg, I investigated this issue as I promised. When I replace 
```
return Array.prototype.filter.call(checkboxes, function(checkbox) {
      return checkbox.textContent.indexOf(checkboxText) !== -1;
 });
```
by this
```
return Array.prototype.filter.call(checkboxes, function(checkbox) {
   return checkbox.textContent === checkboxText;
 });
```
I will get NoSuchElementError.
bug

IgorSasovets:
In my tests case example I fixed it and want to add this example to help other engineers that could have a deal with such problem.

qiyigg:
"checkbox.textContent.indexOf(checkboxText) !== -1" is definitely not equivalent to "return checkbox.textContent === checkboxText;"
Which one is more appropriate is determined by your user case.

Your test case doesn't work probably due to the button text is "Check me!", not Check me!. if you inspect the html, you will find the text content has the extra double quotes.

IgorSasovets:
@qiyigg , thanks! I missed it.

IgorSasovets:
Hi, @qiyigg ! I have one more question. If I got element using standard locators (by.buttonText, cssContainingText, $, .etc) it would have css class in elementArrayFinder_.locator_.value property. Would element have css selector in this property too if I got it using custom locator?

PR #4756, opened 2018-04-02
feat(driverProvider): Add useExistingWebDriver driver provider

andyjack:


andyjack:
Please check my CLA!

kevcenteno:
@qiyigg Any thoughts on this?

andyjack:
Hi, any chance of this PR being merged?

lib/driverProviders/useExistingWebDriver.ts
```diff
@@ -0,0 +1,57 @@
+/*
+ *  This is an implementation of the Use Existing WebDriver Driver Provider.
+ *  It is responsible for setting up the account object, tearing it down, and
+ *  setting up the driver correctly.
+ */
+import * as q from 'q';
```
CrispusDH: Sorry for a simple question, but why you can not use `async ... await` construction instead of `q` library?

andyjack: I mostly copied the code from another driver provider, but I could redo my PR with `async/await` if that helps it getting merged. I don't exactly know what the house rules are for protractor code but I didn't see anything prohibiting `q`, fwiw.

qiyigg: The driver providers were written before Promise was available in Node.js.
Probably it's fine to use async/await for new driver provider, but I am not sure whether it will just work since some other folks tried to refactor them using regular promise, but it turns out not easy. Since all the other driver providers are using q promise, I am also ok with q promise for this one.

qiyigg:
merged, thanks

cnishina:
## Discussion on driverProviderUseExisting

So I'm working on this selenium upgrade process to 4. I understand the idea of this PR; however, I don't think this actually accomplishes the desired task. The selenium session should be passed to the constructor in 4.

```
class WebDriver {
  /**
   * @param {!(./session.Session|IThenable<!./session.Session>)} session Either
   *     a known session or a promise that will be resolved to a session.
   * @param {!command.Executor} executor The executor to use when sending
   *     commands to the browser.
   * @param {(function(this: void): ?)=} onQuit A function to call, if any,
   *     when the session is terminated.
   */
  constructor(session, executor, onQuit = undefined) {
```

Shown here:
https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/webdriver.js#L636

## We might already do this

The code provided looks and feels like if I was using an existing selenium server which is already defined here: https://github.com/angular/protractor/blob/master/lib/driverProviders/driverProvider.ts#L46 The only difference here that I can tell is that you are making the person writing the test also create the webdriver instance. Moving forward, attachToSession is being changed and we need to work on fixing the existing one.

andyjack:
Hi @cnishina - you are correct, my intent with `useExistingWebDriver` was to make it for having regular ` `selenium-webdriverjs` and protractor use the same instance of a `WebDriver`. Some more info [here](https://github.com/angular/protractor/blob/master/docs/server-setup.md#re-using-an-existing-webdriver). I couldn't see any way in the protractor code of instantiating protractor where it wouldn't also create a new `WebDriver`. The decisioning for what provider is used based on protractor config is [somewhat involved](https://github.com/angular/protractor/blob/d4fe1ca79a2101a63e239f8cba30cdd1c40401e6/lib/driverProviders/index.ts#L31) so adding a new case felt reasonable, rather than contorting the config to match the `attachToSession` provider (and I knew that `attachToSession` was already gone in selenium v4).

If there's a replacement for what the current `attachToSession` driver provider does in the v4 upgrade to protractor, that also allows one to instantiate protractor with an already-created driver, then removing `useExisting` sounds like a good idea to me.

Thank you btw for working on the upgrade :tada:

Issue #4708, opened 2018-02-17
Protractor unit test not actually testing anything, then breaks in Jasmine 2.9+


slackersoft:
The "local connect with the selenium standalone jar" spec isn't actually testing anything because it is calling Jasmine's `beforeEach`, `afterEach`, and `it` inside of another `it`. This probably caused the build breakages when Jasmine 2.9 was released, because that release made this throw an error and fail the test suite. The test is [here](https://github.com/angular/protractor/blob/a5dbe25e3dfd412a8b85eb1ee29990d25eba9a27/spec/unit/driverProviders/local_test.js#L41), but when I remove the outer `it` so that the inner one actually execute, the spec fails because the `Error` that is thrown doesn't have any `code` at all on it, instead it is:

```
Error: Do not know how to build driver: foobar explorer; did you forget to call usingServer(url)?
        at Builder.build (/Users/pivotal/workspace/protractor/node_modules/selenium-webdriver/index.js:663:15)
        at Local.getNewDriver (/Users/pivotal/workspace/protractor/built/driverProviders/driverProvider.js:53:33)
        at /Users/pivotal/workspace/protractor/spec/unit/driverProviders/local_test.js:64:19
```

And I'm not sure the right Protractor way to solve this. I was looking to send in a PR to update for Jasmine 3.0, which was released recently, but, I am unable to get a green test suite locally due to this.

PR #4789, opened 2018-04-27
Upgrade to jasmine 3.1.0

### LINKED ISSUE #4788: Upgrade to jasmine 3.1.0
Feature Request
---

Upgrade to jasmine 3.1.0

Jasmine 3 has been released for a while. A particular reason I want protractor to follow up is   jasmine/jasmine/#842 finally allow specDone support async.

awarecan:
resolve #4788 

Jamine 3 change the random default to true, however I want to keep protractor's default to false to avoid breaking changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "blocking-proxy": "^1.0.0",
     "chalk": "^1.1.3",
     "glob": "^7.0.3",
-    "jasmine": "2.8.0",
+    "jasmine": "3.1.0",
     "jasminewd2": "^2.1.0",
     "optimist": "~0.6.0",
     "q": "1.4.1",
@@ -31,7 +31,7 @@
   "devDependencies": {
     "@types/chalk": "^0.4.28",
     "@types/glob": "^5.0.29",
-    "@types/jasmine": "^2.5.47",
+    "@types/jasmine": "^2.8.6",
     "@types/jasminewd2": "^2.0.0",
     "@types/minimatch": "^2.0.28",
     "@types/minimist": "^1.1.28",
--- a/lib/frameworks/jasmine.js
+++ b/lib/frameworks/jasmine.js
@@ -91,13 +91,11 @@ exports.run = function(runner, specs) {
   };
 
   // Run specs in semi-random order
-  if (jasmineNodeOpts.random) {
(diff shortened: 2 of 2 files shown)


awarecan:
I signed it!

awarecan:
resolved #4708 as well

package.json
```diff
@@ -18,7 +18,7 @@
     "blocking-proxy": "^1.0.0",
     "chalk": "^1.1.3",
     "glob": "^7.0.3",
-    "jasmine": "2.8.0",
+    "jasmine": "3.1.0",
```
qiyigg: Please also update package-lock.json

spec/unit/driverProviders/local_test.js
```diff
@@ -65,7 +65,6 @@ describe('local connect', function() {
           webdriver.getNewDriver();
         } catch(e) {
           errorFound = true;
-          expect(e.code).toBe(BrowserError.CODE);
```
qiyigg: why remove this?

spec/unit/runner_test.js
```diff
@@ -65,6 +65,6 @@ describe('the Protractor runner', function() {
     var runner = new Runner(config);
     runner.run().then(function() {
       done.fail('expected error when no custom framework is defined');
-    }, done);
+    }, done());
```
qiyigg: The "then" function receives a Function as parameter, passing done() to it will call "done" first and make the test always pass, won't it?

spec/unit/driverProviders/local_test.js
awarecan: The purpose of this test is validate if error be thrown, not to verify the error code. I am not sure what is the root cause, but the exception don't have code property on my environment, the test will fail.

spec/unit/runner_test.js
awarecan: This test is expecting exception thrown, so if error occurred, done(), test passed.
One big change in Jasmine 3 is done() accept one parameter now, if we pass done function here, done(e) will be called, Jasmine will fail the test. Alternative change here is

```
    runner.run().then(function() {
      done.fail('expected error when no custom framework is defined');
    }, () => {return done();} );
```

PR #4809, opened 2018-05-10
Upgrade to jasmine 3.1.0

### LINKED ISSUE #4788: Upgrade to jasmine 3.1.0

awarecan:
resolve #4788, #4708 

Jamine 3 change the random default to true, however I want to keep protractor's default to false to avoid breaking changes.
done() accept error as parameter in Jasmine 3
beforeEach() no longer allow in it() in Jasmine 3

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,40 +1,72 @@
-machine:
-  node:
-    version: 6.9.1
-  environment:
-    # Fix issue with selenium-server in containers.
-    # See http://github.com/SeleniumHQ/docker-selenium/issues/87
-    DBUS_SESSION_BUS_ADDRESS: /dev/null
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/node:6.14-browsers
+        environment:
+          # Fix issue with selenium-server in containers.
+          # See http://github.com/SeleniumHQ/docker-selenium/issues/87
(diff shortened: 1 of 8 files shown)


awarecan:
@qiyigg Could you help to review it again? I messed up previous PR.

To answer two of your code review concern in previous PR:

1. why remove this?
```expect(e.code).toBe(BrowserError.CODE);```

The purpose of this test is validate if error be thrown, not to verify the error code. I am not sure what is the root cause, but the exception don't have code property on my environment, the test will fail.

2. ```}, done());```
The "then" function receives a Function as parameter, passing done() to it will call "done" first and make the test always pass, won't it?

This test is expecting exception thrown, so if error occurred, done(), test passed.
One big change in Jasmine 3 is done() accept one parameter now, if we pass done function here, done(e) will be called, Jasmine will fail the test. 
I sightly changed the code in this PR

awarecan:
Did I just ruined my PR again?

qiyigg:
@awarecan I guess so... what you need to do is:
fetch the latest master branch
*rebase* your pr on that branch

awarecan:
Recreate a new PR

PR #4820, opened 2018-05-21
Add click and scrollIntoView abstractions

simonua:
It would be nice to abstract the `click` and `scrollIntoView` calls a little, so that code such as 

```
(() => browser.executeScript('arguments[0].scrollIntoView();', this.locatorButton.webMail))
(() => browser.executeScript('arguments[0].click();', this.locatorButton.webMail))
```

could be rewritten as 

```
(() => browser.scrollIntoView(this.locatorButton.webMail))
(() => browser.click(this.locatorButton.webMail))
```

If there's interest in and support for this, I'd be happy to contribute more (unit tests, etc.), if needed, to bring this to completion.

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -1224,7 +1224,7 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
    * @returns {wdpromise.Promise<T>} 
    * @memberof ProtractorBrowser
    */
-  scroll<T>(element: ElementFinder): wdpromise.Promise<T> {
+  scrollIntoView<T>(element: ElementFinder): wdpromise.Promise<T> {
     return this.driver.executeScript('arguments[0].scrollIntoView();', element);
   }
 
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -1203,4 +1203,29 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
       return !!wdpromise.ControlFlow;
     }
   }
+
+  /**
+   * Clicks on the specified element, provided it's visible.
+   * 
+   * @template T 
+   * @param {ElementFinder} element 
+   * @returns {wdpromise.Promise<T>} 
+   * @memberof ProtractorBrowser
+   */
+  click<T>(element: ElementFinder): wdpromise.Promise<T> {
+    return this.driver.executeScript('argments[0].click();', element);
+  }
+
+  /**
+   * Scrolls to the specified element.
+   * 
+   * @template T 
+   * @param {ElementFinder} element 
+   * @returns {wdpromise.Promise<T>} 
+   * @memberof ProtractorBrowser
+   */
+  scroll<T>(element: ElementFinder): wdpromise.Promise<T> {
+    return this.driver.executeScript('arguments[0].scrollIntoView();', element);
+  }
+
 }


simonua:
I signed it!

Pablodotnet:
@simonua Remember scrollIntoView can have an optional parameter to align it to top or not, it should be on your abstraction too as it is very useful to use that parameter 👍

simonua:
Good point, @Pablodotnet. I pushed an update. I'm thinking this may cover:

image

image

image

Pablodotnet:
That's perfect! Hope this pull request get merge soon 💯

simonua:
@qiyigg, I merged upstream/master into my master and the CLA dropped out. Could you please advise what should be done? Also, is there merit to this PR, please?

awarecan:
To be honest, I didn't get it. Is an enhancement of `ElementFinder.scrollIntoView()` better?

Moreover, I don't think `browser.click()` is better than `ElementFinder.click()`, it looks like an anti-protractor design for me.

qiyigg:
@simonua in general, you shouldn't use any "merge" command. You should use "rebase" instead.
I guess you merged other's comment into your comment and now it is contributed by two people and CLA doesn't like it.

For this PR itself. I don't fully understand the purpose of doing so.
what's the purpose of using this:
(() => browser.executeScript('arguments[0].click();', this.locatorButton.webMail))

rather than:
this.locatorButton.webMail.click()

Usually we use browser.executeScript when we want to get the "return value"

awarecan:
1. proposed `browser.click()`
As my understanding, your intention is to click some DOM element which is not clickable by define of [WebDriver protocol](https://www.w3.org/TR/webdriver/#element-click). I remembered there was some discussion when Selenium decided move to WebDriver around the same issue. Your proposal is the way Selenium 1.0 (Selenium RC) handle clicking, and the major argument at that time was WebDriver no longer allow click invisible element. From e2e testing perspective, I strongly support WebDriver way, if user cannot see the element in browser, there is no reason your test script should click it.

2. proposed `browser.scrollIntoView()`
In [WebDriver protocol](https://www.w3.org/TR/webdriver/#element-click), the interactive with element such as click and sendKeys will `scroll into view` first. That eliminates the majority usage of your `browser.scrollIntoView`. Actually, in Selenium, there are already tons of code to make sure the WebDriver protocol got implemented right, for example
https://github.com/SeleniumHQ/selenium/blob/74e584d12ba31bb2e1b61b17a4751402c70c6da6/javascript/atoms/action.js#L601-L619
In my newer project, I use scroll fewer and fewer actually. 
Even you still want to introduce that function, I will suggest that put it into [ElementFinder](https://github.com/angular/protractor/blob/master/lib/element.ts). I feel that
```javascipt
this.locatorButton.webMail.scrollIntoView()
```
is better than
```javascipt
browser.scrollIntoView(this.locatorButton.webMail)
```
How do you think?

simonua:
@qiyigg and @awarecan, thank you very much for your feedback. I'll dig into your suggestions a bit into my project to see if I still see as much validity for this PR. I appreciate you!

Pablodotnet:
@awarecan 

> the interactive with element such as click and sendKeys will scroll into view first. That eliminates the majority usage of your browser.scrollIntoView.

I can't agree with this, personally I struggle with a lot of situations where the test failed because it says "element not visible" or "other element will receive click" and I need to scroll to them using scrollIntoView, that automatic scroll you say fails a lot of times.

simonua:
@Pablodotnet, I've had similar experiences, which motivated me to create the abstraction.

simonua:
I don't think I am opposed to putting it with `ElementFinder`. That would still appear to be a semantic way, just a different angle. Would that change gain support?

PR #4823, opened 2018-05-23
feat(plugin): pre-webdriver plugin hook

krotscheck:
This patch adds an 'initialize' plugin method, which is
invoked before the webdriver session has been initialized. It's aim is
to provide hooks by which a plugin author can modify the protractor configuration
before any serious logic has been executed. For instance, to configure the
multiCapabilities block based on the local environment, or querying a SaaS
selenium API (BrowserStack, SauceLabs, raw Selenium) for available browsers.

Any plugin that hooks into this is presumed to be 'required' - so a rejected
promise (perhaps caused by a missing API key) will terminate the protractor
execution.

### DIFF
--- a/lib/plugins.ts
+++ b/lib/plugins.ts
@@ -24,6 +24,24 @@ export interface PluginConfig {
 
 export interface ProtractorPlugin {
   /**
+   * Set up the protractor runtime before the webdriver session has been started.
+   * Use this method if you'd like to dynamically change the browser capabilities,
+   * multiCapabilites, plugin list, or other parameters.
+   *
+   * Note that if you use this method to inject other plugins, you will have to
+   * carefully manage their own initialize() invocations.
+   *
+   * @this {Object} bound to module.exports.
+   *
+   * @throws {*} If this function throws an error, the process will exit.
+   *
+   * @return {Promise=} Can return a promise, in which case protractor will wait
+   *     for the promise to resolve before continuing.  If the promise is
+   *     rejected, the process will exit.
+   */
+  initialize?(config: Config): void|Promise<void>;
+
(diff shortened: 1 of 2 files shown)


awarecan:
+1. I just think about add similar thing to allow each webdriver session dealy few seconds to release the I/O pressure during my tests initialing.

krotscheck:
@qiyigg Heya, do you know what the process is for landing patches and/or getting discussion going on things like this? I recall from ngConf that y'all are super low on project owners.

qiyigg:
Hi, sorry for the delay. 
In general, I don't know whether it is a good idea to modify the config dynamically. Since the inconsistency of config and runtime can cause confusing and hard to debug. Usually I would suggest having multiple configs if you do need different configurations for different tests.

krotscheck:
You're right, it may not be a good idea - though I'd argue it's not our responsibility to protect users from shooting themselves in the foot (plugin-user-beware). If we draw parallels to other test runners, Karma permits the same via its frameworks, and projects like angular/cli have used that to good effect to reduce configuration overhead for downstream users, while maintaining control of their own internal logic.

Having said that, I'm open to proposing a different patch that meets our need, all I need is a pointer on where to hook into it. Here's the gist: We'd like to make an asynchronous API call to load browserstack's current-list-of-browsers, paring it down to our support matrix, and setting `multiCapabilities` based on that.

krotscheck:
Hey, any response? I'd like to move forward with this.

qiyigg:
Could you add a test for it?
Then we'd see whether we need to have a way to remind user the protractor is controlled by a custom plugin if it's not obvious in the test.

PR #4834, opened 2018-05-31
feat(taskscheduler): allow to use custom scheduler for shards

ReactiveThings:
When shardTestFiles is true, specs will be sharded by file. (i.e. every test file is executed in separated browser instance).
This behavior can now be customized by providing custom shard scheduler function which allows to run multiple specs in one browser instance. For example we can run same amount of specs in each browser instance.


ReactiveThings:
I signed it!

PR #4961, opened 2018-09-13
set correct port for sauceSeleniumAddress for HTTP protocol

sauceaaron:
when sauceSeleniumUseHttp is true, it should set the port to 80 instead of 443


sauceaaron:
signed CLA

PR #4994, opened 2018-11-06
chore(test): move test suite off of the control flow

cnishina:
- set SELENIUM_PROMISE_MANAGER to false
- refactor to use async / await
- prefer using const over var

### DIFF
--- a/spec/basic/elements_spec.js
+++ b/spec/basic/elements_spec.js
@@ -1,159 +1,158 @@
-describe('ElementFinder', function() {
-  beforeEach(function() {
+describe('ElementFinder', () => {
+  beforeEach(async() => {
     // Clear everything between each test.
-    browser.driver.get('about:blank');
+    await browser.driver.get('about:blank');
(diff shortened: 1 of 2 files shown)


spec/basic/elements_spec.js
```diff
@@ -1,595 +1,594 @@
...
+    expect(await element(by.binding('greet')).isPresent()).toBe(true);
+    expect(await element(by.binding('nopenopenope')).isPresent()).toBe(false);
   });
 
-  it('should allow handling errors', function() {
-    browser.get('index.html#/form');
-    $('.nopenopenope').getText().then(function(/* string */) {
+  it('should allow handling errors', async() => {
+    await browser.get('index.html#/form');
+    await $('.nopenopenope').getText().then(() => {
```
CrispusDH: This might be more readable if used `try .. catch` construction.

cnishina:
Yikes this is against the wrong branch.

Issue #4909, opened 2018-08-07
The 'How to connect to Browser Drivers' section at the top in 'config.ts' file should list all 7 options


dineshb-qa:
*Steps to Reproduce* : 
1. Check the [config.ts](https://github.com/angular/protractor/blob/master/lib/config.ts) file
2. Observed that, at the top 5 different options are mentioned for 'How to connect to Browser Drivers'. If you scroll down then there are actually 7 available options to do this
3. The 'To use remote browsers via TestObject' and 'To use remote browsers via Kobiton' options are not mentioned at the top comments sections

PR #4996, opened 2018-11-06
docs(config.ts): Fix list with options to 'How to connect to Browser Drivers'

PauloGoncalvesBH:
Correction of the issue #4909


PauloGoncalvesBH:
I signet it!

Issue #4995, opened 2018-11-06
selenium 4 upgrade process


cnishina:
The `selenium-webdriver@4.0.0-alpha.1` upgrade process will be against the selenium4 branch. These will be code reviewed; however the tests might not pass.

change as much as possible without upgrading selenium-webdriver:

- [x] move test suite off of the control flow (test suite should pass here) - 11/09 #5017
- [x] remove the control flow - 11/23
- [x] convert webdriver.promise.Promise to native Promise - 11/23
- [x] convert q promises to native promises - 11/23
- [x] clean up ignoreSynchronization as deprecated - 11/23
- [x] remove lib/debugger
- [x] remove jasminewd2 - 11/23
  - [x] fix locators_spec.js. Has a commented out test in reference to "custom expectations". See #5001

unforeseen steps (pushing out timeline by a week):

- [x] update internal runners
- [x] update internal test targets

also unforseen internal fixit week (pushing out timeline by another week):

copy over new selenium-webdriver typings to fix api calls:

- [x] fix protractor with new typing changes - ~~11/30 12/07~~ 12/14
- [x] review changelog items from selenium4 - ~~11/30 12/07~~ 12/14

upgrade selenium 4:

- [x] upgrade deps to selenium 4 (tests will probably fail) - ~~11/30 12/07~~ 12/14
- [x] spend a magical amount of time ~~hitting head against wall~~ fixing things (if all goes well, maybe a week doing this) - ~~11/30 12/07~~ 12/14

clean up:

- [ ] move test suite to typescript - ~~12/07 12/14~~ 12/21
- [ ] light refactor with circle.yml file to .circleci folder - ~~12/07 12/14~~ 12/21
- [ ] look into bstack tests on travis failing (currently commented out)

wishlist:

- [ ] ~~remove extended webdriver js - 11/23~~ <-- this might not happen
- [x] maybe migrate over to webdriver-manager-replacement? or remove it from package.json - 12/21

cnishina:
Today I'll work on moving the basic/*_spec.js tests:

- locators_spec.js
- lib_spec.js
- mockmodule_spec.js

and then I'll jump around to other tests in the suite.

@CrispusDH, this means in the basic, I will not touch these files if you want to take a shot at them. Please work on one file at a time as a pull request. 

- expected_conditions_spec.js
- handling_spec.js
- navigation_spec.js
- polling_spec.js
- restart_spec.js
- synchronize_spec.js

[...]

PR #5001, opened 2018-11-07
chore(test): move locators_spec.js off of the control flow

cnishina:


cnishina:
Reference #4995

spec/basic/locators_spec.js
```diff
@@ -1,427 +1,398 @@
...
+  describe('by binding', () => {
+    it('should find an element by binding', async() => {
+      const greeting = element(by.binding('greeting'));
 
-      expect(greeting.getText()).toEqual('Hiya');
+      expect(await greeting.getText()).toEqual('Hiya');
     });
 
-    it('should allow custom expectations to expect an element', function() {
+    it('should allow custom expectations to expect an element', async() => {
```
cnishina: Not sure what I'm doing... :man_shrugging:

cnishina: This is passing because this is now an xit. This should be turned back on when removing the jasmine-wd node module. I will add this into the main issue to track this.

PR #5004, opened 2018-11-07
chore(test): use native promises instead of q onCleanUp

cnishina:


spec/onCleanUpAsyncReturnValueConf.js
```diff
@@ -15,11 +15,9 @@ exports.config = {
...
   baseUrl: env.baseUrl + '/ng1/',
 
-  onCleanUp: function(exitCode) {
-    var deferred = q.defer();
-    setTimeout(function() {
-      deferred.resolve(exitCode);
-    }, 500);
-    return deferred.promise;
+  onCleanUp: async(exitCode) => {
+    return await new Promise((resolve, _) => {
```
CrispusDH: it could be like:
```
return await new Promise(resolve => {
  setTimeout(resolve(exitCode), 500);
});
```

cnishina: Done.

cnishina:
Reference #4995.

PR #5005, opened 2018-11-07
chore(test): use native promises instead of q on onPrepare

cnishina:


spec/onPreparePromiseConf.js
```diff
@@ -17,9 +18,10 @@ exports.config = {
 
   baseUrl: env.baseUrl + '/ng1/',
 
-  onPrepare: function() {
-    return q.fcall(function() {
-      browser.params.password = '12345';
-    }).delay(1000);
+  onPrepare: async() => {
+    browser.params.password = '12345';
+    return await new Promise((resolve, _) => {
```
CrispusDH: it could be like:
```
return await new Promise(resolve => {
      setTimeout(resolve, 1000);
    });
```

cnishina: Done.

cnishina:
Reference #4995.

spec/onPrepare/asyncstartup.js
```diff
@@ -1,5 +1,6 @@
-var q = require('q');
```
heathkit: Yay!

cnishina: Woot!

PR #5007, opened 2018-11-08
chore(test): clean up mocha tests

cnishina:
- this.slow does not appear to be part of the mocha framework or we are
not correctly exposing it. Removed since it was causing errors.
- moved tests to be async / await where appropriate.


cnishina:
Reference #4995.

PR #5011, opened 2018-11-08
chore(test): move navigation_spec off of the control flow

CrispusDH:

I have checked that `TravisCI` run configs that should be disabled now. I have added a fix to `.travis.yml`. I think it is the simplest way to skip long tests that are flaky on Travis, but maybe you would prefer to do it in a different way.


CrispusDH:
Reference #4995

cnishina:
Thank you!

PR #5018, opened 2018-11-08
chore(test): move plugins tests off of the control flow and remove q

cnishina:


spec/plugins/browserGetSyncedConf.js
```diff
@@ -19,20 +19,26 @@ exports.config = {
...
+      onPageLoad: async function() {
+        return await new Promise(resolve => {
+          setTimeout(() => {
+            protractor.ON_PAGE_LOAD = true;
+            resolve();
+          }, 5000);
         });
       },
-      onPageStable: function() {
+      onPageStable: async function() {
```
CrispusDH: here `async` is redundant because function doesn't resolve promises inside

spec/plugins/waitForAngularConf.js
```diff
@@ -1,9 +1,9 @@
-var env = require('../environment.js'),
-    q = require('q');
+var env = require('../environment.js');
```
CrispusDH: could be `const`

spec/plugins/browserGetSyncedConf.js
cnishina: It resolves the await on the return.

spec/plugins/waitForAngularConf.js
cnishina: Done.

spec/plugins/browserGetSyncedConf.js
CrispusDH: yep :)

PR #5019, opened 2018-11-08
chore(test): move interaction test off of the control flow

cnishina:


spec/interaction/interaction_spec.js
```diff
@@ -1,134 +1,137 @@
...
+    await browser.get('index.html');
+    newBrowser = await browser.forkNewDriverInstance(true).ready;
+    expect(await newBrowser.driver.getCurrentUrl()).toMatch('index.html#/form');
   });
 
-  it('should be able to copy mock modules on fork', function() {
-    var mockModule = function() {
+  it('should be able to copy mock modules on fork', async() => {
+    const mockModule = () => {
       var newModule = angular.module('mockModule', []);
```
CrispusDH: it could be `let` or `const`

cnishina: done.

spec/interaction/interaction_spec.js
```diff
@@ -1,134 +1,137 @@
...
   });
 
-  it('should be able to navigate to same url on fork', function() {
-    browser.get('index.html');
-    newBrowser = browser.forkNewDriverInstance(true);
-    expect(newBrowser.driver.getCurrentUrl()).
-        toMatch('index.html#/form');
+  it('should be able to navigate to same url on fork', async() => {
+    await browser.get('index.html');
+    newBrowser = await browser.forkNewDriverInstance(true).ready;
```
heathkit: Apparently, this comes from: https://github.com/angular/protractor/blob/ed955e56a839d7f69da43acb6755763220d3681d/lib/runner.ts#L275

I think this is so convoluted because it needed to support the control flow case as well. This should be cleaned up - forkNewDriverInstance() should just return a promise that resolves to the new browser instance, after its ready.

PR #5024, opened 2018-11-09
chore(test): move hybrid/async_spec off of the control flow

CrispusDH:
- move hybrid/async_spec off of the control flow
- increase waiting time from 4s to 7s due to slow connection during SauceLabs tests in the ng2/async_spec (I have got error in the TravisCI after the first commit)


CrispusDH:
Reference #4995

PR #5027, opened 2018-11-09
chore(test): move ng2/timeout_spec off of the control flow

CrispusDH:
I'm not sure should I or not change expected error message (test is failed with the same result on CircleCI and locally)?
Expected error message: `Timed out waiting for asynchronous Angular tasks to finish`
Current error message: `Failed: script timeout: result was not received in 11 seconds` 


CrispusDH:
Reference #4995

cnishina:
Okay I looked into this a bit more, the failure is there probably because of this controlFlow call.

https://github.com/angular/protractor/blob/master/lib/browser.ts#L660 Probably the error is not properly thrown because we are do not execute the controlFlow(). We will come back to this when we untangle Protractor and the control flow.

cnishina:
@CrispusDH Could you rebase this? I fixed the control flow issue in the browser. Thank you!

CrispusDH:
@cnishina , unfortunately the error still the same as was

cnishina:
Closing this one. It looks like we got it done.

PR #5029, opened 2018-11-09
chore(test): re-work interactive_tests to native Promise

CrispusDH:
- re-work interactive_tests to native Promise. Locally tests were passed. But I'm not sure how to run tests in a proper way because `test.run()` returns Promise that I does not resolve.
- fix `ng2_async_spec`: change waiting value from `1` to `2`

@cnishina , @heathkit  could you take a look at this PR, please


CrispusDH:
Reference #4995

CrispusDH:
I don't know why but I can not get an error. The tests are passed locally. I mean `interactive_tests`. Probably it's broken somewhere inside.

cnishina:
Is it because the thrown error is caught in the .catch? Maybe something like this?

`test.run().then(() => {}).catch(err => {console.log(err)}`

CrispusDH:
I will try tomorrow, thanks.

CrispusDH:
Move PR to https://github.com/angular/protractor/pull/5037

PR #5032, opened 2018-11-12
feat(expectedConditions.ts): add method to check if an element's value changes

Joelasaur:
We needed a way to wait for an element's value to change. The current solution, textToBePresentInElement(), only works if you know what the text is going to be. This code is currently working and deployed in our development cluster. I figured the utility would be nice to share with everyone.

### DIFF
--- a/lib/expectedConditions.ts
+++ b/lib/expectedConditions.ts
@@ -244,6 +244,31 @@ export class ProtractorExpectedConditions {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


Joelasaur:
I signed it!

odinho:
@Joelasaur Thankfully your PR existed, as I thought my PR #5035 actually had that same test failing for some mysterious reason.

PR #5034, opened 2018-11-12
chore(driverProviders): clean up driver provider q usage

cnishina:
- Remove driverProviderUseExistingWebDriver since the generation of the selenium server is already accomplished when providing a selenium address in driverProvider.ts. Also clean up docs and tests.
- Use native promises over q promises.


lib/driverProviders/driverProvider.ts
```diff
@@ -68,22 +67,22 @@ export abstract class DriverProvider {
...
     let driverIndex = this.drivers_.indexOf(driver);
     if (driverIndex >= 0) {
       this.drivers_.splice(driverIndex, 1);
     }
 
     if (driver.getSession() === undefined) {
-      return wdpromise.when(undefined);
+      return Promise.resolve(undefined);
     } else {
       return driver.getSession()
```
heathkit: This can now be `const session = await driver.getSession()`

lib/driverProviders/driverProvider.ts
```diff
@@ -93,56 +92,48 @@ export abstract class DriverProvider {
...
         }))
-        .then(
-            () => {
-              deferred.resolve();
-            },
-            () => {
-              deferred.resolve();
-            });
-    return deferred.promise;
+        .then(() => {});
```
heathkit: Why do we need this?

lib/driverProviders/driverProvider.ts
```diff
@@ -93,56 +92,48 @@ export abstract class DriverProvider {
...
   }
 
   /**
    * Default update job method.
    * @return a promise
    */
-  updateJob(update: any): q.Promise<any> {
-    return q.fcall(function() {});
+  updateJob(update: any): Promise<any> {
+    return Promise.resolve(() => {});
```
heathkit: Why not `Promise.resolve({})`?

lib/driverProviders/driverProvider.ts
```diff
@@ -93,56 +92,48 @@ export abstract class DriverProvider {
...
   /**
    * Default setup environment method, common to all driver providers.
    */
-  setupEnv(): q.Promise<any> {
+  async setupEnv(): Promise<any> {
     let driverPromise = this.setupDriverEnv();
     if (this.config_.useBlockingProxy && !this.config_.blockingProxyUrl) {
       // TODO(heathkit): If set, pass the webDriverProxy to BP.
-      return driverPromise.then(() => this.bpRunner.start());
+      await driverPromise.then(() => this.bpRunner.start());
```
heathkit: No need to chain here. Actually, this whole thing could be a lot simpler

```
await this.setupDriverEnv();
if (this.config_.useBlockingProxy && !this.config_.blockingProxyUrl) {
   await this.bpRunner.start()
}
```

lib/driverProviders/hosted.ts
```diff
@@ -19,11 +17,11 @@ export class Hosted extends DriverProvider {
...
    * @public
-   * @return {q.promise} A promise which will resolve when the environment is
+   * @return {Promise} A promise which will resolve when the environment is
    *     ready to test.
    */
-  protected setupDriverEnv(): q.Promise<any> {
+  protected setupDriverEnv(): Promise<any> {
     logger.info('Using the selenium server at ' + this.config_.seleniumAddress);
-    return q.fcall(function() {});
+    return Promise.resolve(function() {});
```
heathkit: Can just be `Promise.resolve()`

lib/driverProviders/mock.ts
```diff
@@ -21,19 +20,17 @@ export class Mock extends DriverProvider {
...
 
   /**
    * Configure and launch (if applicable) the object's environment.
    * @public
    * @return {q.promise} A promise which will resolve immediately.
    */
-  protected setupDriverEnv(): q.Promise<any> {
-    return q.fcall(function() {});
+  protected setupDriverEnv(): Promise<any> {
+    return Promise.resolve(function() {});
```
heathkit: Just do `Promise.resolve()`

lib/driverProviders/driverProvider.ts
cnishina: Previously this satisfied typings where we would return a Promise<void>. Promise.all returns an Promise<void[]>. Instead since we have async / await, we could just await Promise.all to return. Without specifying a return type, this could be Promise<void>.

Fixing this.

lib/driverProviders/mock.ts
cnishina: Fixed.

lib/driverProviders/hosted.ts
cnishina: Fixed.

lib/driverProviders/driverProvider.ts
cnishina: Fixed.

lib/driverProviders/driverProvider.ts
cnishina: Fixed.

lib/driverProviders/driverProvider.ts
cnishina: Fixed.

cnishina:
Errrr....Fixed it!

first try

lib/browser.ts
```diff
@@ -193,24 +193,18 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
...
-          if (this.bpClient) {
-            const bpCommandPromise = this.bpClient.setWaitParams(value);
-            // Convert to webdriver promise as best as possible
-            return wdpromise.when(bpCommandPromise as any).then(() => this.internalRootEl);
-          }
-          return this.internalRootEl;
-        });
+  async angularAppRoot(valuePromise: string|wdpromise.Promise<string> = null): Promise<string> {
+    if (valuePromise != null) {
+      const value = await Promise.resolve(valuePromise);
```
heathkit: You don't need the Promise.resolve. You can just `await valuePromise`. If It's not a promise, await will just return the string.

lib/browser.ts
```diff
@@ -417,23 +411,17 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
...
-            const bpCommandPromise = this.bpClient.setWaitEnabled(enabled);
-            // Convert to webdriver promise as best as possible
-            return wdpromise.when(bpCommandPromise as any).then(() => enabled);
-          }
-        });
-      }, `Set proxy synchronization enabled to ${enabled}`);
+  async waitForAngularEnabled(enabledPromise: boolean|wdpromise.Promise<boolean> = null):
+      Promise<boolean> {
+    if (enabledPromise != null) {
+      const enabled = await Promise.resolve(enabledPromise);
```
heathkit: Same as above.

lib/browser.ts
```diff
@@ -624,26 +612,20 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
...
+      return Promise.resolve(true);
     }
 
-    let runWaitForAngularScript: () => wdpromise.Promise<any> = () => {
+    let runWaitForAngularScript = async(): Promise<any> => {
       if (this.plugins_.skipAngularStability() || this.bpClient) {
-        return this.driver.controlFlow().execute(() => {
-          return wdpromise.when(null);
-        }, 'bpClient or plugin stability override');
+        return Promise.resolve(null);
```
heathkit: This is an async function, so you can just return null. Anything you return automatically becomes a promise, anyway.

lib/browser.ts
```diff
@@ -978,16 +952,14 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
         .then(() => {
           // Reset bpClient sync
           if (this.bpClient) {
-            return this.driver.controlFlow().execute(() => {
-              return this.bpClient.setWaitEnabled(!this.internalIgnoreSynchronization);
-            });
+            return this.bpClient.setWaitEnabled(!this.internalIgnoreSynchronization);
           }
         })
         .then(() => {
```
heathkit: This promise chain should eventually become async/await, but let's save that for another Pr.

lib/driverProviders/attachSession.ts
```diff
@@ -25,10 +24,10 @@ export class AttachSession extends DriverProvider {
    * @return {q.promise} A promise which will resolve when the environment is
```
heathkit: You can remove all these type annotations from the JSDoc, since this is now in TypeScript.

lib/driverProviders/direct.ts
```diff
@@ -43,7 +42,7 @@ export class Direct extends DriverProvider {
             'browserName ' + this.config_.capabilities.browserName +
                 ' is not supported with directConnect.');
     }
-    return q.fcall(function() {});
+    return Promise.resolve(function() {});
```
heathkit: Since this function doesn't need await, let's not make it async. Just have it return `Promise.resolve()`. The empty function was needed when using `q.fcall()`, but it's not anymore.

lib/driverProviders/driverProvider.ts
```diff
@@ -68,22 +67,20 @@ export abstract class DriverProvider {
...
-          })
-          .catch<void>(function(err: Error) {});
+    try {
+      if (driver.getSession() !== undefined) {
+        const session = await driver.getSession();
+        if (session) {
+          return await driver.quit();
+        }
+      }
+    } catch (_) {
```
heathkit: Let's at least add a comment that we're intentionally ignoring any errors when quitting the browser at the end of the test. Also, is that a good idea?

lib/driverProviders/driverProvider.ts
```diff
@@ -93,56 +90,44 @@ export abstract class DriverProvider {
...
-        .then(
-            () => {
-              deferred.resolve();
-            },
-            () => {
-              deferred.resolve();
-            });
-    return deferred.promise;
+  static async quitDrivers(provider: DriverProvider, drivers: WebDriver[]): Promise<void> {
+    await Promise.all(drivers.map(async (driver: WebDriver) => {
```
heathkit: Awaiting in the map will serialize all the promises. It's not super important, but you could just do
```
await Promise.all(drivers.map((driver: WebDriver) => {
     return provider.quitDriver(driver);
}));
```
which would actually wait for all the promises to resolve in parallel.

lib/driverProviders/local.ts
```diff
@@ -182,10 +171,12 @@ export class Local extends DriverProvider {
...
-  teardownEnv(): q.Promise<any> {
-    return super.teardownEnv().then(() => {
+  async teardownEnv(): Promise<any> {
+    try {
+      await super.teardownEnv();
       logger.info('Shutting down selenium standalone server.');
-      return this.server_.stop();
-    });
+      await this.server_.stop();
+    } catch (_) {
```
heathkit: Why eat the exception here? We could just leave out the try {} catch {} and let the uncaught exception handler take care of it if there's nothing to do here. Or at least log the error.

lib/browser.ts
cnishina: Neat!

lib/browser.ts
cnishina: I agree, this promise chain might be better contained in a separate PR.

lib/driverProviders/attachSession.ts
cnishina: Updated.

lib/driverProviders/direct.ts
cnishina: I removed this completely.

lib/driverProviders/driverProvider.ts
cnishina: Oh yup, that's right. Fixing.

lib/driverProviders/driverProvider.ts
cnishina: Cleaned this up! 

Calling driver.getSession() on a previously quit driver throws errors. So that was removed. If someone called quit in the test (which they probably shouldn't...looking at you interactive tests), then Protractor should try and catch this.

lib/driverProviders/local.ts
cnishina: Neat, I deleted this method.

cnishina:
Fixes are in!

done

lib/browser.ts
```diff
@@ -624,116 +612,91 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
...
-                'Error while waiting for Protractor to ' +
-                'sync with the page: ' + JSON.stringify(browserErr));
+    let browserErr = await runWaitForAngularScript();
+    if (browserErr) {
+      throw new Error(
+          'Error while waiting for Protractor to ' +
+          'sync with the page: ' + JSON.stringify(browserErr));
+    }
+    await this.plugins_.waitForPromise(this);
+    try {
```
heathkit: The `try {` should go on line 630, before runWaitForAngularScript();

If the waitForAngular script returns anything, that's an error. But it can also throw an exception if, for example, we time out waiting for the script to finish. The catch block is looking for those timeout errors and trying to make them more useful.

lib/browser.ts
```diff
@@ -624,116 +612,91 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
...
+        for (key in pendingHttps) {
+          pendingTasks.push(' - $http: ' + pendingHttps[key].url);
+        }
+        if (pendingTasks.length) {
+          errMsg += '. \nThe following tasks were pending:\n';
+          errMsg += pendingTasks.join('\n');
+        }
+        err.message = errMsg;
+        throw err;
+      } else {
```
heathkit: We always `throw err;` whether or not `if (timeout)` is true, so just extract that from the if statement and drop the `else`

lib/browser.ts
```diff
@@ -624,116 +612,91 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
...
+        // Timeout on Firefox
+        timeout = /-?[\d\.]*ms/.exec(err.message);
+      } else if (/Timed out waiting for an asynchronous script/.test(err.message)) {
+        // Timeout on Safari
+        timeout = /-?[\d\.]*\ ms/.exec(err.message);
+      }
+      if (timeout) {
+        let errMsg = `Timed out waiting for asynchronous Angular tasks to finish after ` +
+            `${timeout}. This may be because the current page is not an Angular ` +
+            `application. Please see the FAQ for more details: ` +
```
heathkit: Let's do ourselves a favor and extract as much as possible from this if statement into a private function.

lib/browser.ts
```diff
@@ -624,116 +612,91 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
...
+    }
+    await this.plugins_.waitForPromise(this);
+    try {
+      await this.driver.wait(async () => {
+        let results = await this.plugins_.waitForCondition(this);
+        return results.reduce((x, y) => x && y, true);
+      }, this.allScriptsTimeout, 'Plugins.waitForCondition()');
+    } catch (err) {
+      let timeout: RegExpExecArray;
+      if (/asynchronous script timeout/.test(err.message)) {
```
heathkit: Might as well extract this into a private function to make it more readable.

lib/browser.ts
cnishina: This is a great idea but I think this PR is getting a little out of hand and not keeping on point to what the original description was. I'll write an issue on this for an improvement.

lib/browser.ts
cnishina: Same with this.

cnishina:
Reference #4995

PR #5035, opened 2018-11-13
fix(browser): waitForAngularEnabled reset to 'true' on new browser

odinho:
Breaking change: `browser.forkNewDriverInstance` argument
`copyConfigUpdates` did not actually copy the config, now it does.

If you actually wanted `forkNewDriverInstance` to *not* copy the
previous browsers config and did not set `copyConfigUpdates` explicitly
to `false`, you will have to update your code to do that.

`waitForAngularEnabled` was also always set to true even if
`parentBrowser` had it as false. This would manifest itself as
`waitForAngularEnabled` suddenly reverting to `true` on a new browser
even if you set it to `false`. A work around was to run it twice in two
different promises.


odinho:
I signed it.

odinho:
`CircleCI` is failing with the same error as #5032. So even if I would not be suprised about any failure (I tried running locally, but it also failed without my patch), that particular one seems like a flake.

lib/runner.ts
```diff
@@ -272,26 +272,28 @@ export class Runner extends EventEmitter {
...
-            })
-            .then(() => {
-              return browser_;
-            });
+    browser_.ready = browser_.ready
+                         .then(() => {
+                           return initProperties.waitForAngularEnabled;
+                         })
+                         .then((waitForAngularEnabled) => {
+                           return browser_.waitForAngularEnabled(waitForAngularEnabled);
```
cnishina: Yup, I agree. waitForAngularEnabled is a promise and the promise is not resolved in the waitForAngularEnabled set method. So this makes sense.

lib/runner.ts
```diff
@@ -272,26 +272,28 @@ export class Runner extends EventEmitter {
...
+                         });
 
     browser_.getProcessedConfig = () => {
       return wdpromise.when(config);
     };
 
     browser_.forkNewDriverInstance =
         (useSameUrl: boolean, copyMockModules: boolean, copyConfigUpdates = true) => {
-          let newBrowser = this.createBrowser(plugins);
+          let newBrowser = this.createBrowser(plugins, copyConfigUpdates ? browser_ : null);
```
cnishina: This makes sense to me.

cnishina:
Thanks for the PR. I'll have to look through the CI logs...master branch is having some issues.

PR #5037, opened 2018-11-13
move interactive_test_util off q, rework to ES6 syntax

CrispusDH:
- move interactive_test_util off q, rework to ES6 syntax
- `interactive_test.js` and `with_base_url.js` still does not work as expected. Maybe we should rework commands or it does not work correctly together with common `executor`.


CrispusDH:
@cnishina , @heathkit could you take a look at this PR, please

circle.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/node:8.11-browsers
+      - image: circleci/node:10.13.0-browsers
```
cnishina: I will change this back. We had a discussion about this and it should be 8.11 for now.

CrispusDH:
@cnishina could you change label to `yes` and merge :)

CrispusDH:
@cnishina should I make some additional changes?

CrispusDH:
@cnishina , I merge `selenium4` branch into this one. Now it is up to date.

cnishina:
Fixed the CLAs. This PR is great. I'm trying to remove q and this aligns well.

I also think I stabilized the Travis tests by updating the browser versions we use in Sauce Labs. If it doesn't pass, I'll restart it but I'm not concerned if it doesn't pass.

PR #5041, opened 2018-11-14
Update config.ts sauceSeleniumAddress comment to reflect 443 default

enriquegh:
Per the default behavior set in:
https://github.com/angular/protractor/blob/master/lib/driverProviders/sauce.ts#L68-L70

the Sauce OnDemand address will default to ondemand.saucelabs.com:443 unless changed.
The config.ts doc says the default is ondemand.saucelabs.com:80.

This PR simply changes the port number on the config.ts comment.


enriquegh:
I signed it!

PR #5044, opened 2018-11-15
chore(promises): move locator wdpromsie to native Promise

cnishina:


cnishina:
Hahaha. Yah.... so let's try that again.

PR #5047, opened 2018-11-16
chore(promises): move element wdpromise to native Promise

cnishina:


cnishina:
Work in progress.  I want to pass the full CI (not on my computer)

lib/element.ts
```diff
@@ -414,21 +404,20 @@ export class ElementArrayFinder extends WebdriverWebElement {
...
-        });
+  async count(): Promise<number> {
+    try {
+      const arr = await this.getWebElements();
+      return arr.length;
+    } catch (err) {
+      if (err instanceof wderror.NoSuchElementError) {
+        return 0;
+      } else {
+        throw err;
```
CrispusDH: why not `throw new Error(err)`? I read that this is better practice :)

cnishina: Without a try catch, this will automatically throw the `err`. Wrapping the `err` in a new `Error` could give us the wrong stack trace and error message. When returning a number of elements, we are just catching where there are no elements to return a count of 0.

I think it is okay to leave this as is.

lib/element.ts
```diff
@@ -481,7 +469,7 @@ export class ElementArrayFinder extends WebdriverWebElement {
       ElementArrayFinder {
     let callerError = new Error();
     let actionResults = this.getWebElements()
-                            .then((arr: any) => wdpromise.all(arr.map(actionFn)))
+                            .then((arr: any) => Promise.all(arr.map(actionFn)))
                             .then(
                                 (value: any) => {
                                   return {passed: true, value: value};
```
CrispusDH: it could be `return {passed: true, value};`

lib/element.ts
```diff
@@ -647,19 +631,18 @@ export class ElementArrayFinder extends WebdriverWebElement {
...
-  map<T>(mapFn: (elementFinder?: ElementFinder, index?: number) => T | any):
-      wdpromise.Promise<T[]> {
-    return this.asElementFinders_().then<T[]>((arr: ElementFinder[]) => {
-      let list = arr.map((elementFinder?: ElementFinder, index?: number) => {
-        let mapResult = mapFn(elementFinder, index);
-        // All nested arrays and objects will also be fully resolved.
-        return wdpromise.fullyResolved(mapResult) as wdpromise.Promise<T>;
-      });
-      return wdpromise.all(list);
+  async map<T>(mapFn: (elementFinder?: ElementFinder, index?: number) => T | any): Promise<T[]> {
```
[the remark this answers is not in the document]
cnishina: Could you write an issue about this? We will address this later. (Procrastination :stuck_out_tongue:)

PR #5052, opened 2018-11-22
chore(promises): remove q promises and webdriver promises

cnishina:
- changes starting with cleaning up q promises and webdriver promises from plugins

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -717,7 +717,8 @@ export interface Config {
   frameworkPath?: string;
 
   /**
-   * Deprecated: Element explorer is no longer supported.
+   * Deprecated: Element explorer depends on the WebDriver control flow, and
+   * thus is no longer supported.
    */
   elementExplorer?: any;
   debug?: boolean;
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -715,6 +715,10 @@ export interface Config {
   nodeDebug?: boolean;
   debuggerServerPort?: number;
   frameworkPath?: string;
+
+  /**
+   * Deprecated: Element explorer is no longer supported.
+   */
   elementExplorer?: any;
   debug?: boolean;
   unknownFlags_?: string[];
--- a/lib/launcher.ts
+++ b/lib/launcher.ts
@@ -6,12 +6,13 @@ import * as fs from 'fs';
 
 import {Config} from './config';
 import {ConfigParser} from './configParser';
-import {ConfigError, ErrorHandler, ProtractorError} from './exitCodes';
+import {ConfigError, ErrorHandler} from './exitCodes';
 import {Logger} from './logger';
 import {Runner} from './runner';
 import {TaskRunner} from './taskRunner';
 import {TaskScheduler} from './taskScheduler';
-import * as helper from './util';
+import {runFilenameOrFn_} from './util';
+
 
 let logger = new Logger('launcher');
 let RUNNERS_FAILED_EXIT_CODE = 100;
@@ -107,7 +108,7 @@ let initFn = async function(configFile: string, additionalConfig: Config) {
   logger.debug('Your base url for tests is ' + config.baseUrl);
 
   // Run beforeLaunch
-  await helper.runFilenameOrFn_(config.configDir, config.beforeLaunch);
+  await runFilenameOrFn_(config.configDir, config.beforeLaunch);
   // 1) If getMultiCapabilities is set, resolve that as
   // `multiCapabilities`.
   if (config.getMultiCapabilities && typeof config.getMultiCapabilities === 'function') {
@@ -139,22 +140,11 @@ let initFn = async function(configFile: string, additionalConfig: Config) {
     config.multiCapabilities = [{browserName: 'chrome'}];
   }
 
-  // 3) If we're in `elementExplorer` mode, run only that.
+  // 3) If we're in `elementExplorer` mode, throw an error and exit.
   if (config.elementExplorer || config.framework === 'explorer') {
-    if (config.multiCapabilities.length != 1) {
-      throw new Error('Must specify only 1 browser while using elementExplorer');
-    } else {
-      config.capabilities = config.multiCapabilities[0];
-    }
-    config.framework = 'explorer';
-    try {
-      let runner = new Runner(config);
-      let exitCode = await runner.run();
-      process.exit(exitCode);
-    } catch (err) {
-      logger.error(err);
-      process.exit(1);
-    }
+    const err = new Error('elementExplorer is no longer');
+    logger.error(err);
+    process.exit(1);
   }
 
   // 4) Run tests.
@@ -195,8 +185,7 @@ let initFn = async function(configFile: string, additionalConfig: Config) {
   // Run afterlaunch and exit
   const cleanUpAndExit = async (exitCode: number) => {
     try {
-      let returned =
-          await helper.runFilenameOrFn_(config.configDir, config.afterLaunch, [exitCode]);
+      const returned = await runFilenameOrFn_(config.configDir, config.afterLaunch, [exitCode]);
       if (typeof returned === 'number') {
         process.exit(returned);
       } else {
@@ -208,7 +197,7 @@ let initFn = async function(configFile: string, additionalConfig: Config) {
     }
   };
 
-  let totalTasks = scheduler.numTasksOutstanding();
+  const totalTasks = scheduler.numTasksOutstanding();
   let forkProcess = false;
   if (totalTasks > 1) {  // Start new processes only if there are >1 tasks.
     forkProcess = true;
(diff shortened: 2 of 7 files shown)


cnishina:
This does not include the browser changes and I have doubts that the test suite will pass.

cnishina:
Circle CI is passing...
i don't believe you

lib/launcher.ts
```diff
@@ -93,7 +92,7 @@ let taskResults_ = new TaskResults();
  * @param {string=} configFile
  * @param {Object=} additionalConfig
  */
-let initFn = function(configFile: string, additionalConfig: Config) {
+let initFn = async function(configFile: string, additionalConfig: Config) {
   let configParser = new ConfigParser();
   if (configFile) {
```
CrispusDH: do we need it still here with `TS`? Probably, not.

lib/launcher.ts
```diff
@@ -3,7 +3,6 @@
  * input configuration and launching test runners.
  */
 import * as fs from 'fs';
```
CrispusDH: `fs` will work without importing

lib/launcher.ts
```diff
@@ -3,7 +3,6 @@
...
-import * as q from 'q';
 
 import {Config} from './config';
 import {ConfigParser} from './configParser';
```
CrispusDH: on line 14: `import * as helper from './util';`
But we use here just `runFilenameOrFn_` function, so we can import just it `import { runFilenameOrFn_ } from './configParser';`

lib/launcher.ts
```diff
@@ -108,197 +107,170 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
-                process.exit(1);
-              });
-        }
-      })
-      .then(() => {
-        // 4) Run tests.
-        let scheduler = new TaskScheduler(config);
+  // 3) If we're in `elementExplorer` mode, run only that.
+  if (config.elementExplorer || config.framework === 'explorer') {
+    if (config.multiCapabilities.length != 1) {
```
CrispusDH: I think it might be better to use `!==` instead of `!=`

lib/launcher.ts
```diff
@@ -108,197 +107,170 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
-          forkProcess = true;
-          if (config.debug) {
-            throw new ConfigError(
-                logger, 'Cannot run in debug mode with multiCapabilities, count > 1, or sharding');
-          }
-        }
+  // Run afterlaunch and exit
+  const cleanUpAndExit = async (exitCode: number) => {
+    try {
+      let returned =
```
CrispusDH: it can be `const`

lib/launcher.ts
```diff
@@ -108,197 +107,170 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
-        };
-        // Start `scheduler.maxConcurrentTasks()` workers for handling tasks in
-        // the beginning. As a worker finishes a task, it will pick up the next
-        // task
-        // from the scheduler's queue until all tasks are gone.
-        for (let i = 0; i < scheduler.maxConcurrentTasks(); ++i) {
-          createNextTaskRunner();
-        }
-        logger.info('Running ' + scheduler.countActiveTasks() + ' instances of WebDriver');
+  let totalTasks = scheduler.numTasksOutstanding();
```
CrispusDH: it can be `const`

lib/launcher.ts
```diff
@@ -108,197 +107,170 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
+  let forkProcess = false;
```
CrispusDH: it can be `const`

lib/launcher.ts
```diff
@@ -108,197 +107,170 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
 
-        // By now all runners have completed.
-        deferred.promise
-            .then(function() {
-              // Save results if desired
-              if (config.resultJsonOutputFile) {
-                taskResults_.saveResults(config.resultJsonOutputFile);
+  const createNextTaskRunner = async () => {
+    return new Promise(resolve => {
+      let task = scheduler.nextTask();
```
CrispusDH: it can be `const`

lib/launcher.ts
```diff
@@ -108,197 +107,170 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
+      if (task) {
+        let taskRunner = new TaskRunner(configFile, additionalConfig, task, forkProcess);
```
CrispusDH: it can be `const`

lib/launcher.ts
```diff
@@ -108,197 +107,170 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
+        taskRunner.run()
+            .then(async (result) => {
```
CrispusDH: why you don't want to use `async ... await` here?

lib/launcher.ts
```diff
@@ -108,197 +107,170 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
+            .catch(async (err: Error) => {
+              const errorCode = ErrorHandler.parseError(err);
+              logger.error('Error:', (err as any).stack || err.message || err);
+              await cleanUpAndExit(errorCode ? errorCode : RUNNERS_FAILED_EXIT_CODE);
+            });
+      }
+    });
+  };
+
+  for (let i = 0; i < scheduler.maxConcurrentTasks(); ++i) {
```
CrispusDH: why not `for (let i = 1; i < scheduler.maxConcurrentTasks(); i++) {`

lib/launcher.ts
```diff
@@ -108,197 +107,170 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
-      .done();
+            .catch(async (err: Error) => {
+              const errorCode = ErrorHandler.parseError(err);
+              logger.error('Error:', (err as any).stack || err.message || err);
+              await cleanUpAndExit(errorCode ? errorCode : RUNNERS_FAILED_EXIT_CODE);
+            });
+      }
+    });
+  };
+
```
CrispusDH: it might be better to make some variable `const maxConcurrentTask = scheduler.maxConcurrentTasks()` and then use it in `for () {` instead of calculate it each iteration in a loop

lib/launcher.ts
```diff
@@ -108,197 +107,170 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
-                logger.error(err);
-                process.exit(1);
-              });
-        }
-      })
-      .then(() => {
-        // 4) Run tests.
-        let scheduler = new TaskScheduler(config);
+  // 3) If we're in `elementExplorer` mode, run only that.
+  if (config.elementExplorer || config.framework === 'explorer') {
```
heathkit: ElementExplorer won't work without the control flow. If it's enabled, we should just log an error and exit.

lib/launcher.ts
```diff
@@ -108,197 +107,170 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
-            let protractorError = e as ProtractorError;
-            ProtractorError.log(logger, errorCode, protractorError.message, protractorError.stack);
-            process.exit(errorCode);
-          } else {
-            logger.error(e.message);
-            logger.error(e.stack);
-            process.exit(ProtractorError.CODE);
-          }
-        });
+  // process.on('uncaughtException', (exc: (Error|string)) => {
```
heathkit: Why is this commented out?

lib/runner.ts
```diff
@@ -381,120 +356,106 @@ export class Runner extends EventEmitter {
...
+
+    // 1) Setup environment
+    // noinspection JSValidateTypes
+    await this.driverprovider_.setupEnv();
+
+    // 2) Create a browser and setup globals
+    browser_ = this.createBrowser(plugins);
+    this.setupGlobals_(browser_);
+    try {
+      let session = await browser_.ready.then(browser_.getSession);
```
heathkit: const

lib/launcher.ts
heathkit: Not this one - it gets set later.

lib/launcher.ts
cnishina: Done.

lib/launcher.ts
cnishina: Done.

lib/launcher.ts
cnishina: Done.

lib/launcher.ts
cnishina: Done.

lib/launcher.ts
cnishina: Done.

lib/launcher.ts
cnishina: Good idea. Changing this to async / await

lib/launcher.ts
cnishina: Done.

lib/runner.ts
```diff
@@ -144,16 +143,15 @@ export class Runner extends EventEmitter {
...
-        .then((returned): number | any => {
-          if (typeof returned === 'number') {
-            return returned;
-          } else {
-            return exitCode;
-          }
-        });
-  };
+  async exit_(exitCode: number): Promise<number> {
+    let returned =
```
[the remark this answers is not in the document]
cnishina: Done.

lib/launcher.ts
cnishina: I actually do not believe this is ever being called. The thrown call is unhandledPromiseRejectionWarning. This is not being bucketed into uncaughtException. I have 3 error tests that failed in the big PR that I am not sure how to resolve. Should I add a TODO?

lib/runner.ts
cnishina: Done.

lib/launcher.ts
cnishina: I am not sure. Could you clarify?

lib/launcher.ts
cnishina: Done. Also updated runner with the same fix.

lib/launcher.ts
cnishina: Done.

lib/config.ts
```diff
@@ -715,6 +715,10 @@ export interface Config {
   nodeDebug?: boolean;
   debuggerServerPort?: number;
   frameworkPath?: string;
+
+  /**
+   * Deprecated: Element explorer is no longer supported.
```
heathkit: Maybe expand this: 'Element explorer depends on the WebDriver control flow, and thus is no longer supported'

lib/launcher.ts
```diff
@@ -108,197 +108,157 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
-                process.exit(1);
-              });
-        }
-      })
-      .then(() => {
-        // 4) Run tests.
-        let scheduler = new TaskScheduler(config);
+  // 3) If we're in `elementExplorer` mode, throw an error and exit.
+  if (config.elementExplorer || config.framework === 'explorer') {
+    const err = new Error('elementExplorer is no longer');
```
heathkit: Needs more text in the error

lib/launcher.ts
heathkit: 'unhandledRejection' and 'uncaughtException' are separate events, see: https://nodejs.org/api/process.html#process_event_unhandledrejection

We should have handlers for both. uncaughtException should never happen, unless there's a Protractor bug or something wrong with the test framework integration. unhandledRejection is a sign that promises weren't chained properly in the test.

lib/config.ts
cnishina: Done.

lib/launcher.ts
cnishina: Done.

lib/launcher.ts
cnishina: Pre and post increment will not matter; however, `i` should always start at 0 or we will be off by one.

lib/launcher.ts
cnishina: Adding both back in.

lib/launcher.ts
```diff
@@ -108,197 +107,160 @@ let initFn = function(configFile: string, additionalConfig: Config) {
...
+
+  taskResults_.reportSummary();
+  let exitCode = 0;
+  if (taskResults_.totalProcessFailures() > 0) {
+    exitCode = RUNNERS_FAILED_EXIT_CODE;
+  } else if (taskResults_.totalSpecFailures() > 0) {
+    exitCode = 1;
+  }
+  await cleanUpAndExit(exitCode);
+  // Start `scheduler.maxConcurrentTasks()` workers for handling tasks in
```
juliemr: This comment should be up by `const maxConcurrentTasks`.

lib/runner.ts
```diff
@@ -381,120 +353,106 @@ export class Runner extends EventEmitter {
...
+
+    // 7) Teardown
+    results = joinTestLogs(results, plugins.getResults());
+    this.emit('testsDone', results);
+    testPassed = results.failedCount === 0;
+    if (this.driverprovider_.updateJob) {
+      await this.driverprovider_.updateJob({'passed': testPassed});
+      await this.driverprovider_.teardownEnv();
+    } else {
+      await this.driverprovider_.teardownEnv();
```
juliemr: Simplify to:

```ts
if (this.driverprovider.updateJob) {
  await this.driverprovider-.updateJob({'passed': testPassed});
}
await this.driverprovider_.teardownEnv();

lib/launcher.ts
cnishina: Done.

PR #5056, opened 2018-11-26
fix(sauce): use same hostname for Sauce rest API as for Sauce backend

christian-bromann:
## Problem

SauceLabs has now a DC in Europe with backend url `ondemand.eu-central-1.saucelabs.com:443/wd/hub`. If people want to run their tests in this DC the job report should also be done their.

## Solution

Parse sauce selenium address and set hostname for SauceLabs client accordingly.

## Comment

It should not be called "selenium address" and rather "webdriver address" or "webdriver endpoint" but this is a separate PR.

### DIFF
--- a/lib/driverProviders/sauce.ts
+++ b/lib/driverProviders/sauce.ts
@@ -53,8 +53,14 @@ export class Sauce extends DriverProvider {
    *     ready to test.
    */
   protected setupDriverEnv(): q.Promise<any> {
-    let deferred = q.defer();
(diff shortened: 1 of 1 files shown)


wswebcreation:
@cnishina 

Can you also cover this in your milestone https://github.com/angular/protractor/milestone/45?

Tnx!

wswebcreation:
Going to close this in favour of a new PR that will handle multiple EU/US cases with Sauce Labs

PR #5036, opened 2018-11-13
move test_util out of q, rework due to es6

CrispusDH:
- [x] rework `test_util`
- [x] check passed tests locally
- [x] check correct behavior if some test failed
- [x] update node version in CircleCI to LTS (tests were failed on previous v8)
- [ ] change octaral literals to something different

I don't know how to fix it:
```
Octal literals are not allowed in strict mode
console.log('\n>>> \033[1;32mpass\033[0m');
```

PR #5058, opened 2018-11-27
chore(tests): update restart spec specs and plugin specs

cnishina:
- update to ES6
- fix the expected conditions to await when the browser is ready after being forked
- enable more tests


cnishina:
After #5036, will refactor as a fix to tests rather to fix test_util

PR #5063, opened 2018-11-28
chore(test): fix attach session test

cnishina:


scripts/driverProviderAttachSession.js
```diff
@@ -2,105 +2,130 @@
...
-
-// 2. After making the request to create a selenium session, read the selenium
-// session id.
-var setBody = function(chunk) {
-  var body = chunk.toString();
-  sessionId = JSON.parse(body).sessionId;
+// Delete session method to be used at the end of the test as well as
+// when the tests fail.
+const deleteSession = async (sessionId, err) => {
+  await new Promise(resolve => {
```
CrispusDH: `await` is redundant here, because async function will return `Promise<something>` with `await` or without.

scripts/driverProviderAttachSession.js
```diff
@@ -2,105 +2,130 @@
...
+    const deleteOptions = {
+      hostname: 'localhost',
+      port: 4444,
+      path: '/wd/hub/session/' + sessionId,
+      method: 'DELETE'
+    };
+    var req = http.request(deleteOptions, res => {
```
CrispusDH: const instead var

scripts/driverProviderAttachSession.js
```diff
@@ -2,105 +2,130 @@
...
+      port: 4444,
+      path: '/wd/hub/session/' + sessionId,
+      method: 'GET'
+    };
+    const req = http.request(checkOptions, (res) => {
+      let state = '';
+      res.on('data', (chunk) => {    
+        state = JSON.parse(chunk.toString()).state;
+      });
+      res.on('end', function() {
```
CrispusDH: arrow function instead common

scripts/driverProviderAttachSession.js
```diff
@@ -2,105 +2,130 @@
...
-    method: 'DELETE'
-  };
-  var req = http.request(deleteOptions);
-  req.end();
-};
+  // 5. Delete the selenium session.
+  await deleteSession(sessionId); 
+}
+
+run().then();
```
CrispusDH: EOF

scripts/driverProviderAttachSession.js
cnishina: I think this should be
```
const deleteSession = (sessionId, err) => {
  return new Promise(resolve => {
```

scripts/driverProviderAttachSession.js
cnishina: done.

scripts/driverProviderAttachSession.js
cnishina: done.

scripts/driverProviderAttachSession.js
cnishina: done.

PR #5067, opened 2018-11-28
chore(tests): fix spec/ng2/timeout_spec test

cnishina:


CrispusDH:
in file `.../spec/ng2/async_spec.js` you should change line `76` (add `await`):
from: `const text = timeout.$('.val').getText();`
to: `const text = await timeout.$('.val').getText();`

cnishina:
Thanks! I remember making this change in the 95 file PR.

PR #5068, opened 2018-11-28
HOLD till 2019: migrate altRoot to ts

CrispusDH:
It is just an example.
Should I change other files somewhere that depends on `altRoot`? @cnishina


CrispusDH:
@cnishina any updates?

CrispusDH:
kind of mess here, probably will create new branch

lib/taskRunner.ts
```diff
@@ -7,12 +7,12 @@ import {Runner} from './runner';
...
-  specs: Array<string>;
-  capabilities: any;
+  taskId?: number;
+  specs?: Array<string>;
+  capabilities?: any;
   failedCount: number;
-  exitCode: number;
-  specResults: Array<any>;
+  exitCode?: number;
+  specResults?: Array<any>;
```
cnishina: I knew it. :smile:  (A comment from a previous PR that these should be optional)

cnishina: I also ran into this when working through my PRs.

PR #5071, opened 2018-11-30
chore(ignoreSynchornization): clean up to use waitForAngularEnabled

cnishina:


lib/browser.ts
```diff
@@ -216,19 +216,13 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
    *
    * Initialized to `false` by the runner.
    *
-   * This property is deprecated - please use waitForAngularEnabled instead.
+   * ignoreSynchornization is deprecated.
```
heathkit: If we're making this actually, totally private then we can remove this jsdoc.

lib/browser.ts
```diff
@@ -937,15 +931,15 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
             }
           }
         })
-        .then(() => {
+        .then(async () => {
```
heathkit: Can we async/await this promise chain?

cnishina: This was done in my previous 95 file PR.

cnishina: Fixed.

lib/browser.ts
cnishina: Done.

cnishina:
Ready for review.

PR #5072, opened 2018-11-30
chore(test): remove jasmine addMatcher test

cnishina:
Removing the addMatchers test since we no longer support async calls
resolve with jasminewd since we removed jasminewd. Also Jasmine does
not appear to support async calls in custom expectations or the
compare method.


cnishina:
Removed the test.

Issue #5061, opened 2018-11-28
Move lib/framework JavaScript files to TypeScript


cnishina:
Reference comment: https://github.com/angular/protractor/pull/5059 for file lib/frameworks/debugprint.js

- [ ] lib/frameworks/debugprint.js
- [ ] lib/frameworks/jasmine.js
- [ ] lib/frameworks/mocha.js

cnishina:
@CrispusDH Do you want to take a stab at this one? It looks pretty straight forward.

CrispusDH:
@cnishina , yep I will grab it :)

cnishina:
@CrispusDH Could you also take on the setupAfterEach and mocha?

CrispusDH:
@cnishina , yep. I think it will be done before NY )

cnishina:
Take your time. I'm off for the holidays and won't check back in till 2019.

CrispusDH:
@cnishina I have referenced last two PRs

PR #5075, opened 2018-12-01
rework jasmine to ts

CrispusDH:
@cnishina , could you help me with `spec/custom/framework.js`. How to modified this file properly?
@heathkit , could you take a look at it if you have a time


lib/frameworks/jasmine.js
```diff
@@ -1,125 +0,0 @@
...
-  require('./setupAfterEach').setup(runner, specs);
-
-  // Filter specs to run based on jasmineNodeOpts.grep and jasmineNodeOpts.invert.
-  jasmine.getEnv().specFilter = function(spec) {
-    var grepMatch = !jasmineNodeOpts ||
-        !jasmineNodeOpts.grep ||
-        spec.getFullName().match(new RegExp(jasmineNodeOpts.grep)) != null;
-    var invertGrep = !!(jasmineNodeOpts && jasmineNodeOpts.invertGrep);
-    if (grepMatch == invertGrep) {
-      spec.disable();
```
CrispusDH: `jasmine.Spec` does not have such method

CrispusDH:
Reference #5061

CrispusDH:
@cnishina any updates?

CrispusDH:
@cnishina I found the right path

spec/custom/framework.js
```diff
@@ -4,6 +4,5 @@
  *
  * @param {Runner} runner The current Protractor Runner.
  * @param {Array} specs Array of Directory Path Strings.
- * @return {q.Promise} Promise resolved with the test results
  */
-exports.run = require('../../lib/frameworks/jasmine.js').run;
+exports.run = require('../../built/frameworks/jasmine').run;
```
cnishina:
```suggestion
exports.run = require('../../built/lib/frameworks/jasmine').run;
```

The lib path still is required.

CrispusDH: Hm, when I made `tsc` there is no `lib` folder inside the `built`. Also, I checked that with current path it works.

CrispusDH:
@cnishina , both CIs were passed

spec/custom/framework.js
cnishina: Yup. I agree.

CrispusDH:
create new PR #5189

PR #5081, opened 2018-12-06
Updated API examples to use async/await

### LINKED ISSUE #5080: Request to update examples in Protractor API docs for async/await
Some of the examples provided in Protractor API docs should be updated to include how to use the API methods with async/await.

http://www.protractortest.org/#/api

It will be helpful if async/await examples are provided for below API methods. I use some of the API methods extensively in my code, refactoring the code to make it work with async/await.

- Filter
http://www.protractortest.org/#/api?view=ElementArrayFinder.prototype.filter

- Map
http://www.protractortest.org/#/api?view=ElementArrayFinder.prototype.map

- Each
http://www.protractortest.org/#/api?view=ElementArrayFinder.prototype.each

devoto13:
Fixes #5080

### DIFF
--- a/lib/browser.ts
+++ b/lib/browser.ts
@@ -747,8 +747,8 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
    * the wrapped webdriver directly.
    *
    * @example
-   * browser.get('https://angularjs.org/');
-   * expect(browser.getCurrentUrl()).toBe('https://angularjs.org/');
+   * await browser.get('https://angularjs.org/');
+   * expect(await browser.getCurrentUrl()).toBe('https://angularjs.org/');
    *
    * @param {string} destination Destination URL.
    * @param {number=} opt_timeout Number of milliseconds to wait for Angular to
@@ -894,9 +894,9 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
    * Browse to another page using in-page navigation.
    *
    * @example
-   * browser.get('http://angular.github.io/protractor/#/tutorial');
-   * browser.setLocation('api');
-   * expect(browser.getCurrentUrl())
+   * await browser.get('http://angular.github.io/protractor/#/tutorial');
+   * await browser.setLocation('api');
+   * expect(await browser.getCurrentUrl())
    *     .toBe('http://angular.github.io/protractor/#/api');
    *
    * @param {string} url In page URL using the same syntax as $location.url()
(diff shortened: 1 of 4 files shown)


lib/browser.ts
```diff
@@ -747,8 +747,8 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
    * the wrapped webdriver directly.
    *
    * @example
-   * browser.get('https://angularjs.org/');
-   * expect(browser.getCurrentUrl()).toBe('https://angularjs.org/');
+   * await browser.get('https://angularjs.org/');
+   * expect(await browser.getCurrentUrl()).toBe('https://angularjs.org/');
```
IgorSasovets: Actually, you don't need to add `await` into expect when using Jasmine because result comparison performs when promise in expect is already resolved.

devoto13: In JavaScript world, yes. But in TypeScript latest type definitions for Jasmine would not allow it... Let me see if I can correct type definitions, before changing this.

IgorSasovets: Ok, thanks for the quick response

devoto13: I've had a look and it seems that the functionality was available, because Protractor patched Jasmine and it relies on the control flow, which is being removed now. Jasmine itself does not support passing `Promise` into `expect()` function.

So you'll have to use the form described in this PR or alternatively new [expectAsync](https://jasmine.github.io/api/3.3/global.html#expectAsync) function.

IgorSasovets: @devoto13 , thanks for the explanation

CrispusDH: @devoto13 is right, `jasminewd` was removed in `selenium4` branch

cnishina: Yup, `jasminewd` was removed because we are no longer using the control flow (because selenium-webdriver deprecated their control flow). Also if I can get jasmine3.3 to work, we could also write this as:

`expectAsync(browser.getCurrentUrl()).toBe('https://angularjs.org');`

Could be nifty but I think the way that this is written is great.

vsravuri:
@cnishina All the API examples provided for async / await use => syntax

IE11 doesn't support => syntax. Does it mean that next version of Protractor cannot be used in IE11?. Also i'm converting my code to async / await following the updated examples in this PR. selenium4 branch you are working on is testable? I would like to run my code on any working Protractor branch to see my code is working

cnishina:
This is fine. 

The node process that runs Protractor can use the fat arrows. We send client side functions to the browser that are not fat arrows. As long as we do not change the client side scripts to `=>` we should be okay.

cnishina:
Squash merging this. Thanks everyone for the comments and @devoto13 thank you for this PR!

PR #5084, opened 2018-12-11
typings(selenium): try out new version of typings

cnishina:


cnishina:
This appears to be working. I plan to use these typings in place until we have a selenium-webdriver 4 typings in DefinitelyTyped.

cnishina:
Looks like there are some issues with this. Some incremental fixes are in this PR: https://github.com/angular/protractor/pull/5089

cnishina:
This will be removed when DefinitelyTyped is updated.

juliemr:
Can you split the change to jasmine version into a separate PR?

Were any changes necessary to the typings?

cnishina:
I think I changed them by mistake.

heathkit:
Approved, but I just saw the Travis failure and it looks "interesting"

cnishina:
The Travis  failures are weird. I'm not sure if the Travis failures are based on Travis and Sauce connections. 100% of the time they will pass on Circle and maybe 40% of the time they will pass on Travis + Sauce Labs.

I've also noticed that Travis gets a lot of unhandled promise rejections. This is weird. I could maybe throw an exception so it cause a failure. I've tried to inspect the error error, but it isn't very helpful. We can discuss this on Monday.

cnishina:
There are also a lot of these Travis error messages: "Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received"

Maybe we need to adjust our builds to wait more in Travis? Put `travis_wait` in front of every statement and hope for the best? Instead of wrapping the entire test.js as one node process, maybe each test is kicked of as a separate travis command so the wait is not 20+ minutes for one node process to complete.

PR #5085, opened 2018-12-11
chore(browser): remove timing issues with restart and fork

### LINKED ISSUE #5031: browser.forkNewDriverInstance should return a promise that resolves to the new browser
Reference #5019

https://github.com/angular/protractor/blob/ed955e56a839d7f69da43acb6755763220d3681d/lib/runner.ts#L275

"I think this is so convoluted because it needed to support the control flow case as well. This should be cleaned up - forkNewDriverInstance() should just return a promise that resolves to the new browser instance, after its ready."

browser.forkNewDriverInstance should return a promise that resolves to the new browser.

### LINKED ISSUE #5054: Upgrade to Jasmine 3.3.0
Feature Request
---

It would be great to upgrade to Jasmine 3.3.0. 

There's already an issue https://github.com/angular/protractor/issues/4788 and a PR https://github.com/angular/protractor/pull/4842 to upgrade to 3.1.0, but 3.3.0 has some new features including Expectation.withContext() function, which would be a very useful thing for e2e tests.

cnishina:
- remove `.ready` since forking should automatically return a browser
- getNewDriver should return a promised WebDriver that can be awaited
- fix interaction tests and local driver tests
- update unit tests for async await due to getNewDriver fix

closes #5031

### DIFF
--- a/lib/driverProviders/driverProvider.ts
+++ b/lib/driverProviders/driverProvider.ts
@@ -42,7 +42,7 @@ export abstract class DriverProvider {
    * @public
    * @return webdriver instance
    */
-  getNewDriver() {
+  async getNewDriver() {
     let builder: Builder;
     if (this.config_.useBlockingProxy) {
       builder =
@@ -56,7 +56,13 @@ export abstract class DriverProvider {
     if (this.config_.disableEnvironmentOverrides === true) {
       builder.disableEnvironmentOverrides();
     }
-    let newDriver = builder.build() as WebDriver;
+    let newDriver: WebDriver;
+    try {
+      newDriver = await builder.build();
+    } catch (e) {
+      e.code = 135;
+      throw e;
+    }
     this.drivers_.push(newDriver);
     return newDriver;
   }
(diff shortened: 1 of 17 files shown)


cnishina:
This appears to be having some problems. I thought it was something to do with a chrome and chromedriver mismatch. I'll have to find the culprit tests. This might take a while.

cnishina:
So this journey started out as a jasmine upgrade. And its no longer a jasmine upgrade. Maybe that can be the next step after this merge.

lib/driverProviders/driverProvider.ts
```diff
@@ -56,7 +56,13 @@ export abstract class DriverProvider {
     if (this.config_.disableEnvironmentOverrides === true) {
       builder.disableEnvironmentOverrides();
     }
-    let newDriver = builder.build() as WebDriver;
+    let newDriver: WebDriver;
+    try {
+      newDriver = await builder.build();
+    } catch (e) {
+      e.code = 135;
```
heathkit: 135?

cnishina: Probably should actually throw the right error instead of adding the correct error code.

CrispusDH:
I think in this PR could be fixed `interaction_spec`, because there appears `.ready()` method

cnishina:
`.ready` is not a method but a property to a promise for the forked browser instance. This is a weird way to do it. Removing it makes sense.

PR #5088, opened 2018-12-12
deps(webdriver-manager): use replacement

cnishina:
- Current workaround to use webdriver-manager-replacement until we
publish a beta release of webdriver-manager


cnishina:
This will unblock the currently failing tests.

cnishina:
Looks like it passes!

PR #5095, opened 2018-12-15
 deps(selenium): upgrade to selenium 4

cnishina:
- Elements workaround for WebElement.equals
- Added a better unhandled rejection warning message in the launcher
- Remove global function wrappers for mocha (these wrappers went away with
control flow)
- Fix the attach to session driver provider
    
Typing exported from Protractor:
  
- removed ActionSequence and EventEmitter
- removed promise.Promise
- removed Promise, defer, delayed, createFlow, controlFlow, all,
fulfilled, filter, when
    
Typings exported from WebDriver:
    
- removed attachToSession
- removed WebDriver instance methods: touchActions, call
- removed WebElement getSize and getLocation for getRect
- removed redefined global vars for testing
 
Tests:
   
- Fix element equals
- Add missing 'await' in colorList test. This caused unhandled promise
rejections.
- Remove control flow related tests
- Disable the install test. Installing from "file:../../" is not
working.
- Fix the attach to session driver provider test to exit with a 1 if
errors are encountered


cnishina:
Depends on https://github.com/angular/protractor/pull/5084

lib/element.ts
```diff
@@ -1140,11 +1140,13 @@ export class ElementFinder extends WebdriverWebElement {
...
    *     resolved to whether the two WebElements are equal.
    */
-  equals(element: ElementFinder|WebElement): Promise<any> {
-    return WebElement.equals(
-               this.getWebElement(),
-               (element as any).getWebElement ? (element as ElementFinder).getWebElement() :
-                                                element as WebElement) as Promise<any>;
+  async equals(element: ElementFinder|WebElement): Promise<boolean> {
+    const a = await this.getWebElement();
+    const b = (element as any).getWebElement ? await(element as ElementFinder).getWebElement() :
```
CrispusDH: space between `await` and `(`

cnishina: Done.

cnishina: Turns out, clang does not like this.

lib/browser.ts
```diff
@@ -505,11 +505,11 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
     }
 
     // TODO(selenium4): fix promise cast.
```
heathkit: Is this todo still relevant?

lib/driverProviders/attachSession.ts
```diff
@@ -38,7 +38,10 @@ export class AttachSession extends DriverProvider {
   getNewDriver(): WebDriver {
     const httpClient = new http.HttpClient(this.config_.seleniumAddress);
     const executor = new http.Executor(httpClient);
-    const newDriver = WebDriver.attachToSession(executor, this.config_.seleniumSessionId, null);
+    const session = new Session(this.config_.seleniumSessionId, null);
+
+    // const newDriver = WebDriver.attachToSession(executor, this.config_.seleniumSessionId, null);
```
heathkit: Should this be deleted?

lib/launcher.ts
```diff
@@ -171,7 +171,15 @@ let initFn = async function(configFile: string, additionalConfig: Config) {
   });
 
   process.on('unhandledRejection', (reason: Error | any, p: Promise<any>) => {
-    logger.warn('Unhandled rejection at:', p, 'reason:', reason);
+    // let e = new Error();
```
heathkit: Old comment?

spec/install/test.js
```diff
@@ -58,7 +50,6 @@ function test(file) {
   }
 }
 
-install();
 tsc();
-test('tmp/conf.js');
-test('tmp/typescript_conf.js');
+// test('tmp/conf.js');
```
heathkit: Do these tests not work anymore?

lib/driverProviders/attachSession.ts
cnishina: Yes.

lib/launcher.ts
cnishina: Yup.

spec/install/test.js
cnishina: The entire install script does not work. We need to rethink this whole thing.

PR #5101, opened 2018-12-19
chore(expectedConditions): update generic Function typings

cnishina:
- Use `() => Promise<boolean>` over `Function` typings.
- Fix an ExpectedConditions test where it was set to a const.
- Fix a TypeScript typing interface issue with RunResults in taskRunner.

### DIFF
--- a/lib/expectedConditions.ts
+++ b/lib/expectedConditions.ts
@@ -59,11 +59,10 @@ export class ProtractorExpectedConditions {
    *
    * @returns {!function} An expected condition that returns the negated value.
    */
-  not(expectedCondition: Function): Function {
-    return (): Function => {
-      return expectedCondition().then((bool: boolean): boolean => {
-        return !bool;
-      });
+  not(expectedCondition: Function): (() => Promise<boolean>) {
+    return async(): Promise<boolean> => {
+      const bool = await expectedCondition();
+      return !bool;
     };
   }
 
@@ -78,20 +77,19 @@ export class ProtractorExpectedConditions {
    * @returns {!function} An expected condition that returns a promise which
    *     evaluates to the result of the logical chain.
    */
-  logicalChain_(defaultRet: boolean, fns: Array<Function>): Function {
+  logicalChain_(defaultRet: boolean, fns: Array<Function>): (() => Promise<boolean>) {
     let self = this;
-    return (): boolean => {
+    return async(): Promise<boolean> => {
       if (fns.length === 0) {
         return defaultRet;
       }
-      let fn = fns[0];
-      return fn().then((bool: boolean): boolean => {
-        if (bool === defaultRet) {
-          return self.logicalChain_(defaultRet, fns.slice(1))();
-        } else {
-          return !defaultRet;
-        }
-      });
+      const fn = fns[0];
+      const bool = await fn();
+      if (bool === defaultRet) {
+        return self.logicalChain_(defaultRet, fns.slice(1))();
+      } else {
+        return !defaultRet;
+      }
     };
   }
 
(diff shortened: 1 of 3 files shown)


cnishina:
That worked! There was a bug in the test suite when setting the EC to a const. Since it is a global var it can be changed to a different instance. This is because the EC is tied to the browser, browsers are tied to the session. If you restart the browser or fork the browser, the const EC points to an old session.

I am not sure how this passed in previous tests. We must have made a change recently.

lib/expectedConditions.ts
```diff
@@ -301,11 +301,10 @@ export class ProtractorExpectedConditions {
    * @returns {!function} An expected condition that returns a promise
    *     representing whether the URL contains the string.
    */
-  urlContains(url: string): Function {
-    return () => {
-      return this.browser.driver.getCurrentUrl().then((actualUrl: string): boolean => {
-        return actualUrl.indexOf(url) > -1;
-      });
+  urlContains(url: string): (() => Promise<boolean>) {
```
cnishina: This might be better either reverted or changed in a separate commit. I'm not too worried about these methods.

spec/basic/expected_conditions_spec.js
```diff
@@ -1,8 +1,9 @@
-const EC = protractor.ExpectedConditions;
+let EC;
```
cnishina: I guess this could have just been `let EC = protractor.ExpectedConditions`

cnishina:
Backed out Jasmine 3.3 changes for ExpectedConditions and other tsc errors.

lib/expectedConditions.ts
cnishina: Made all changes to EC in this commit.

spec/basic/expected_conditions_spec.js
cnishina: Fixed.

lib/expectedConditions.ts
```diff
@@ -205,13 +201,16 @@ export class ProtractorExpectedConditions {
...
+  textToBePresentInElement(elementFinder: ElementFinder, text: string): (() => Promise<boolean>) {
+    let hasText = async () => {
+      try {
+        const actualText = await elementFinder.getText();
         // MSEdge does not properly remove newlines, which causes false
         // negatives
         return actualText.replace(/\r?\n|\r/g, '').indexOf(text) > -1;
-      }, falseIfMissing);
+      } catch (e) {
+        falseIfMissing(e);
```
heathkit: Shouldn't this return?

lib/expectedConditions.ts
```diff
@@ -232,11 +231,15 @@ export class ProtractorExpectedConditions {
...
-      return elementFinder.getAttribute('value').then((actualText: string): boolean => {
+  textToBePresentInElementValue(elementFinder: ElementFinder, text: string):
+      (() => Promise<boolean>) {
+    let hasText = async () => {
+      try {
+        const actualText = await elementFinder.getAttribute('value');
         return actualText.indexOf(text) > -1;
-      }, falseIfMissing);
+      } catch (e) {
+        falseIfMissing(e);
```
heathkit: Same as above

lib/browser.ts
```diff
@@ -838,7 +838,7 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
 
     // Reset bpClient sync
     if (this.bpClient) {
-      await this.bpClient.setWaitEnabled(!this.internalIgnoreSynchronization);
+      await this.bpClient.setWaitEnabled(await this.waitForAngularEnabled());
```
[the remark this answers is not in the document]
cnishina: Done.

lib/expectedConditions.ts
cnishina: done.

lib/expectedConditions.ts
cnishina: done.

PR #5102, opened 2018-12-19
 deps(jasmine): upgrade jasmine 3.3

cnishina:


cnishina:
This PR is chained to https://github.com/angular/protractor/pull/5101. So the commit that should be reviewed is this one: https://github.com/angular/protractor/pull/5102/commits/e6d3939a076ebdf13a1e8f6b213d56e9565387a3

spec/basic/expected_conditions_spec.js
```diff
@@ -1,8 +1,9 @@
-const EC = protractor.ExpectedConditions;
-
 describe('expected conditions', () => {
+  let EC = null;
```
heathkit: Why do we need to move this into a beforeEach?

lib/expectedConditions.ts
```diff
@@ -278,12 +280,15 @@ export class ProtractorExpectedConditions {
...
-  titleIs(title: string): Function {
-    return () => {
-      return this.browser.driver.getTitle().then((actualTitle: string): boolean => {
-        return actualTitle === title;
-      });
+  titleIs(title: string): (() => Promise<boolean>) {
+    return async(): Promise<boolean> => {
+      const actualTitle = await this.browser.driver.getTitle();
+      return actualTitle === title;
+<<<<<<< HEAD
```
heathkit: Conflict markers

lib/expectedConditions.ts
```diff
@@ -205,13 +201,16 @@ export class ProtractorExpectedConditions {
...
+  textToBePresentInElement(elementFinder: ElementFinder, text: string): (() => Promise<boolean>) {
+    let hasText = async () => {
+      try {
+        const actualText = await elementFinder.getText();
         // MSEdge does not properly remove newlines, which causes false
         // negatives
         return actualText.replace(/\r?\n|\r/g, '').indexOf(text) > -1;
-      }, falseIfMissing);
+      } catch (e) {
+        falseIfMissing(e);
```
heathkit: Should return

lib/browser.ts
```diff
@@ -838,7 +838,7 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
 
     // Reset bpClient sync
     if (this.bpClient) {
-      await this.bpClient.setWaitEnabled(!this.internalIgnoreSynchronization);
+      await this.bpClient.setWaitEnabled(await this.waitForAngularEnabled());
```
heathkit: I think this change is also in the other PR - it's ok to use internalIgnoreSynchronization here

lib/expectedConditions.ts
cnishina: Fixed in a different PR.

cnishina:
Not working because it still relies on the previous PR to be approved. I will rebase it later.

cnishina:
Rebasing...

praneethmendu:
Can someone please add some reference to why it was rolled back to 2.8 and how, if it is even possible, to work with 3.3 now ?

PR #5126, opened 2019-01-24
chore(jasmine): prevent random execution order in jasmine 3

cnishina:


lib/frameworks/jasmine.js
```diff
@@ -69,6 +69,12 @@ exports.run = async function(runner, specs) {
   const reporter = new RunnerReporter(runner);
   jasmine.getEnv().addReporter(reporter);
 
+  // Jasmine 3 allows for tests to be in random order by default. This does not
+  // work well with e2e tests where the browser state is determined by the
+  // order of the tests. Setting to false will prevent random execution.
+  // See https://jasmine.github.io/api/3.3/Env.html
+  jasmine.getEnv().randomizeTests(false);
```
heathkit: We could also make this a config option, though I'd understand if you didn't want to bother with that. I think ideally, this would be the default and we'd disable it for tests that are broken by random execution.

cnishina: This actually further down in the file. See https://github.com/angular/protractor/blob/9c3b39e6ee11986a87f88e4e0629f87c3f06c863/lib/frameworks/jasmine.js#L94

```
  // Run specs in semi-random order
  if (jasmineNodeOpts.random) {
    jasmine.getEnv().randomizeTests(true);
    // Sets the randomization seed if randomization is turned on
    if (jasmineNodeOpts.seed) {
      jasmine.getEnv().seed(jasmineNodeOpts.seed);
    }
```

PR #5127, opened 2019-01-24
chore(types): fix types to use not @types/selenium-webdriver

cnishina:


lib/browser.ts
```diff
@@ -882,7 +860,7 @@ export class ProtractorBrowser extends AbstractExtendedWebDriver {
    * await browser.get('http://angular.github.io/protractor/#/tutorial');
    * await browser.setLocation('api');
    * expect(await browser.getCurrentUrl())
-   *     .toBe('http://angular.github.io/protractor/#/api');
+   *     .toBe('http://angular.g../../ithub.io/protractor/#/api');
```
CrispusDH: Not sure about it line

PR #5142, opened 2019-02-07
source-map-support version updated to 0.5.10 (latest)

selimbez:
Removed [DEP0005] DeprecationWarning: Buffer()


cnishina:
Could you sign the CLA and rebase? Also could we use ~0.5.11? Is there a reason to pin the version of the source map support?

PR #5144, opened 2019-02-08
deps(package): update package.json versions for vulnerabilities

cnishina:


CrispusDH:
You might need to update here as well? https://github.com/cnishina/protractor/blob/sourcemap/circle.yml#L57

cnishina:
There's probably something in pretest / gulpfile that should be updated. I (for some reason) backed out that change.

circle.yml
```diff
@@ -54,8 +54,8 @@ jobs:
           name: Selenium Start
           background: true
           command: |
-            ./node_modules/.bin/webdriver-manager update
-            ./node_modules/.bin/webdriver-manager start
+            ./node_modules/.bin/webdriver-manager update --versions.chrome=2.43
```
heathkit: Why do we need to pin the chromedriver version? Is Circle stuck on Chrome 69 for some reason?

cnishina: Yes CircleCI is giving me problems. This could be resolved if we could download and run a Chromium  version.  For now this is the fix.

cnishina:
Closing this. There have been other PRs that were submitted that do audit checks.

PR #5189, opened 2019-03-26
re-write jasmine custom framework from js to ts

CrispusDH:
@cnishina some strange fails on CI. I tried to copy what was done from #5075


cnishina:
It looks like there is some issue with the Chrome installation on circle ci. Travis passes so that's a big win! Let me figure this out.

PR #5193, opened 2019-03-27
Add built:copy:typings task to prepublish task

sbley:
This fixes typescript compile errors due to missing typings.


sbley:
I signed it!

PR #5200, opened 2019-03-29
chore(jasmine): remove random test options

cnishina:


juliemr:
Should we document this as a breaking change?

lib/frameworks/jasmine.js
```diff
@@ -73,13 +73,15 @@ exports.run = async function(runner, specs) {
   // work well with e2e tests where the browser state is determined by the
   // order of the tests. Setting to false will prevent random execution.
   // See https://jasmine.github.io/api/3.3/Env.html
+  jasmine.getEnv().configuration().random = false;
   jasmine.getEnv().randomizeTests(false);
+  jasmineNodeOpts.random = false;
```
juliemr: Is there any point to keeping this around? AFAIK this does nothing now?

cnishina:
Yes this should be marked as a breaking change to the random option.

lib/frameworks/jasmine.js
cnishina: Removed.

PR #5237, opened 2019-05-15
Updated the link for mobile doc because previous was expired or moved…

Abhiknoldur:
… from that link
oss4


Abhiknoldur:
I did sign the CLA

PR #5294, opened 2019-08-08
feat(config): include the ability to always fork taskRunners

barret-p:
This will provide the ability to fork taskRunner process even running only a single spec file. 

**What's the purpose of this feature?**
Right now the output format of the logs are different depending on if there are multiple tests or a single spec file. I would like to have the ability to make the log outputs consistent.

[Output log of single spec file](https://i.imgur.com/sfJwi49.png)
image

[Output log of spec file that is part of suite or multiple spec run](https://i.imgur.com/1XLN2jU.png)
image

**Differences**
In the case of the logs for the suite for multiple specs, the log prints the name of the spec each time before each run and also prints the testId in the left column (e.g. ```#01-9```).

Our log files are parsed and analyzed for data. In the case where run a suite, and then only a single spec fails and retries, the logs of the retried run of a single spec will break our parsing function.


barret-p:
@googlebot I signed it!

barret-p:
@googlebot I fixed it.

barret-p:
@cnishina I believe yall's Travis CI build is misconfigured.

PR #5326, opened 2019-09-30
Fix incorrect exports in index.ts

### LINKED ISSUE #5325: Incorrect imports in lib/index.ts
- Typescript: `3.7`, currently available as `typescript@3.7.0-beta`
- Protractor Version: `master`

lib/index.ts incorrectly re-exports two types from plugins.ts:

```ts
import {PluginConfig, ProtractorPlugin} from './plugins';
export type PluginConfig = PluginConfig;
export type ProtractorPlugin = ProtractorPlugin;
```

The references in the type aliases should be circular, but in current versions of Typescript, are not. The correct way to write this is:

```ts
export {PluginConfig, ProtractorPlugin} from './plugins';
```

In Typescript 3.7 the current code will be an error. The beta is available as `typescript@3.7.0-beta`. 

I'll create a PR shortly with the fix.

sandersn:
Fixes #5325

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -2,7 +2,7 @@ import {ElementHelper, ProtractorBrowser} from './browser';
 import {ElementArrayFinder, ElementFinder} from './element';
 import {ProtractorExpectedConditions} from './expectedConditions';
 import {ProtractorBy} from './locators';
-import {PluginConfig, ProtractorPlugin} from './plugins';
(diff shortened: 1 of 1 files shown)


sandersn:
Weird, Travis' PR build timed out. I doubt that problem is caused by the change in this PR.

daniraja:
@sandersn thank you!
any ETA, when this patch will be out?

PR #5328, opened 2019-10-07
Fix: Now correctly allows to override ElementFinder and ElementArrayFinder methods

Xotabu4:
Fix: Now correctly allows to override ElementFinder and ElementArrayFinder methods

Because monkeypatching was done in constructor against 'this', all objects that was trying to extend ElementFinder or ElementArrayFinder and override some of WebElements methods: 
let WEB_ELEMENT_FUNCTIONS = [
  'click', 'sendKeys', 'getTagName', 'getCssValue', 'getAttribute', 'getText', 'getRect',
  'isEnabled', 'isSelected', 'submit', 'clear', 'isDisplayed', 'getId', 'takeScreenshot'
];

```
class Component extends ElementFinder {
  constructor(elementFinder: ElementFinder) {
    super(elementFinder.browser_, elementFinder.elementArrayFinder_);
  }

 async click() {
     console.log('OWN click!')
     await super.click()
 }
}

new Component($('button')).click()
```

Before this PR code above would not be working - since monkey-patching is overrides methods declared in child classes

### DIFF
--- a/lib/element.ts
+++ b/lib/element.ts
@@ -80,17 +80,6 @@ export class ElementArrayFinder extends WebdriverWebElement {
       public browser_: ProtractorBrowser, public getWebElements: () => Promise<WebElement[]> = null,
       public locator_?: any, public actionResults_: Promise<any> = null) {
     super();
-
-    // TODO(juliemr): might it be easier to combine this with our docs and just
-    // wrap each one explicity with its own documentation?
-    WEB_ELEMENT_FUNCTIONS.forEach((fnName: string) => {
-      this[fnName] = (...args: any[]) => {
-        let actionFn = (webElem: any) => {
-          return webElem[fnName].apply(webElem, args);
-        };
-        return this.applyAction_(actionFn);
-      };
-    });
   }
 
   /**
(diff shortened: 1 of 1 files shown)


lib/element.ts
```diff
@@ -730,6 +719,16 @@ export class ElementArrayFinder extends WebdriverWebElement {
     return this.applyAction_(allowAnimationsTestFn);
   }
 }
+// TODO(juliemr): might it be easier to combine this with our docs and just
+// wrap each one explicity with its own documentation?
+WEB_ELEMENT_FUNCTIONS.forEach((fnName) => {
```
CrispusDH: if `fnName` can be only `string` you could add this type

lib/element.ts
```diff
@@ -730,6 +719,16 @@ export class ElementArrayFinder extends WebdriverWebElement {
...
+  ElementArrayFinder.prototype[fnName] = function (...args: any[]) {
```
CrispusDH: do you need here `()` as you did for `ElementFinder` (`(ElementFinder.prototype[fnName])`)?

CrispusDH:
@cnishina could you take a look at it useful PR? :)

PR #5338, opened 2019-10-18
Update package.json for saucelabs

mearleycf:
https-proxy-agent has a vulnerability. Saucelabs has taken it out of their package, but the version of saucelabs you're using has not removed it yet. If you upgrade your saucelabs package dependency it will eliminate the need for https-proxy-agent and fix the vulnerability.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "glob": "^7.0.3",
     "jasmine": "^3.3.1",
     "optimist": "~0.6.0",
-    "saucelabs": "^1.5.0",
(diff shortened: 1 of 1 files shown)


mearleycf:
@googlebot i signed it!

mearleycf:
@googlebot i signed it!

evilaliv3:
This seems to be necessary also for other reasons as saucelabs included in protractor is really outdated: https://github.com/angular/protractor/issues/5419

PR #5398, opened 2020-02-10
docs(release): Update release docs for 5.4 series.

heathkit:


abhinaba1080:
@heathkit does this mean more regular commits and codes coming in for protractor from your/Google's side in this year?

Issue #5145, opened 2019-02-08
multiCapabilities doesn't invoke parallel instances of WebDriver


vsravuri:
**Bug report**
Using Protractor = 6.0.0-beta, multiCapabilities execute individual captibilities in sequential order rather than running  the capabilities in parallel.

- Node Version: `v10.15.1`
- Protractor Version: `6.0.0-beta`
- Selenium Standalone: `3.141.59`
- Browser(s): `Chrome 72/ Firefox 64`
- Chrome Driver: `2.46`
- Operating System and Version `Mac OS 10.14.2`

**- Your protractor configuration file**
// conf.js
exports.config = {
    framework: 'jasmine',
    seleniumAddress: 'http://localhost:4444/wd/hub',
    multiCapabilities: [{
        browserName: 'chrome',
        specs: ['spec.js'],
    }, {
        browserName: 'chrome',
        specs: ['spec.js'],
    }]
};

**- A relevant example test**
// spec.js
describe('Protractor Demo App',  () => {
    it('should have a title', async () => {
        await browser.get('http://juliemr.github.io/protractor-demo/');
        await browser.driver.sleep(5000);
        expect(await browser.getTitle()).toEqual('Super Calculator');
    });
});

**- Output from running the test**
➜  standard git:(vr-async-await) ✗ sudo ./node_modules/protractor/bin/protractor config.js                 
.[13:10:24] I/testLogger - 
------------------------------------
[13:10:24] I/testLogger - [chrome #01] PID: 11283
[chrome #01] Specs: /Users/vr/Documents/new-async-await/kb-automation/standard/spec.js
[chrome #01] 
[chrome #01] [13:10:13] I/local - Starting selenium standalone server...
[chrome #01] [13:10:14] I/local - Selenium standalone server started at http://192.168.0.11:56758/wd/hub
[chrome #01] DEPRECATION: Setting specFilter directly on Env is deprecated, please use the specFilter option in `configure`
[chrome #01] Randomized with seed 98392
[chrome #01] Started
[chrome #01] .
[chrome #01] 
[chrome #01] 
[chrome #01] 1 spec, 0 failures
[chrome #01] Finished in 6.914 seconds
[chrome #01] Randomized with seed 98392 (jasmine --random=true --seed=98392)

[13:10:24] I/testLogger - 

.[13:10:34] I/testLogger - 
------------------------------------

[13:10:34] I/testLogger - [chrome #11] PID: 11309
[...]

PR #5406, opened 2020-03-11
Bugfix max instances

yadimon:
fix for #5145
+ cosmetics


yadimon:
@googlebot I signed it!

yadimon:
added some kind of unit tests for this case, i hope its acceptable

PR #5408, opened 2020-03-12
fix(web-server): since node 12 there is no util.puts

yadimon:
see https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md
and https://github.com/nodejs/node/pull/25377
added jshint cosmetic fixes.

### DIFF
--- a/testapp/scripts/web-server.js
+++ b/testapp/scripts/web-server.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 var express = require('express');
-var bodyParser = require('body-parser')
+var bodyParser = require('body-parser');
 var optimist = require('optimist');
 var path = require('path');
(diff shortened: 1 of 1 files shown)


testapp/scripts/web-server.js
```diff
@@ -27,7 +26,7 @@ var main = function() {
   testApp.use(bodyParser.json());
   testApp.use(testMiddleware);
   testApp.listen(port);
-  util.puts(["Starting express web server in", testAppDir ,"on port", port].
+  console.log(["Starting express web server in", testAppDir ,"on port", port].
```
tymfear: Console log will add spaces for you, no need to join via array
```javascript
console.log("Starting express web server in", testAppDir ,"on port", port)
```

Issue #5413, opened 2020-03-26
protractor > optimist > minimist Prototype Pollution


sdavids:
Bug report
---

- Node Version: `v13.10.1`
- Protractor Version: `5.4.3`
- Angular Version: `9.1.0`

```
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ minimist                                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=0.2.1 <1.0.0 || >=1.2.3                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ protractor [dev]                                             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ protractor > optimist > minimist                             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/1179                            │
└───────────────┴──────────────────────────────────────────────────────────────┘
```

optimist is [deprecated](https://www.npmjs.com/package/optimist)--maybe update to [yargs](https://github.com/chevex/yargs)?

colbyhill21:
The change has been completed by @alan-agius4 although it still hasn't been merged in. Is there any estimate on when these changes will be merged and released?

inpercima:
Is there a way to merge this existing pull request?

evilaliv3:
@alan-agius4 @kyliau: Considering the value of this patch i would suggest to issue a new release.

What do you think?

This could be relevant for many projects. Among the updates if anywone would like to review it i would kindly ask if we could try to include: https://github.com/angular/protractor/pull/5421

\cc @cnishina @sjelin @sjelin

alan-agius4:
@evilaliv3, a release containing this fix has already been cut (5.4.4)

evilaliv3:
Ah, i'm sorry for the confusion, the fact that you edited many comments that you wrote time ago messed up many of the dates that github shows and i fooled me. thank you for the clarification @alan-agius4  and @kyliau

PR #5416, opened 2020-03-31
fix: security prototype pollution

alan-agius4:
Optimist has been deprecated over 2 years ago as has a security vulnerability. With this change we use it's successor `yargs`.

Closes: #5413

**_Kindly merge this to master and patch (release-5.4 branch)_**
//cc @IgorMinar

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -187,12 +187,13 @@ is complete before continuing.
 
 How do I switch off an option in the CLI?
 -----------------------------------------
-i.e. `webdriver-manager update --chrome=false` does not work. 
-This has to do with the way `optimist` parses command line args. In order to pass a false value, do one of the following:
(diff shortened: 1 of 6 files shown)


docs/faq.md
```diff
@@ -188,11 +188,11 @@ is complete before continuing.
 How do I switch off an option in the CLI?
 -----------------------------------------
 i.e. `webdriver-manager update --chrome=false` does not work. 
-This has to do with the way `optimist` parses command line args. In order to pass a false value, do one of the following:
+This has to do with the way `yargs` parses command line args. In order to pass a false value, do one of the following:
```
kyliau, line 191: Is this still the case with current `yargs`? Can it parse `--chrome=false` correctly?

alan-agius4: Good catch, that should be addressed. I’ll remove this section

alan-agius4: Updated.

PR #5424, opened 2020-04-15
fix: security prototype pollution

kyliau:
Optimist has been deprecated over 2 years ago as has a security vulnerability. With this change we use it's successor `yargs`.

Closes: #5413


alan-agius4:
@googlebot I consent.

PR #5436, opened 2020-05-13
doc: update compatibility change with nodejs

kyliau:
Protractor 5 is compatible with nodejs v6 and v8 only.

Protractor 6 has been deprecated.

Protractor 7 is compatible with nodejs v10, v12, and newer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,11 @@ Protractor [![Build Status](https://travis-ci.org/angular/protractor.svg?branch=
 Compatibility
 -------------
 
-Protractor 5 is compatible with nodejs v6 and newer.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,7 +6,11 @@ Protractor [![Build Status](https://travis-ci.org/angular/protractor.svg?branch=
 Compatibility
 -------------
 
-Protractor 5 is compatible with nodejs v6 and newer.
+Protractor 5 is compatible with nodejs v6 and v8 only.
```
alan-agius4, line 9: Nit: Node.js instead of nodejs. (Both lines)

PR #5438, opened 2020-05-14
fix: remove elementExplorer api and command line

alan-agius4:
Protractor debugger and element explorer cannot be used for Node.js 8+ and were removed in version 7.

See: https://github.com/angular/protractor/blob/master/docs/debugging.md#enabled-control-flow


kyliau:
I'll have to make another release and update the github page which is linked to https://www.protractortest.org/#/debugging

alan-agius4:
No rush for this.

PR #5462, opened 2020-08-05
Replace deprecated command

Desnoo:
The command `new Buffer()` is deprecated. Instead, we can use `Buffer.from(...)`.
I have updated the faq.md section of how to save a screenshot.


Desnoo:
@googlebot I signed it!

PR #5468, opened 2020-09-01
Fix for error "Can't return an ElementFinder from an async function" in TS 3/4

awaisraad:
Fixes ElementFinder so that protractor can be used with Typescript v3, v4.

Typescript fails to compile when ElementFinder is returned from an async function. This is because of optional property "then" present on ElementFinder. Previously this property was removed in #1926 but re-added back in #3835 as an optional. So I suggest that this property be removed again so that protractor can be used with Typescript version 3 and onward.

### DIFF
--- a/lib/element.ts
+++ b/lib/element.ts
@@ -816,9 +816,6 @@ export class ElementArrayFinder extends WebdriverWebElement {
 export class ElementFinder extends WebdriverWebElement {
   parentElementArrayFinder: ElementArrayFinder;
   elementArrayFinder_: ElementArrayFinder;
-  then?:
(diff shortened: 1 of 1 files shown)


awaisraad:
@googlebot I signed it!

awaisraad:
r

debajyoti3061:
@awaisraad Thanks for fixing this.Which version of protractor has this fix?

lyubomiraleksiev-abertay:
@debajyoti3061 this pr has not been approved, hence has not been merged into master.

awaisraad:
@debajyoti3061 I have removed my PR because the email was different in commits and the one with which I signed SLA.

Issue #5481, opened 2020-11-26
Security issue: Please update yargs to 16.1.1 to fix CVE in y18n 


SymbioticKilla:
Protractor Version 7.0.0
Security issue:
[CVE in y18n](https://www.whitesourcesoftware.com/vulnerability-database/CVE-2020-7774)

Thank you!

neil-gok:
Is there an ETA on when yargs will be updated.  There is a critical vulnerability on one of its dependancies, y18n@4.0.0.

SymbioticKilla:
Recreate lock file => it should update y18n to 4.0.1+ version:
This affects the package y18n **before 3.2.2, 4.0.1 and 5.0.5**
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-7608

PR #5482, opened 2020-12-04
#5481 Security issue: Please update yargs to 16.1.1 to fix CVE in y18n

SymbioticKilla:
#5481


neil-gok:
When will an update be released to address the vulnerability?

PR #5518, opened 2021-05-27
ci: upgrade to node 14

kyliau:


.circleci/config.yml
```diff
@@ -1,13 +1,16 @@
-version: 2
+version: 2.1
+orbs:
+  browser-tools: circleci/browser-tools@1.1.0
 jobs:
   build:
     docker:
-      - image: circleci/node:8.11-browsers
+      - image: cimg/node:14.17.0-browsers
```
Splaktar, line 7: What is `cimg` and why is it better than the `circleci/node` images? 

Most Angular repos have standardized on NodeJS `14.16.1` as supported by @josephperrott. Should we use that version here as well?

Splaktar (commented):
Glad to see an effort to support NodeJS LTS!

.circleci/config.yml
josephperrott: `cim/node` is the newer image "designed to supercede the legacy CircleCI Node.js image, `circleci/node`"

Though I agree that ideally we should use `14.16.1` here unless we have a compelling reason to use `14.17.0`

Splaktar:
Circle CI is failing with:
```
[03:18:46] E/direct - Error code: 135
[03:18:46] E/direct - Error message: Could not find geckodriver at null. Run 'webdriver-manager update' to download binaries.
[03:18:46] E/direct - Error: Could not find geckodriver at null. Run 'webdriver-manager update' to download binaries.
[03:19:00] E/driverProvider - Error code: 135
[03:19:00] E/driverProvider - Error message: Unable to create new service: GeckoDriverService
[03:19:00] E/driverProvider - Error: Unable to create new service: GeckoDriverService
Build info: version: '4.0.0-alpha-1', revision: 'd1d3728cae', time: '2019-04-24T16:15:24'
System info: host: '59788622abe5', ip: '192.168.128.3', os.name: 'Linux', os.arch: 'amd64', os.version: '4.15.0-1102-aws', java.version: '11.0.11'
Driver info: driver.version: unknown
```

Splaktar:
Travis CI is reporting some issues with Sauce Connect 4.5.4 failing to connect.
Related note from Sauce Labs support:
```
Sauce Connect Reminder: End of Support Dates
As part of our continued effort to deliver the best and most secure Sauce Connect experience, 
we are focusing on the most recent version, 4.6.3. As announced in March, 
we’ve ended support for versions below 4.6.3, and effective July 26, 
versions 4.5.4 and below have stopped working. 

Action Required: If you are running version 4.5.4 or below, 
please upgrade as soon as possible to continue using Sauce Connect. 
```

evilaliv3:
Thank you @Splaktar for looking into this!!

In relation to travis we shall look into updating https://github.com/angular/protractor/blob/4bc80d1a459542d883ea9200e4e1f48d265d0fda/scripts/sauce_connect_setup.sh

PR #5557, opened 2023-01-27
Project Status Page

dwelch2344:
With Protractor becoming LTS and given the partnership with [HeroDevs](https://www.herodevs.com) to provide support past August 2023, we had discussed adding documentation to explain what the future holds. 

Happy to collaborate on the messaging, style, or any other aspect! We just wanted to take a swing at it to help with the lift! 

Screen shot of change:
Screen Shot 2023-01-27 at 4 27 39 PM

**Note:** This PR builds on #5556 (as it was required to run the site for review)

PR #5556, opened 2023-01-27
Build/Docs: Update Dependencies for building documentation website

dwelch2344:
Working on getting a PR together regarding the project status but couldn't build the project and/or doc website, so I wanted to provide a clean PR to update some instructions and dependencies. 

Ran into a surprising amount of roadblocks, most due to a lot of version mismatches (and the worst due to native binaries not being available for old versions of node + Apple's new silicon chipset) so I tried to clean things up the best I could for a first pass. 

I updated `website/README.md` to reflect the correct build and launch process now, as well as included a monkey-patch script that _should_ get the project running locally. Aiming to update all of this down the road when I have more time, but hopefully this will be acceptable in the meantime.

PR #5560, opened 2023-04-21
Project Status Page (5.4 branch) 

dwelch2344:
With Protractor becoming LTS and given the partnership with [HeroDevs](https://www.herodevs.com/) to provide support past August 2023, we had discussed adding documentation to explain what the future holds.

The bulk of this is adding a `banner` to every page and a `status page` that gives more information about EOL. Both of these have 3 minor variants of verbiage, depending on the browser's date on render. Screenshots below:

## Now - Aug 2023
current

## Sept 2023 - Aug 2024
sept 2023 to aug 2024

## Sept 2024+
sept 2024

_Note_: This PR replaces #5557 (and #5556) to both address EOL and update the docs / builds for the website. Apologies for another PR, but it looks like the current site is currently running off a 5.4 commit that never made it to the release branch.

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -6,7 +6,8 @@ Generates the documentation for the protractor website.
 ## How to run
 
 ```shell
-npm run compile_to_es5
+npm i
+npm run build
 cd website/
 npm install
 npm run build
@@ -16,10 +17,29 @@ Then copy all the files from the `build` directory to the gh-pages branch.
 If you want to run the website locally then run the following command:
 
 ```shell
-npm start
+npm run website
 ```
 
 And point your browser to: [localhost:8080](http://localhost:8080/)
+
+
+### Build Failures on Apple Silicon / Arm64
+
+Some of the nested dependencies do not have compatible binaries for newer architectures. This generally presents as a fatal failure like the following: 
+```
+#
+# Fatal error in , line 0
+# Check failed: !value_obj->IsJSReceiver() || value_obj->IsTemplateInfo().
+#
+#
+#
+#FailureMessage Object: 0x306a626c0
+```
+
+Eventually the project should migrate to Gulp 5 / Node 18+ to solve this issue, but in the meantime a simple workaround is: 
+1. `cd website`
+2. `npm i && npm run arm64_repair`
+3. `npx gulp liveReload` (to avoid the `prestart` script)
 
 ## How to run the tests
 
--- a/website/partials/status.html
+++ b/website/partials/status.html
@@ -1,0 +1,40 @@
+<div class="lh-1-5">
+  <h1 class="mb-50">Project Status</h1>
+
+  <h2>Is Protractor actively maintained?</h2>
+  <div ng-hide="isEOL">
+    <p>
+      Although active development on Protractor ended in December of 2022, the project will continue to be supported under <strong>Long Term Support</strong> and will continue to receive critical security &amp; compatibility fixes via the community through August 2023. This is important as there are many actively developed projects still benefiting from their existing code coverage, and we anticipate Protractor to continue functioning well into the future. 
+    </p>
+    <p>
+      In short: Protractor's current status is <strong>Long Term Support</strong> and <strong>will End Of Life in September 2023</strong>.
+    </p>
+  </div>
+  
+  <div ng-show="isEOL">
+    <p>
+      Although active development on Protractor ended in December of 2022, the project continued to be supported under <strong>Long Term Support</strong> with critical security &amp; compatibility fixes via the community through August 2023.
+    </p>
+    <p>
+      As of September 2023, Protractor's status is officially <strong>End Of Life</strong>.
+    </p>
+  </div>
+  
+  <h2>What does End Of Life mean?</h2>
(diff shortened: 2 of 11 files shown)


lib/selenium-webdriver/webdriver.js
```diff
@@ -236,7 +236,7 @@ webdriver.WebDriver.prototype.executeAsyncScript = (script, var_args) => {};
  * button.click();
  * browser.call(logText, counter);
  *
- * @param {function(...): (T|promise.Promise<T>)} fn The function to
+ * @param {function(...*): (T|promise.Promise<T>)} fn The function to
```
mgechev: If this change is not related to the banner update, I'd recommend reverting back. I assume you're not able to build without it?

website/README.md
```diff
@@ -6,21 +6,39 @@ Generates the documentation for the protractor website.
 ## How to run
 
 ```shell
-npm run compile_to_es5
+npm i
+npx gulp
 cd website/
 npm install
-npm run build
+npx gulp
```
dgp1130: Can you include the version of `gulp` you're using for future reference? That way updates to `gulp` don't break this process.

website/README.md
```diff
@@ -6,21 +6,39 @@ Generates the documentation for the protractor website.
...
+#
+# Fatal error in , line 0
+# Check failed: !value_obj->IsJSReceiver() || value_obj->IsTemplateInfo().
+#
+#
+#
+#FailureMessage Object: 0x306a626c0
+```
+
+Eventually the project should migrate to Gulp 5 / Node 18+ to solve this issue, but in the meantime a simple workaround to run `npm run arm64_repair` (after every `npm install` in the `website/` directory)
```
dgp1130: Should this be included in the "How to run" instructions above?

website/partials/status.html
```diff
@@ -0,0 +1,40 @@
...
+    <p>
+      Although active development on Protractor ended in December of 2022, the project continued to be supported under <strong>Long Term Support</strong> with critical security &amp; compatibility fixes via the community through August 2023.
+    </p>
+    <p>
+      As of September 2023, Protractor's status is officially <strong>End Of Life</strong>.
+    </p>
+  </div>
+  
+  <h2>What does End Of Life mean?</h2>
+  The packages will remain accessible on GitHub, npm, Bower, and Release archive. This website will also remain here indefinitely. The project will be archived on GitHub and remain accessible indefinitely, but will no longer be open to issues / comments / etc.
```
dgp1130: > This website will also remain here indefinitely.

@mgechev, are we comfortable committing to the website staying up _forever_? I doubt GitHub Pages would push a significant breaking change here, but if they did and it was a non-trivial effort to move off it, I expect we probably wouldn't bother given Internet Archive as the ultimate workaround.

dgp1130 (changes requested):
I recommend squashing these commits, as the last two don't provide a lot of value and just fixup stuff from the first.

Wording of the messages sounds good to me, just one note about the website lifetime.

website/partials/status.html
mgechev: Yeah, I agree. Let's change to "foreseeable future"

We're not anticipating making the website unavailable, but inevitably it'll happen one day.

lib/selenium-webdriver/webdriver.js
dwelch2344: Happy to do so! Yeah, I'm a little concerned about the build process to actually deploy this as there were a few docs / build scripts that felt like they may be half-updated. I was able to get it to build, but it required updating the `dgeni` version for a bug fix (and thus where this came in)

website/README.md
dwelch2344: I believe `npx gulp` will first look in the `node_modules/.bin/` first, so it should pin directly to whatever's listed in the `package.json`. That said, happy to include the specific version if preferable!

dwelch2344:
@mgechev @dgp1130 Thanks for the review! Cleaned up 3 of the 4 comments, left a question on the final one for @dgp1130

website/README.md
dgp1130: Oh I didn't realize that was properly listed as a dev dependency. How about we make `"gulp": "gulp"` a script and then document it as `npm run gulp`. That way it's specifically tied to the installed dependency, not downloaded over the network (ex. if the user forgot to `npm install` first).

mgechev (approved):

website/README.md
dwelch2344: Great call! I had totally missed it, but `npm run build` would essentially do just that

dwelch2344:
@dgp1130 I think we're good to go. I squashed to a single commit and cleaned up the issues, but definitely let me know if  there's anything else we can do on our end! Thanks

dgp1130 (approved):
Yep, looks great. Thanks @dwelch2344!

dgp1130:
I'll take a look at building and pushing to `gh-pages` tomorrow to see if we can get this out.

dgp1130:
Sorry, got distracted with v16. Pushed the release now and protractortest.org seems to be updated correctly.

I needed to add back in that typing change to get it to build, not sure how you were avoiding that. I also needed to swap a boolean condition as it seems the header message was displayed in the wrong order.

https://github.com/angular/protractor/commit/bb0e475b449fe2ed2a71c2e63bc1ab46f7a03309

